PR #24, opened 2018-11-28
Update readme and samples

pierreca:
- Separate samples in multiple files
- Add comments
- Update readme to reflect those changes


azure-iot-hub-devicesdk/samples/authentication_providers.py
```diff
@@ -0,0 +1,37 @@
...
+# to connect securely to an Azure IoT hub.
+
+# An Authentication Provider can be created from a few different things:
+# - A shared access signature:
+sas_auth_provider = from_shared_access_signature(os.getenv("IOTHUB_DEVICE_SAS_STRING"))
+
+# - A symmetric key:
+key_auth_provider = from_connection_string("IOTHUB_DEVICE_CONNECTION_STRING")
+
+# - Pre-defined environment variables (this is especially useful when running on Azure IoT Edge)
```
BertKleewein: >especially useful [](start = 47, length = 17)

I think you copy/pasted this phrase from me, but I would say "this only works when running as an Azure IoT Edge Module"

azure-iot-hub-devicesdk/samples/send_telemetry_device.py
```diff
@@ -0,0 +1,52 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+import os
+from threading import Timer
+from azure.iot.hub.devicesdk.device_client import DeviceClient
+from azure.iot.hub.devicesdk.auth.authentication_provider_factory import from_connection_string
+
+# The connection string for a device should never be stored in code. For the sake of simplicity we're using an environment variable here.
+conn_str = os.getenv("IOTHUB_DEVICE_CONNECTION_STRING")
+# The "Authentication Provider" is the object in charge of creating authentication "tokens" for the device client.
```
BertKleewein: >"tokens" [](start = 83, length = 8)

putting this in quotes makes it lose credibility.  I would maybe say: "The authentication provider is the object that manages the connection credentials for the device client."  I don't like the word "manages" either, but I'm not sure how else to phrase it.

azure-iot-hub-devicesdk/samples/send_telemetry_device.py
```diff
@@ -0,0 +1,52 @@
...
+# This function will be called by a timer on a regular basis, once connected
+def send_payload():
+    print("sending!")
+    device_client.send_event("test_payload")
+    start_sender()
+
+
+def start_sender():
+    # This defines a timer that fires after 5 seconds
+    event_sender = Timer(5.0, send_payload)
```
BertKleewein: >event_sender [](start = 4, length = 12)

same comment as from readme.  This needs "global event_sender" to let the interpreter know that this is not a local.  Same for cancel_sender.

PR #26, opened 2018-11-29
 make client operations synchronous

BertKleewein:
The goal of this commit is to to make the "default" ModuleClient and DeviceClient operations synchronous.  

Later we will add async counterparts for Python versions that support the async/await syntax, and this design of this change takes future async/await functionality into consideration.  Specifically, the choice of using a node-like callback parameter on transport operations.  That callback is called by the transport when a specific operation (connect, disconnect, and send_event) is complete.  In the current (sync) implementation, that callback is just used to set an event that the main thread is waiting for.  In future (async/await) implementations, that callback can be used to complete a Future which is being awaited by the client code.

In order to make this work consistently, and in order to fix some problems that Pierre saw when trying to call send_event without connecting, I decided to simplify the state machine.  Since we are looking forward to async/await and want to support overlapped sends, I decided to remove the "sending" state.  This was made possible by the addition of the callback parameter, so each send_event operation can have its own callback, and there is no reason to leave the state machine in a "sending" state while we want for the acknowledgement from the service.  Now we have 4 states: disconnected, connecting, connected, and disconnecting.  For operations (like send_event) that happen while disconnected, we have an event queue that doesn't get processed until we're in the connected state.

I also realized that the Locked state machine does not behave the way I thought when dealing with overlapped transitions.  I thought that transitions would be queued if another transition was in progress, but it turns out that transitions are only blocked while another transition is in progress.  This is the root cause of Pierre's problem with send_event.  Part of the simplification allowed me to change the machine into a queued state machine.  This means that transitions will be queued if another transition is in progress.  This also means that we don't have to worry about what we do in before, after, on_enter, and on_exit functions because any transitions that are triggered inside those functions are queued rather than executed immediately.  

I'm happy that the unit tests didn't have to change much.  This makes me feel confident about these changes.  The samples are all updated and run correctly.  I'm updating the E2E tests as one final verification.


azure-iot-hub-devicesdk/README.md
```diff
@@ -25,30 +25,30 @@ Our samples rely on having a connection string for the device set in an environm
...
     print("connection status: " + status)
-    if (status == 'connected'):
-      device_client.disconnect()
 
 
 device_client.on_connection_state = connection_state_callback
 device_client.connect()
-
-input("Press Enter at any time to quit...\n\n")
+time.sleep(1)
```
pierreca: why do we need this?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
```diff
@@ -49,21 +50,37 @@ def _create_mqtt_client(self):
         def on_connect_callback(client, userdata, flags, result_code):
             logger.info("connected with result code: %s", str(result_code))
             # TODO: how to do failed connection?
-            self.on_mqtt_connected("connected")
+            try:
+                self.on_mqtt_connected()
+            except:  # noqa: E722 do not use bare 'except'
+                logger.error("Unexpected error calling on_mqtt_connected")
+                logger.error(traceback.format_exc())
```
pierreca: nit: seems like a pattern worth extracting in a helper function?

pierreca:
>         # 1. apt insatll graphviz

nit+unrelated: typo

---
Refers to: azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py:115 in 1a9f86e. [](commit_id = 1a9f86eb6c6edcff1981663c2c9e1c62adad01d1, deletion_comment = False)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -137,60 +121,89 @@ def _on_transition_complete(event_data):
...
-        self._event_queue.put_nowait(event_data.args[0])
+        logger.info("_on_provider_connect_complete")
+        self._trig_provider_connect_complete()
+
+        if self.on_transport_connected:
+            self.on_transport_connected("connected")
+        callback = self._connect_callback
+        if callback:
+            self._connect_callback = None
+            callback()
```
pierreca: for later: this pattern is akin to having a global state - it's not always bad, but it means we now have properties that have their state "tied" to the state machine but enforced by us. I wonder if there's a way to pass these around rather than store them. 
In any case, it's worth keeping an eye on how this "global state" grows over time. 2 of these is OK, if we start having 5 or 7 properties for which we need to ensure consistency, it becomes tricky.

pierreca: also, nitpick: there seems to be a pattern on how we handle callback - consider helper functions/classes? maybe only if it "grows"?

---
In reply to: [237660851](https://github.com/Azure/azure-iot-sdk-python-preview/pull/26#discussion_r237660851) [](ancestors = 237660851)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -137,60 +121,89 @@ def _on_transition_complete(event_data):
...
         self._mqtt_provider.disconnect()
 
-    def _before_action_add_event_to_queue(self, event_data):
+    def _on_provider_connect_complete(self):
         """
-        Queue an event for sending later.
-        This is meant to be called by the state machine as an "before" action
+        Callback that is called by the provider when the connection has been established
         """
-        # TODO: throw here if event_data.args[0] is not an event/message
```
pierreca: this TODO is disappearing - just checking we're not losing track of something?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -137,60 +121,89 @@ def _on_transition_complete(event_data):
...
+            callback()
+
+    def _add_event_to_queue(self, event_data):
+        """
+        Queue an event for sending later.  All events that get sent end up going into
+        this queue, even if they're going to be sent immediately.
+        """
+        logger.info("Adding event to queue for later sending")
+        # TODO: throw here if event_data.args[0] is not an event/message
+        self._event_queue.put_nowait((event_data.args[0], event_data.args[1]))
```
pierreca: >((event_data.args[0], event_data.args[1])) [](start = 36, length = 42)

args in arrays are going to be hard to track, nothing we can do about this?

BertKleewein: this comes from transitions.  They pack this for us.  I don't like this either. 

---
In reply to: [237662131](https://github.com/Azure/azure-iot-sdk-python-preview/pull/26#discussion_r237662131) [](ancestors = 237662131)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
BertKleewein: That's a good idea.  I could see value in adding some sort of context guard, like:
with log_and_swallow_all_exceptions():
    self.on_mqtt_disconnected("disconnected")

with the log_and_swallow guard doing the try/except big and the logging.  

I think this is possible, but context guards might not be the right way to do it.  I suspect a decorator might be better.  But I don't want to fall into the "learning a new python language feature" trap right now.


---
In reply to: [237658858](https://github.com/Azure/azure-iot-sdk-python-preview/pull/26#discussion_r237658858) [](ancestors = 237658858)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
BertKleewein: I agree, this became painfully obvious in the Node SDK.  I'm going to chew on this.  

---
In reply to: [237661215](https://github.com/Azure/azure-iot-sdk-python-preview/pull/26#discussion_r237661215) [](ancestors = 237661215,237660851)

PR #27, opened 2018-11-29
Message class with properties

olivakar:


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,53 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
BertKleewein: >Copyright [](start = 2, length = 9)

silly request: can you please squash before submitting PRs?  Ideally 1 commit per PR, or 2 if there's different features that are cleanly separated.  In codeflow, I see all 8 iterations at the top, and this one is simple so I only have to look at #8, but it gets more complicated when you add more iterations.  If your updated code (after feedback) ends up is at iteration 12 and your first is at iteration 8, I need to know this to see what you changed.  This can be a pain to figure out.  It's much easier if I can compare iteration 2 to iteration 1 to see what changed.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,53 @@
...
+
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class Message(object):
+    def __init__(
+        self, data, creation_time_utc, message_id=None, content_encoding=None, content_type=None
```
BertKleewein: >creation_time_utc [](start = 20, length = 17)

why is this required?  Can't the object figure this out by calling some OS function to return the current time?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,53 @@
...
+    def add_property(self, key, value):
+        """
+        Adds the key-value pair to the collection. The strings 'Authorization'
+        and 'iothub-to' are considered as reserved key names and will cause the
+        function to log error.
+        """
+        if key == "Authorization" or key == "iothub-to":  # Not sure of this. Was part of Node code.
+            logger.error(
+                "Cannot add these keys"
+            )  # Should we throw exception ? Should we return false ?
```
BertKleewein: ># Should we throw exception ? Should we return false ? [](start = 15, length = 54)

always throw on failure unless you have a really good reason.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Message(object):
+    def __init__(
+        self, data, creation_time_utc, message_id=None, content_encoding=None, content_type=None
+    ):
+        """
+        :param data: The  data that constitutes the payload
+        :param creation_time_utc: Date and time the message was created
+        :param message_id: A user-settable identifier for the message used for request-reply patterns.Format: A case-sensitive string (up to 128 characters long) of ASCII 7-bit alphanumeric characters + {'-', ':', '.', '+', '%', '_', '#', '*', '?', '!', '(', ')', ',', '=', '@', ';', '$', '''}.
```
BertKleewein: >message_id [](start = 15, length = 10)

do we need to add details on what gets used if nothing is provided?  I assume a message_id gets auto-generated, but what about encoding and type?  Can those be inferred from data?  e.g. if data is a UTF-8 string, can we just assume type and encoding without the user having to tell us?

azure-iot-hub-devicesdk/tests/test_message.py
```diff
@@ -0,0 +1,76 @@
...
+
+from azure.iot.hub.devicesdk.message import Message
+import pytest
+from datetime import datetime
+
+
+def test_construct_message_with_string():
+    s = "After all this time? Always"
+    time = datetime.utcnow()
+    mes = Message(s, time)
```
BertKleewein: >mes [](start = 4, length = 3)

Nit: This shortening of message reads funny to me.  I've always seen it as msg, which I see you use for a string below.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
olivakar: Sure. I generally squash before I am about to merge to master. I assume you speaking from the perspective of code flow. Yes, Its a bit cumbersome to see 8 iterations for 8 commits. From the PR on Github however it looks fine and only shows changes that are made.  

---
In reply to: [237660103](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237660103) [](ancestors = 237660103)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
pierreca: Also: please check if this is a property we should set, or if the service does that for us?

---
In reply to: [237660405](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237660405) [](ancestors = 237660405)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
pierreca: that's debatable, in the spirit of saving bandwidth we should only set properties that the customer wants (applies to creation-time if it's not mandatory too). 

---
In reply to: [237661641](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237661641) [](ancestors = 237661641)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,53 @@
...
+        :param content_encoding: Content encoding of the message data. can be 'utf-8', 'utf-16' or 'utf-32'.
+        :param content_type: Content type property used to routes with the message body. Should be 'application/json'.
```
pierreca: > Should be 'application/json'. [](start = 88, length = 30)

only if routing based on the body is used.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
pierreca: I believe Authorization is forbidden only in HTTP and iothub-to actually allowed on the service side, that could share the Message class. I believe it's premature to add those checks.

---
In reply to: [237661027](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237661027) [](ancestors = 237661027)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,53 @@
...
+        self.correlation_id = None
+        self.user_id = None
+        self.ack = None
+        self.content_encoding = content_encoding
+        self.content_type = content_type
+        self.device_id = None
+        self.auth_generation_id = None
+        self.auth_method = None
+
+    def add_property(self, key, value):
```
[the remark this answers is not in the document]
pierreca: why not simply let the user set custom_properties by accessing the dictionary directly? after all, it's public. this helper method just gives them another way to do it. Then in the transport, where there might be some limitation on property names, we throw about them when we translate them. 

---
In reply to: [237663623](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237663623) [](ancestors = 237663623)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
olivakar: I believe so. I included this as the Microsoft docs mention this to be explicitly set. On further thoughts I feel it means to be set from the device OS level , I shall delete this. 

---
In reply to: [237660405](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237660405) [](ancestors = 237660405)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
olivakar: message_id : from documentation it looks like it is user settable. So it should be possible to auto generate as well as user set from before.

Mostly the content type and encoding is always explicitly stated. Need to discuss more on this.

---
In reply to: [237661641](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237661641) [](ancestors = 237661641)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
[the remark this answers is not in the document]
olivakar: I am gone between the 2 names. I am fine if we want to call it `user_properties` .. C# calls it that but also C# also has another dictionary called `system_properties`. Node calls it just properties. Have changed the name to `add_custom_property`  Updated the documentation which explicitly can call out that we can add to some user defined properties. 

---
In reply to: [237663623](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237663623) [](ancestors = 237663623)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
olivakar: The exception as there already, Should have removed the comment

---
In reply to: [237661027](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237661027) [](ancestors = 237661027)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,53 @@
...
+        self.expiry_time_utc = None
+        self.enqueued_time = None
+        self.correlation_id = None
+        self.user_id = None
+        self.ack = None
+        self.content_encoding = content_encoding
+        self.content_type = content_type
+        self.device_id = None
+        self.auth_generation_id = None
+        self.auth_method = None
```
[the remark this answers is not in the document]
olivakar: deleted as I could not find usage of them in other SDKs too.

---
In reply to: [237664629](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237664629) [](ancestors = 237664629)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
olivakar: message id has been changed to `self.message_id = message_id if message_id else uuid.uuid4()`

---
In reply to: [237723135](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237723135) [](ancestors = 237723135,237661641)

pierreca: why would we populate it ourselves? it's just more bandwidth. if not provided, leave blank. 

---
In reply to: [238907119](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r238907119) [](ancestors = 238907119,237723135,237661641)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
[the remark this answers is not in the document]
pierreca: what's the value of adding a method instead of accessing the (public) dictionary object?

---
In reply to: [237723191](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r237723191) [](ancestors = 237723191,237663623)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
```diff
@@ -0,0 +1,43 @@
...
+        self.content_type = content_type
+
+    def add_custom_property(self, key, value):
+        """
+        Adds the key-value pair to the collection of user defined properties. The strings 'Authorization'
+        and 'iothub-to' are considered as reserved key names and will cause the
+        function to log error.
+        """
+        if key == "Authorization" or key == "iothub-to":  # Not sure of this. Was part of Node code.
+            raise TypeError("The keys 'iothub-to' or 'Authorization' are not allowed")
```
pierreca: it's not clear to me that this is needed - it's transport specific or use-case specific if anything. you should either remove it or run it by the service and confirm that this is the right place because it happens on all protocols in all messages.
if we end up removing it, i don't see the value-add anymore of the whole add_custom_property method over accessing the dictionary directly.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
olivakar: i thought if not provided we should auto generate the id. But I am seeing no code in Node that does that. In C# there is a code but it is never called as far as I can see. I am going to revert this change to before.

---
In reply to: [238907590](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r238907590) [](ancestors = 238907590,238907119,237723135,237661641)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
olivakar: there is a requirement in Node that says `SRS_NODE_IOTHUB_PROPERTIES_07_004: [** If `itemKey` contains any of the reserved key names then the `add` function will return `false`. **]`. 

---
In reply to: [238908248](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r238908248) [](ancestors = 238908248)

pierreca: yep and I think this is premature right now. it's protocol-specific and MQTT doesn't force any of that. we'll add this if/when we need this

---
In reply to: [239227272](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r239227272) [](ancestors = 239227272,238908248)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message.py
pierreca: thanks. we should *NOT* autogenerate it

---
In reply to: [239226953](https://github.com/Azure/azure-iot-sdk-python-preview/pull/27#discussion_r239226953) [](ancestors = 239226953,238907590,238907119,237723135,237661641)

PR #28, opened 2018-11-30
POC for adding _async wrappers for InternalClient functions

BertKleewein:
This is one way we could create async wrappers for our code.  I like this because:
1) All transport calls run in a background thread, so we never block the main event loop.
2) transport callbacks signal the awaiting thread when operations are done by using a Future object.  This way the main event loop isn't blocked waiting operations to complete, and we don't need to make a thread that waits for our operation to complete.  
3) async methods are added to InternalClient with _async suffixes.  We could also make InternalClientAsync to hold the async functions, but that seems like it could explode as we add DeviceClientAsync and ModuleClientAsync, and I'm sure we would need to add more.  

I've tested this with sending 1000 events in parallel and it seems to do as well as we want.  I've also tested that we can still use the client object with Python 2.7 because the code that adds the _async implementations is protected by try/except.


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
```diff
@@ -0,0 +1,95 @@
...
+    This is just a test docstring to make sure we can see documentation for these functions
+    when we call help(InternalClient), and we can, so life is good and everyone is happy.
+    """
+    logger.info("async connecting to transport")
+
+    # we need the event loop so we can run our code in side it.
+    loop = asyncio.get_event_loop()
+
+    # the future is how we communicate the result from the callback task back into the "main" task
+    future = asyncio.Future()
```
olivakar: >future = asyncio.Future() [](start = 4, length = 25)

Is it not possible to use `loop.create_future()` ?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
```diff
@@ -0,0 +1,95 @@
...
+
+async def _connect_async(self):
+    """
+    This is just a test docstring to make sure we can see documentation for these functions
+    when we call help(InternalClient), and we can, so life is good and everyone is happy.
+    """
+    logger.info("async connecting to transport")
+
+    # we need the event loop so we can run our code in side it.
+    loop = asyncio.get_event_loop()
```
olivakar: >loop = asyncio.get_event_loop() [](start = 4, length = 31)

What happens when we use `get_running_loop()` ?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
BertKleewein: that wasn't added until 3.7.  Do you know if there's any advantage to using that or disadvantage to doing it the way that I did?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
olivakar: okay. it is written in 3.7 as to being preferable to use. The reason being cited is `Because this function has rather complex behavior (especially when custom event loop policies are in use), using the get_running_loop() function is preferred to get_event_loop() in coroutines`. I am thinking this one does not return the `currently running event loop` but returns the current thread bound one though I am not exactly sure of the implication. 

---
In reply to: [238760013](https://github.com/Azure/azure-iot-sdk-python-preview/pull/28#discussion_r238760013) [](ancestors = 238760013)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
olivakar: No particular advantage. 

---
In reply to: [238758002](https://github.com/Azure/azure-iot-sdk-python-preview/pull/28#discussion_r238758002) [](ancestors = 238758002)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
BertKleewein: That's really good info.  I'm reading that it works better in complex situations, so maybe we should have the code use get_running_loop if it can, and fallback to get_event_loop.  Hopefully people with complex async/await setups are using 3.7, so this would satisfy them while still working downlevel.

azure-iot-hub-devicesdk/samples/async_send_telemetry_device.py
```diff
@@ -0,0 +1,54 @@
...
+    # ... and connect the client.
+    await device_client.connect_async()
+
+    async def send_test_message_async(i):
+        print("sending message #" + str(i))
+        await device_client.send_event_async("test_payload message " + str(i))
+        print("done sending message #" + str(i))
+
+    # send `messages_to_send` messages in parallel
+    await asyncio.wait([send_test_message_async(i) for i in range(1, messages_to_send + 1)])
```
olivakar: >asyncio.wait [](start = 10, length = 12)

it seems `asyncio.gather` will do the same thing.  The different being that `asyncio.wait` has a return_when property that can be set as an condition.  Any reason for the wait pattern ?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/internal_client.py
```diff
@@ -136,3 +136,22 @@ def from_authentication_provider(cls, authentication_provider, transport_name):
...
+
+# This is where we add our _async functions to InternalClient.  We do this at runtime
+# by adding them to the InternalClient prototype, using what the python community calls
+# "monkeypatching".  If we're running on python 2.7, the import will throw a SyntaxError
+# and we'll just carry on without the _async functions.
+#
+# Don't worry, adding them to InternalClient at runtime will also add them to DeviceClient
+# and ModuleClient.  Python has a module cache like node, so the first time anyone imports
+# internal_client.py (to use it for a base class, for instance), it will monkeypatch it here
+# and use the patched class definition for anyone else who imports internal_client.py
```
pierreca: As soon as possible please sync up with Oliva on the docs generation mechanism to see how that looks like in the docs

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
pierreca: suggestion: typically a good question to ask on the Python teams channel

---
In reply to: [238881039](https://github.com/Azure/azure-iot-sdk-python-preview/pull/28#discussion_r238881039) [](ancestors = 238881039)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_internal_client.py
```diff
@@ -0,0 +1,95 @@
...
+    logger.info("async sending event")
+    loop = asyncio.get_event_loop()
+    future = asyncio.Future()
+
+    def callback():
+        logger.info("async sending finished")
+        loop.call_soon_threadsafe(future.set_result, None)
+
+    await loop.run_in_executor(None, self._transport.send_event, event, callback)
+    await future
```
pierreca: i'm going to make the same comment i always do.. this looks like a pattern ^^
maybe a bit early to refactor but something to track as APIs multiply.

PR #29, opened 2018-12-03
feat(device): add x509 authentication

pierreca:
First stab at adding X509 authentication for devices.

Please start the review with the sample called `x509_send_telemetry.py`

Things I need help with:
- decide whether from_connection_string should support X509 connection strings
  - if Yes: how do we pass the cert/key later? other SDKs do set_options, everybody hates that.
  - if No: well, it's still a valid connection string... so not ideal
- The structure of the x509 properties inside the x509 authentication provider. Not sure why we'd need to encapsulate cert_file and key_file in the X509Credentials class but I liked having everything under the same structure. not sure why. maybe because of some node SDK influence
- I have not tested with properly signed device certs, as well as cert chains. those should behave the same though, code would not change.
- The tests for the AuthenticationProvider constructor and AuthenticationProvider factories are exactly the same - not sure I enjoy the duplication

Other things to note:
- I added a couple of unrelated tests, it's not a mistake/left-over
- I had to add a `__init__.py` file in the test folder for VS Code to properly detect tests and show the Run/Debug buttons in the code-lens. Anyone made it work without it?

Have tested only on python 3.7 so far.

### DIFF
--- a/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
+++ b/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
@@ -96,12 +96,10 @@ class MQTTProvider(object):
 
         logger.info("Created MQTT provider, assigned callbacks")
 
-    def connect(self, password):
+    def _create_ssl_context(self, x509):
         """
-        This method connects the upper transport layer to the mqtt broker.
-        This method should be called as an entry point before sending any telemetry.
+        This method creates the SSLContext object used by Paho to authenticate the connection
         """
-        logger.info("connecting to mqtt broker")
 
         ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
         if self._ca_cert:
@@ -110,6 +108,21 @@ class MQTTProvider(object):
             ssl_context.load_default_certs()
         ssl_context.verify_mode = ssl.CERT_REQUIRED
         ssl_context.check_hostname = True
+
+        if x509 is not None:
+            logger.info("configuring SSL context with client-side certificate and key")
+            ssl_context.load_cert_chain(x509.cert_file, x509.key_file, x509.passphrase)
+        return ssl_context
+
+    def connect(self, password=None, x509=None):
+        """
+        This method connects the upper transport layer to the mqtt broker.
+        This method should be called as an entry point before sending any telemetry.
+        """
+        logger.info("connecting to mqtt broker")
+
+        ssl_context = self._create_ssl_context(x509)
+
         self._mqtt_client.tls_set_context(ssl_context)
         self._mqtt_client.tls_insecure_set(False)
         self._mqtt_client.username_pw_set(username=self._username, password=password)
@@ -117,12 +130,15 @@ class MQTTProvider(object):
         self._mqtt_client.connect(host=self._hostname, port=8883)
         self._mqtt_client.loop_start()
 
-    def reconnect(self, password):
+    def reconnect(self, password=None, x509_credentials=None):
         """
         This method reconnects the mqtt broker, possibly because of a password (sas) change
         Connect should have previously been called.
         """
         logger.info("reconnecting transport")
+
+        ssl_context = self._create_ssl_context(x509_credentials)
+        self._mqtt_client.tls_set_context(ssl_context)
         self._mqtt_client.username_pw_set(username=self._username, password=password)
         self._mqtt_client.reconnect()
 
--- a/azure-iot-hub-devicesdk/samples/authentication_providers.py
+++ b/azure-iot-hub-devicesdk/samples/authentication_providers.py
@@ -9,6 +9,7 @@ from azure.iot.hub.devicesdk.auth.authentication_provider_factory import (
     from_shared_access_signature,
     from_connection_string,
     from_environment,
+    from_x509,
 )
 
 # To understand authentication providers, it is beneficial to be familiar with the Azure IoT Hub Security model:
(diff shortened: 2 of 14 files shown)


scripts/gen_self_signed_cert.py
```diff
@@ -0,0 +1,63 @@
...
+cert = x509.CertificateBuilder().subject_name(
+    subject
+).issuer_name(
+    issuer
+).public_key(
+    key.public_key()
+).serial_number(
+    x509.random_serial_number()
+).not_valid_before(
+    datetime.utcnow()
```
BertKleewein: Just out of curiosity, have you verified that this works?  If so, I am very happy.
As a broader question, does the CertificateBuilder object have a dependency on the openssl executable like Node does?  Please say that it doesn't.

azure-iot-hub-devicesdk/tests/test_mqtt_transport.py
```diff
@@ -5,11 +5,19 @@
...
+    MQTTProvider
+)
+from azure.iot.hub.devicesdk.auth.authentication_provider_factory import (
+    from_connection_string,
+    from_x509
+)
+import azure.iot.hub.devicesdk.auth.sk_authentication_provider as sk_auth
+import azure.iot.hub.devicesdk.auth.x509_authentication_provider as x509_auth
 from six import add_move, MovedModule
-from six.moves import mock
```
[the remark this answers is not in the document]
pierreca: something was complaining about it. can't remember but figured i'd find out soon enough if it broke something.

---
In reply to: [238394107](https://github.com/Azure/azure-iot-sdk-python-preview/pull/29#discussion_r238394107) [](ancestors = 238394107)

scripts/gen_self_signed_cert.py
pierreca: I have verified this. I could send messages :)
WRT to openssl: the package does have references to openssl headers and says this in the FAQ:
Installing cryptography with OpenSSL 0.9.8 or 1.0.0 fails
The OpenSSL project has dropped support for the 0.9.8 and 1.0.0 release series. Since they are no longer receiving security patches from upstream, cryptography is also dropping support for them. To fix this issue you should upgrade to a newer version of OpenSSL (1.0.2 or later). This may require you to upgrade to a newer operating system.

So it may need at least the library - not sure about the executable themselves...

---
In reply to: [238395103](https://github.com/Azure/azure-iot-sdk-python-preview/pull/29#discussion_r238395103) [](ancestors = 238395103)

azure-iot-hub-devicesdk/tests/test_mqtt_transport.py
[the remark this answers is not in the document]
pierreca: I think it's related to #31

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
```diff
@@ -17,7 +17,7 @@ class MQTTProvider(object):
     to publish/subscribe messages.
     """
 
-    def __init__(self, client_id, hostname, username, password, ca_cert=None):
+    def __init__(self, client_id, hostname, username, password, x509_info = None, ca_cert=None):
```
BertKleewein: >x509_info = None [](start = 64, length = 16)

Should the x509_info parameter be on the connect function because that's where it's used?

For SAS renewal, I moved the password to connect (and reconnect) because that makes the upper layer more logical and saves us from caching the credentials down here.  

FWIW, I think the hostname and username should maybe also move to the connect method for the same reason, but I wanted to limit my delta.

azure-iot-hub-devicesdk/samples/authentication_providers.py
```diff
@@ -27,6 +28,13 @@
 # - Pre-defined environment variables (this is especially useful when running as an Azure IoT Edge module)
 env_auth_provider = from_environment()
 
+# - X509 Certificate Authentication Provider'
+cert_file = "../scripts/certificate.pem"
+key_file = "../scripts/key.pem"
```
BertKleewein: >"../scripts/key.pem" [](start = 11, length = 20)

Recommend a comment on how to create these files.  I assume you just have to run ../scripts/gen_self_signed_cert.py, but I'd rather know for sure.

azure-iot-hub-devicesdk/samples/x509_send_telemetry.py
```diff
@@ -0,0 +1,40 @@
...
+from azure.iot.hub.devicesdk.auth.authentication_provider_factory import from_x509
+
+
+# X509 certificate and key paths, and optional passphrase
+cert_file = "./scripts/certificate.pem"
+key_file = "./scripts/key.pem"
+passphrase = "passphrase"
+
+# The "Authentication Provider" will hold references to the device identity and certificate information and provide them to the transport when necessary
+x509_auth_provider = from_x509(os.getenv("DEVICE_ID"), os.getenv("HOSTNAME"), cert_file, key_file, passphrase)
```
BertKleewein: >DEVICE_ID [](start = 42, length = 9)

Let's standardize on environment variable names.  We use IOTHUB_DEVICE_ and IOTHUB_MODULE_ as prefixes in the other samples and we should be consistent.

azure-iot-hub-devicesdk/tests/test_authentication_provider_factory.py
```diff
@@ -6,14 +6,18 @@
 from azure.iot.hub.devicesdk.auth.authentication_provider_factory import (
     from_connection_string,
     from_shared_access_signature,
+    from_x509
 )
```
BertKleewein: out of scope: these names should be changed.  auth_provider = from_x509() sounds wrong.  It should be one of:
auth_provider = AuthorizationProvider.from_x509() 
or
auth_provider = authorization_provider_from_x509()

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
BertKleewein: >x509_info [](start = 64, length = 9)

Info is a horrible name.  It smells like you don't know what this object is.  for consistency, I think x509_info should maybe be x509_credentials.

scripts/gen_self_signed_cert.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
```
BertKleewein: >python [](start = 15, length = 6)

to lay it on heavy, should this be "gen_self_signed_cert_for_testing_do_not_use_for_production.py"?

azure-iot-hub-devicesdk/tests/test_mqtt_transport.py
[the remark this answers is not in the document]
BertKleewein: It's exactly the same as #31.  You'll have to merge that in before you push to master.

---
In reply to: [239682669](https://github.com/Azure/azure-iot-sdk-python-preview/pull/29#discussion_r239682669) [](ancestors = 239682669)

pierreca:
completely outdated - let's forget about this one...

PR #30, opened 2018-12-06
Connection String Empty

olivakar:


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/sas_authentication_provider.py
```diff
@@ -57,6 +57,8 @@ def parse(sas_token_str):
         SharedAccessSignature sr=<resource_uri>&sig=<signature>&skn=<keyname>&se=<expiry>
         :return: The Shared Access Signature Authentication Provider constructed
         """
+        if not sas_token_str or not sas_token_str.strip():
+            raise ValueError("Shared Access Signature string - Empty or Blank")
```
BertKleewein: >Shared Access Signature string - Empty or Blank [](start = 30, length = 47)

Suggest saying "shared access string is required", otherwise it's not completely clear what the function expects.

PR #33, opened 2018-12-07
add sas token renewal logic for shared key auth and iotedge auth

BertKleewein:
This adds code for the shared key and iotedge auth providers to automatically re-generate their sas tokens based on expiry timers.  In order to do this, several things needed to happen:
1) The MQTT provider needed to have a reconnect() function.  This is how the upper layers send a new SAS token down to PAHO.
2) The MQTT transport needed to have a callback function where the auth provider could notify the
transport that a new SAS token is available.  Since this operation requires the transport to reconnect, I needed to add a new trigger to the state machine.  There is currently no need to add a "reconnecting" state since "reconnecting" would behave exactly like "connecting".
3) A timer needed to be added to the auth providers so they could automatically renew themselves, and a callback needed to be defined so they could notify the transport when this renewal happens.  
4) A "disconnect" function needed to be added to the auth providers so they can cancel their update timers.  This is necessary because otherwise, a timer would prevent the auth object from being garbage collected at the right time.
5) The transport disconnect function needs to call disconnect on the auth provider so the application can shut down after disconnect is called.  Tests also need to call disconnect on the auth provider for the same reason.


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/base_renewable_token_authentication_provider.py
```diff
@@ -81,14 +96,108 @@ def generate_new_sas_token(self):
...
+        definitions of what "current time" is, so the margin needs to be set so there is a
+        very small chance that there is no time overlap where one computer thinks the token
+        is expired and another doesn't.
+
+        When the timer is set to renew the SAS token, the timer is set for
+        (token_renewal_interval - token-renewal_margin) seconds in the future.  In this way,
+        the token will be renewed close to it's expiration time, but not so close that
+        we risk a problem caused by clock drift.
+
+        :param token_renewal_interval: The number of seconds, in the future, that
```
pierreca: >token_renewal_interval [](start = 15, length = 22)

nit: that sounds more like a token_validity_period or token_validity_interval and then token_renewal_interval = token_valid_interval - token_renewal_margin.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/base_renewable_token_authentication_provider.py
```diff
@@ -81,14 +96,108 @@ def generate_new_sas_token(self):
             token = _device_token_format.format(quoted_resource_uri, signature, str(expiry))
 
         self.sas_token_str = str(token)
+        self._schedule_token_update(self.token_renewal_interval - self.token_renewal_margin)
+        self._notify_token_updated()
+
+    def set_token_renewal_interval(self, token_renewal_interval, token_renewal_margin):
```
pierreca: >set_token_renewal_interval [](start = 8, length = 26)

not sure how much value this brings compared to editing the two attributes directly? they seem to be public and I seem to recall getters/setters aren't really a thing in python? is the only benefit to have logging?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/base_renewable_token_authentication_provider.py
```diff
@@ -81,14 +96,108 @@ def generate_new_sas_token(self):
...
+    def _cancel_token_update_timer(self):
+        """
+        Cancel any future token update operations.  This is typically done as part of a
+        teardown operation
+        """
+        t = self._token_update_timer
+        self._token_update_timer = None
+        if t:
+            logger.info("Canceling token update timer for (%s,%s)", self.device_id, self.module_id)
+            t.cancel()
```
pierreca: >t.cancel() [](start = 12, length = 10)

can this raise an exception if the timer has already been cancelled? should we catch it?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/base_renewable_token_authentication_provider.py
```diff
@@ -81,14 +96,108 @@ def generate_new_sas_token(self):
...
+
+    def _schedule_token_update(self, seconds_until_update):
+        """
+        Schedule an automatic sas token update to take place seconds_until_update seconds in
+        the future.  If an update was previously scheduled, this method shall cancel the
+        previously-scheduled update and schedule a new update.
+        """
+        self._cancel_token_update_timer()
+        logger.info(
+            "Scheduleing token update for (%s,%s) for %d seconds in the future",
```
pierreca: >Scheduleing [](start = 13, length = 11)

nit: typo (Scheduling)

azure-iot-hub-devicesdk/tests/test_base_renewable_token_authentication_provider.py
```diff
@@ -0,0 +1,176 @@
...
+    device_auth_provider, fake_timer_object
+):
+    device_auth_provider.generate_new_sas_token()
+    assert (
+        fake_timer_object.call_args[0][0]
+        == DEFAULT_TOKEN_RENEWAL_INTERVAL - DEFAULT_TOKEN_RENEWAL_MARGIN
+    )
+
+
+def test_genrate_new_sas_token_cancels_and_reschedules_update_timer_with_correct_modified_timeout(
```
pierreca: >test_genrate_new_sas_token_cancels_and_reschedules_update_timer_with_correct_modified_timeout [](start = 4, length = 93)

nit: typo (generate)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -86,6 +89,17 @@ def __init__(self, auth_provider):
...
+                "trigger": "_trig_on_shared_access_string_updated",
+                "source": "connected",
+                "dest": "connecting",
+                "after": "_call_provider_reconnect",
+            },
+            {
+                "trigger": "_trig_on_shared_access_string_updated",
+                "source": ["disconnected", "disconnecting"],
+                "dest": None,
+            },
```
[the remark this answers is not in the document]
BertKleewein: This is a really good question.  I think we want to queue a reconnect, and I think that might be easy, but I need to chew on this.

---
In reply to: [240406093](https://github.com/Azure/azure-iot-sdk-python-preview/pull/33#discussion_r240406093) [](ancestors = 240406093)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/base_renewable_token_authentication_provider.py
BertKleewein: It does not throw if you cancel a timer that hasn't been started or if you cancel a timer that's already been cancelled.  Either way, the code here should protect against double-cancel.


---
In reply to: [240404496](https://github.com/Azure/azure-iot-sdk-python-preview/pull/33#discussion_r240404496) [](ancestors = 240404496)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/base_renewable_token_authentication_provider.py
BertKleewein: Not much value.  I previously had this function with the side-effect of renewing the token, but I wanted that to be manual.  I'm going to remove this and just do attributes.

---
In reply to: [240404233](https://github.com/Azure/azure-iot-sdk-python-preview/pull/33#discussion_r240404233) [](ancestors = 240404233)

PR #34, opened 2018-12-10
Encoded properties on MQTT with message class

olivakar:
- Tests are updated
- Sample updated using message
- Removed six moves mock from 1 file (this change is unrelated to the actual change)


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -245,6 +244,48 @@ def _get_telemetry_topic(self):
...
+            system_properties["$.ce"] = message_to_send.content_encoding
+
+        if message_to_send.expiry_time_utc:
+            system_properties["$.exp"] = (
+                message_to_send.expiry_time_utc.isoformat()
+                if isinstance(message_to_send.expiry_time_utc, date)
+                else message_to_send.expiry_time_utc
+            )
+
+        # Question TODO : haven't encoded these properties here, i see it encoded in JAVA
```
pierreca: >Question TODO [](start = 10, length = 13)

unless you intend to solve this before checking in, please have a work item on the task board to track this down.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -245,6 +244,48 @@ def _get_telemetry_topic(self):
...
+        # system_properties["$.cdid"]
+        # system_properties["$.cmid"]
+        if system_properties and len(system_properties) > 0:
```
pierreca: >system_properties [](start = 11, length = 17)

you initialize this yourself on line 248. it's always going to be truthy, right?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -245,6 +244,48 @@ def _get_telemetry_topic(self):
...
+            system_properties_encoded = urllib.parse.urlencode(system_properties)
+            topic += system_properties_encoded
+
+        topic += "&"
```
pierreca: >topic += "&" [](start = 8, length = 12)

isn't this needed ONLY if you have custom properties?

azure-iot-hub-devicesdk/samples/send_telemetry_device.py
```diff
@@ -30,7 +32,13 @@ def connection_state_callback(status):
 # send 5 messages with a 1 second pause between each message
 for i in range(0, 5):
     print("sending message #" + str(i))
-    device_client.send_event("test_payload message " + str(i))
+    msg = Message("test weather payload " + str(i))
+    msg.message_id = uuid.uuid4()
+    msg.user_id = "AZIOT1234"
```
pierreca: >msg.user_id = "AZIOT1234" [](start = 4, length = 25)

nit: I don't think any other SDK uses the user_id property and i'm worried it might be confusing to someone reading the samples (given we also talk of auth providers etc). let's document the property, but not advertise it in samples.

requirements.txt
```diff
@@ -5,7 +5,6 @@ pytest-cov==2.6.0
 wheel==0.32.1
 paho-mqtt==1.4.0
 transitions==0.6.8
-enum34==1.1.6
 flake8
```
pierreca: just double-checking that this is intentional.

azure-iot-hub-devicesdk/tests/test_mqtt_transport.py
```diff
@@ -17,7 +20,61 @@
...
+
+message_id = "spell-1234"
+user_id = "HG749134"
+
+
+def create_fake_message(telemtry_str):
+    msg = Message(telemtry_str)
+    msg.message_id = "spell-1234"
+    msg.user_id = "HG749134"
+    msg.custom_properties["name"] = "hermoinee granger"
```
pierreca: >hermoinee [](start = 37, length = 9)

nit: i'm no harry potter fan but I'm pretty sure there's a typo here

azure-iot-hub-devicesdk/tests/test_mqtt_transport.py
```diff
@@ -17,7 +20,61 @@
...
+    msg.custom_properties["school"] = "hogwarts"
+    msg.custom_properties["birthstatus"] = "muggle"
+    msg.custom_properties["profession"] = "witch"
+    return msg
+
+
+def create_encoded_fake_topic(fake_message):
```
pierreca: >create_encoded_fake_topic [](start = 4, length = 25)

that looks like a duplication of the code we have in the transport. meaning if we have a bug there, we have it here.

I'd much rather see code that deconstructs the topic (splits, uri decoding) based on what we expect and check that we can parse the properties the way the service would. sort of "reverse" our encoding algorithm and verify we fall back on our feet, if you will

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -196,13 +195,13 @@ def _publish_events_in_queue(self, event_data):
...
-                (event_to_send, callback) = self._event_queue.get_nowait()
+                (message_to_send, callback) = self._event_queue.get_nowait()
             except queue.Empty:
                 logger.info("done checking queue")
                 return
             logger.info("retrieved event from queue. publishing.")
-            topic = self._get_telemetry_topic()
-            mid = self._mqtt_provider.publish(topic, event_to_send)
+            encoded_topic = self._encode_properties(message_to_send, self.topic)
+            mid = self._mqtt_provider.publish(encoded_topic, message_to_send.data)
```
BertKleewein: question for the masses: node, at least, doesn't require a message object. to wrap the data.  What if the user just passes down a dictionary or a string?  Do we fail and make then wrap it in a Message, or do we send it as-is?

azure-iot-hub-devicesdk/samples/send_telemetry_device.py
```diff
@@ -30,7 +32,13 @@ def connection_state_callback(status):
...
+    msg.custom_properties["temp"] = "71.89"
+    msg.custom_properties["humidity"] = "98%"
+    msg.custom_properties["windy"] = "yes"
```
BertKleewein: Is this a good example of what custom properties might be used for?  This seems more like something we traditionally put into the body of the message.

azure-iot-hub-devicesdk/tests/test_message.py
```diff
@@ -4,7 +4,8 @@
 # --------------------------------------------------------------------------------------------
 
 from azure.iot.hub.devicesdk.message import Message
-import pytest
+import six.moves.urllib as urllib
+import datetime
```
BertKleewein: unused

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
olivakar: we should probably wrap in a message and then send. 

---
In reply to: [240783586](https://github.com/Azure/azure-iot-sdk-python-preview/pull/34#discussion_r240783586) [](ancestors = 240783586)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
olivakar: [https://msazure.visualstudio.com/One/_workitems/edit/3605257](https://msazure.visualstudio.com/One/_workitems/edit/3605257)

---
In reply to: [240413057](https://github.com/Azure/azure-iot-sdk-python-preview/pull/34#discussion_r240413057) [](ancestors = 240413057)

azure-iot-hub-devicesdk/tests/test_mqtt_transport.py
olivakar: I am not sure what you mean by the reverse in this context since I need the topic to be created for an assertion. This was just a helper. I have created the topic without the code now.

---
In reply to: [240414779](https://github.com/Azure/azure-iot-sdk-python-preview/pull/34#discussion_r240414779) [](ancestors = 240414779)

requirements.txt
olivakar: yup

---
In reply to: [240414068](https://github.com/Azure/azure-iot-sdk-python-preview/pull/34#discussion_r240414068) [](ancestors = 240414068)

PR #35, opened 2018-12-21
send to output in module client

olivakar:


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/module_client.py
```diff
@@ -4,7 +4,19 @@
...
-    pass
+    def send_to_output(self, message, output_name):
+        """
+        Sends an event/message to the given module output.
+        This is an outgoing events and  means to be "output events"
+        :param output_name: Name of the output to send the event to.
+        :param message: message to send to the given output
+        """
+        if not isinstance(message, Message):
+            message = Message(message)
```
BertKleewein: >message = Message(message) [](start = 12, length = 26)

If send_to_output does this at the client level, then internal_client.send_event should do this at the client level for symmetry sake.  (I think we have code at a lower level that does the same wrapping when internal_client.send_event is called)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/module_client.py
```diff
@@ -4,7 +4,19 @@
...
+
+        message.output_name = output_name
+        self.send_event(message)
```
BertKleewein: >self.send_event(message) [](start = 8, length = 24)

I think this is a great reuse of code and I like how you leveraged the existing code.  I recommend finding a way to do this without calling self.send_event, maybe calling transport.send_event from inside here, or maybe make a private function like "_send_event_common" that is called by both public functions  I'm not sure if there's a coding "principle" that speaks against public functions that call other public functions, but this smells like bad layering.    It's important to have one set of functions that only gets called by customers (the public API) and a different set of functions that only get called by internal code (private functions).  This doesn't matter too much now because send_to_output does a little bit more than send_event, but it will matter more later if send_event needs to do something that send_to_output doesn't.

azure-iot-hub-devicesdk/samples/send_to_output_module.py
```diff
@@ -0,0 +1,50 @@
...
+
+# send 5 messages with a 1 second pause between each message
+for i in range(0, 5):
+    print("sending message #" + str(i))
+    msg = Message("test wind speed " + str(i))
+    msg.message_id = uuid.uuid4()
+    msg.correlation_id = "correlation-1234"
+    msg.custom_properties["tornado-warning"] = "yes"
+    module_client.send_to_output(msg, "twister")
+    time.sleep(1)
```
BertKleewein: this block confuses the sample.  If you want to show how these properties work, maybe we should have a property sample.

azure-iot-hub-devicesdk/samples/send_to_output_module.py
```diff
@@ -0,0 +1,50 @@
...
+import time
+from azure.iot.hub.devicesdk.module_client import ModuleClient
+from azure.iot.hub.devicesdk.auth.authentication_provider_factory import from_connection_string
+from azure.iot.hub.devicesdk.message import Message
+import uuid
+
+# The connection string for a device should never be stored in code. For the sake of simplicity we're using an environment variable here.
+conn_str = os.getenv("IOTHUB_MODULE_CONNECTION_STRING")
+# The "Authentication Provider" is the object in charge of creating authentication "tokens" for the device client.
+auth_provider = from_connection_string(conn_str)
```
BertKleewein: for modules, from_environment() is the factory method that you should use.  No connection string is necessary.

azure-iot-hub-devicesdk/samples/send_to_output_module.py
```diff
@@ -0,0 +1,50 @@
...
+# For now, the SDK only supports MQTT as a protocol. the client object is used to interact with your Azure IoT hub.
+# It needs an Authentication Provider to secure the communication with the hub, using either tokens or x509 certificates
```
BertKleewein: This sample does not interact with IoTHub.  It interacts with IoTEdge and it only works with IotEdgeAuthenticationProvider, though I'm not sure how deeply you want to explain that here.

azure-iot-hub-devicesdk/samples/send_to_output_module.py
```diff
@@ -0,0 +1,50 @@
...
+module_client = ModuleClient.from_authentication_provider(auth_provider, "mqtt")
+
+
+# The connection state callback allows us to detect when the client is connected and disconnected:
+def connection_state_callback(status):
+    print("connection status: " + status)
+
+
+# Register the connection state callback with the client...
+module_client.on_connection_state = connection_state_callback
```
BertKleewein: useful information, but it's more code to confuse the sample.  Having a connection status callback is not a requirement for calling send_to_output.

azure-iot-hub-devicesdk/tests/test_module_client.py
```diff
@@ -0,0 +1,83 @@
...
+    client.connect()
+    client.send_to_output(event, output_name)
+
+    assert event.output_name == output_name
+    assert mock_transport.send_event.call_count == 1
+    assert mock_transport.send_event.call_args[0][0] == event
+
+
+@patch("azure.iot.hub.devicesdk.module_client.isinstance")
+@patch("azure.iot.hub.devicesdk.module_client.Message")
```
BertKleewein: this looks very cool.   It looks like this works the same as @patch.object() only I think it's just patching the Message instance that was imported into module_client.py.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/module_client.py
BertKleewein: Comment #2: all other functions (so far) have a 1:1 mapping from client function name to transport function name.  I don't think that's a requirement, but it seems strange that client.connect calls transport.connect, client.disconnect calls transport.disconnect, client.send_event calls transport.send_event, and (the one exception) client.send_to_output calls transport.send_event.  By having the transport functions be exactly the same name as the client functions, we're implying a relationship, but this code breaks that relationship.  If the transport function handles telemetry and output events both, that OK, but calling it send_event implies that it only handles telemetry.  Maybe we should rename the transport function since the 1:1 mapping doesn't exist here.  

---
In reply to: [246254214](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246254214) [](ancestors = 246254214)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/module_client.py
olivakar: yes there is a code to wrap it in message before encoding the properties. The reason the wrapping was needed in module client is because we are already accessing the output name here. We should have symmetry and put it in internal client as well. If that's the case then we will no longer need it in transport. 

---
In reply to: [246253488](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246253488) [](ancestors = 246253488)

azure-iot-hub-devicesdk/samples/send_to_output_module.py
olivakar: I actually thought that this is one of the options. When I see in node I see that module client has been constructed with other type of authentication providers specially in the sample it has been created with `ModuleClient.fromConnectionString(connectionString, Protocol);`

---
In reply to: [246254580](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246254580) [](ancestors = 246254580)

azure-iot-hub-devicesdk/samples/send_to_output_module.py
olivakar: yes we have discussed this before but this is almost similar to the device sample. The device sample does have a connection state callback. Is it more useful to know this for a device and not the module ?

---
In reply to: [246255057](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246255057) [](ancestors = 246255057)

azure-iot-hub-devicesdk/tests/test_module_client.py
olivakar: yes i was about to say this yesterday , but i forgot. And i think this is similar to `@patch.object(azure.iot.hub.devicesdk.module_client, "Message")`

---
In reply to: [246255965](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246255965) [](ancestors = 246255965)

azure-iot-hub-devicesdk/tests/test_module_client.py
```diff
@@ -0,0 +1,83 @@
...
+def test_module_client_send_string_constructs_message_assigns_output_name_and_calls_transport(
+    mock_message_constructor, mock_instance_method, authentication_provider, mock_transport
+):
+    mock_instance_method.return_value = False
```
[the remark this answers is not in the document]
olivakar: It needs to be done as the patching of message provides us with a magic mock , so when the `isinstance` is called for the first time , since its second parameter is the clazz itself it will return the MagicMock and it runs into exception. If we were checking `isinstance` on another clazz it would have been fine. 

---
In reply to: [246256363](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246256363) [](ancestors = 246256363)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/module_client.py
olivakar: The reason I utilized the send event code is because of the threading code. The only thing that send to output does now is setting the output name. It does nothing else as of now. But true it may do something in the future.
For comment 2 s it is a correct point that the transport should have 1:1 mapping. The trigger called can be same. So if the transport has another end point called send output , we don't need the common function.


Also I am not sure if the AbstractTransport should have that send output as an abstract method send output

---
In reply to: [246257157](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246257157) [](ancestors = 246257157,246254214)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -230,10 +230,12 @@ def _publish_events_in_queue(self, event_data):
...
             logger.info("retrieved event from queue. publishing.")
 
-            if isinstance(message_to_send, Message):
-                encoded_topic = self._encode_properties(message_to_send, self.topic)
-            else:
-                message_to_send = Message(message_to_send)
+            # if isinstance(message_to_send, Message):
+            #     encoded_topic = self._encode_properties(message_to_send, self.topic)
+            # else:
+            #     message_to_send = Message(message_to_send)
```
pierreca: please delete commented code.

azure-iot-hub-devicesdk/samples/send_to_output_module.py
pierreca: technically both exist and would be correct but if we assume that our primary target is edge developers, modules are likely going to be used in the context of edge and therefore with from_environment(). thanks for making that change :)

---
In reply to: [246598483](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246598483) [](ancestors = 246598483,246254580)

azure-iot-hub-devicesdk/samples/send_to_output_module.py
pierreca: that's right. suggestion:
"Inputs/Ouputs are only supported with Azure IoT Edge and the module client"

---
In reply to: [246254837](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246254837) [](ancestors = 246254837)

BertKleewein:
We have iothub modules (which talk directly to iothub) and iotedge modules (which only talk through edgehub).  Output and input messages are a feature that only _really­_ works with edge modules.  (you can do it, but I think it shows up in iothub has telemetry).

The _recommended_ way to authenticate edge modules is from the environment.  It’s _possible_ to authenticate an edge modules with a connection string _if you are able to configure your SSL certificate correctly_.  That last part is tricky and there’s a few different ways to do this and there’s probably articles on the Microsoft web site to describe some of the options and how they might work on a customer site.

The node sample you saw was probably written before connecting through the environment was available and it should probably be changed.  I’m guessing that it also shows how you push the certificate from the app down to the transport.  I’m pretty sure we support pushing the certificate down in the python-preview repo, but I’m not 100% sure.  Eventually we will need it for test and for some devdiv scenarios, but I don’t want to push customers down this route because the only way to make it work well is to really understand how edgehub authentication works at a somewhat deep level.

azure-iot-hub-devicesdk/samples/send_to_output_module.py
olivakar: are you referring to the properties set on the message ? I think in the past I have changed the device client sample too to have some properties.  are you referring to the fact that we should just send only string messages ? 

---
In reply to: [246254489](https://github.com/Azure/azure-iot-sdk-python-preview/pull/35#discussion_r246254489) [](ancestors = 246254489)

PR #39, opened 2019-01-09
Added Async Clients, MQTT Async Adapter, asyncio_compat library

cartertinney:
- Restructured the clients to prioritize Async design patterns while maintaining a Sync client for compatibility scenarios

- Provided an adapter for the MQTT Transport that emulates Async behavior

- Added an asyncio_compat library to azure.iot.common that provides helpers to cover version specific differences in asyncio, as well as async emulation functionality for adapting sync code to behave asynchronously

- Adjusted samples to be based around async patterns in Python 3.5+

NB: Tests are not added yet, and thus it will not pass our gate at the moment


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -1,24 +1,22 @@
...
+import six
 from threading import Event
-from .transport.mqtt.mqtt_transport import MQTTTransport
+from .transport.mqtt import MQTTTransport
 
 logger = logging.getLogger(__name__)
 
 
-class InternalClient(object):
+@six.add_metaclass(abc.ABCMeta)
```
cartertinney: I defined this as an abstract class, which I think might kind of be overkill. My reasoning was, GenericClient shouldn't ever be instantiated, and by defining abstract methods (connect, etc.) we could always guarantee that there was parity between Sync and Async versions of a given method - but only if an abstract method was defined. Not really sure how common it would be for someone to remember to define the abstract method, but then forget to do one of the sync/async ones.

Additionally, the subclasses, GenericSyncClient and GenericAsyncClient are also not supposed to be instantiated, but there's no way to stop them from being, given that there are Device and Module specific methods (i.e. c2d) on the horizon.

I still went ahead and used the abstract base class, but depending on your thoughts, I may remove it - I don't want to add something just for the sake of it, especially when it A) nets us very little, and B) still will require us to fall back on convention. Perhaps convention alone would be sufficient.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -90,68 +157,20 @@ def callback():
...
-        When creating the client, you need to pass in an authorization provider and a transport_name
-
-        The authentication_provider parameter is an object created using the authentication_provider_factory
-        module.  It knows where to connect (a network address), how to authenticate with the service
-        (a set of credentials), and, if necessary, the protocol gateway to use when communicating
-        with the service.
-
-        The transport_name is a string which defines the name of the transport to use when connecting
-        with the service or the protocol gateway.
+class DeviceClientSync(GenericSyncClient):
```
cartertinney: How do you guys feel about this naming?

Having the Sync in different places in the generic, and the user facing version annoys me, but GenericSyncClient sounds better than GenericClientSync (especially given the contrast with GenericAsyncClient), while DeviceClientSync sounds better than DeviceSyncClient (because it has to contrast with DeviceClient)

Open to suggestion here.

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -0,0 +1,111 @@
...
+    return future
+
+
+def emulate_async(fn):
+    """
+    Apply as a decorator to emulate async behavior with a sync function/method
+    via usage of multithreading
+    """
+
+    async def async_fn_wrapper(*args, **kwargs):
```
BertKleewein: >async def async_fn_wrapper(*args, **kwargs): [](start = 4, length = 44)

I think you want to mark this with @wraps(fn)

azure-iot-hub-devicesdk/samples/send_telemetry_device.py
```diff
@@ -3,48 +3,59 @@
...
-from azure.iot.hub.devicesdk.device_client import DeviceClient
-from azure.iot.hub.devicesdk.auth.authentication_provider_factory import from_connection_string
-from azure.iot.hub.devicesdk.message import Message
+import asyncio
+import logging
 import uuid
+from azure.iot.hub.devicesdk import DeviceClient, Message
+from azure.iot.hub.devicesdk.auth.authentication_provider_factory import (
+    from_connection_string,
+)  # this is a overlong import, fix
```
BertKleewein: ># this is a overlong import, fix [](start = 3, length = 32)

I agree 100% as you've seen in my other comments.  We should open a TFS item to make sure we do this.

azure-iot-hub-devicesdk/samples/authentication_providers.py
```diff
@@ -3,35 +3,44 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 import os
-from azure.iot.hub.devicesdk.device_client import DeviceClient, ModuleClient
+import asyncio
+from azure.iot.hub.devicesdk import DeviceClient, ModuleClient
```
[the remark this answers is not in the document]
BertKleewein: I saw your other comment.  Looks like we're both thinking the same thing.

---
In reply to: [246908955](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246908955) [](ancestors = 246908955)

azure-iot-hub-devicesdk/samples/send_telemetry_device.py
```diff
@@ -3,48 +3,59 @@
...
+    def connection_state_callback(status):
+        print("connection status: " + status)
 
-# send only string messages
-for i in range(5, 10):
-    print("sending message #" + str(i))
-    device_client.send_event("test payload message " + str(i))
-    time.sleep(1)
+    # Register the connection state callback with the client...
+    # TODO: open question: connection_state_callback is a sync callback.  Do we want to support async callbacks?
```
BertKleewein: ># TODO: open question: connection_state_callback is a sync callback.  Do we want to support async callbacks? [](start = 4, length = 108)

Lets close on both the TODOs in this file ASAP.  Is there any reason to not do async callbacks and async factory methods?  I think async callbacks are required (otherwise customers need to jump through hoops to call async code) but I'm only 90% convinced about the factory methods.

azure-iot-hub-devicesdk/samples/send_telemetry_device_sync.py
```diff
@@ -0,0 +1,49 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
BertKleewein: This is OK for now, but Pierre floated the idea of having a single uber-sample for sync code.  We should close on this.  If we decide to stick with multiple sync samples, maybe they should go into a different location (like samples/sync or samples_sync)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/async_adapter.py
```diff
@@ -0,0 +1,47 @@
+import logging
+from .mqtt_transport import MQTTTransport
+from azure.iot.common import asyncio_compat
+
+logger = logging.getLogger(__name__)
+
+
+class MQTTAsyncTransport(MQTTTransport):
```
BertKleewein: >MQTTAsyncTransport [](start = 6, length = 18)

For naming, we should be consistent about placement of "Sync" and "Async" modifiers.  It should always be a prefix or it should always be a suffix.  Putting it in the middle might flow off the tongue better, but the inconsistency is frustrating.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
BertKleewein: Not overkill.  I think that forgetting to implement either sync or async functions could be super common.  Any time a developer is required to do the same (or mostly same) thing in more than one place, someone is going to mess it up unless you enforce it.  It's like having a required parameter but not enforcing it.

---
In reply to: [246596935](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246596935) [](ancestors = 246596935)

azure-iot-hub-devicesdk/samples/send_telemetry_device_sync.py
cartertinney: I agree that this sample is kind of unnecessary since we don't want tons of sync samples. However I didn't want to remove anything that already existed, I figured removals could be a separate PR.

---
In reply to: [246912675](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246912675) [](ancestors = 246912675)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/async_adapter.py
cartertinney: I'd advocate for the end then, because putting Async at the beginning throws off alphabetization.

---
In reply to: [246941850](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246941850) [](ancestors = 246941850)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
BertKleewein: I strongly believe in consistency here, and I agree that it sounds better as GenericSyncClient, but I always regret going for "It sounds like better English, so I'm going to break the pattern here".  
If it makes the little high-school English teacher that stands on your shoulder happy, you can think of Sync as the same as Synchronous, and if you call it SyncGenericClient, it really means SynchronousGenericClient and that's two adjectives modifying a noun. (I don't like having mine up there, but she gives better advice than both the devil on my left shoulder and the angel on my right shoulder.)

---
In reply to: [246597321](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246597321) [](ancestors = 246597321)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_clients.py
```diff
@@ -0,0 +1,80 @@
...
+
+        If the connection to the service has not previously been opened by a call to connect, this
+        function will open the connection before sending the event.
+
+        :param message: The actual message to send.
+        """
+        await self._transport.send_event(message)
+
+
+class DeviceClient(GenericAsyncClient):
```
BertKleewein: I have mixed feelings about moving these clients into the same file, mostly based on expected code growth, but I guess I can live with this.

azure-iot-hub-devicesdk/samples/send_telemetry_device.py
cartertinney: This may require some teamwork to address - I largely left this functionality alone in the overhaul, because I wasn't entirely clear on the implications of making changes related to callbacks. I think this will require a separate PR, however I agree that it 100% should be addressed.

---
In reply to: [246911794](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246911794) [](ancestors = 246911794)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
BertKleewein: However, if we opt for async callbacks, then we might not need generic clients because sharing callback code seems to be most of the justification for having these classes.

---
In reply to: [246948932](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246948932) [](ancestors = 246948932,246596935)

BertKleewein:
>> NB: Tests are not added yet, and thus it will not pass our gate at the moment
My personal recommendation would be to get the existing tests passing and then push it and write new tests in a separate commit.   If some code-coverage limit blocks the commit, let's adjust the limit until you can write the tests in a separate commit.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/async_clients.py
cartertinney: We can always move as it grows if need be, but I don't think Module and Device will have enough to justify their own module for quite some time. Python convention is to group related classes in a module, and not one module per class like Java. Especially given that there are now seven (7!) client classes I felt this was an important time to start combining things.

---
In reply to: [246952657](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246952657) [](ancestors = 246952657)

BertKleewein:
By the way, I really like the way you structured this.  It's a big improvement IMHO.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
cartertinney: Yup. I didn't want to touch the callbacks yet since they deal with the transport, and the implications there get a bit more nebulous to me. We would still need the GenericSync / GenericAsync clients however for shared Device/Module functionality

---
In reply to: [246953315](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246953315) [](ancestors = 246953315,246948932,246596935)

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -0,0 +1,111 @@
...
+
+def create_task(coro):
+    """
+    "Best effort" function. If avaialable (Python 3.7+), use asyncio.create_task,
+    which is preferred as it is more specific for the goal of immediately scheduling
+    a task from a coroutine. If not avaialble, use the more general puprose 
+    asyncio.ensure_future
+    """
+    try:
+        task = asyncio.create_task(coro)
```
olivakar: >coro [](start = 35, length = 4)

Should this be given as coro()

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -0,0 +1,111 @@
...
+    except AttributeError:
+        loop = asyncio.get_event_loop()
+    return loop
+
+
+def create_task(coro):
+    """
+    "Best effort" function. If avaialable (Python 3.7+), use asyncio.create_task,
+    which is preferred as it is more specific for the goal of immediately scheduling
+    a task from a coroutine. If not avaialble, use the more general puprose
```
olivakar: >avaialble [](start = 36, length = 9)

spelling i am seeing correctly..

azure-iot-common/azure/iot/common/asyncio_compat.py
cartertinney: Good catch. Since I hadn't written tests yet (and this isn't used anywhere), this slipped by.

---
In reply to: [246957140](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246957140) [](ancestors = 246957140)

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -0,0 +1,111 @@
...
+    Helper function to abstract event loop setup/cleanup. Uses asyncio.run() if it is
+    available. Otherwise, manually gets, runs and closes the event loop
+    """
+    try:
+        asyncio.run(coro, debug=debug)
+    except AttributeError:
+        loop = asyncio.get_event_loop()  # NOT running loop
+        loop.set_debug(debug)
+        loop.run_until_complete(coro)
+        loop.close()
```
olivakar: should we do this in try finally ?

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -0,0 +1,111 @@
...
+def emulate_async(fn):
+    """
+    Apply as a decorator to emulate async behavior with a sync function/method
+    via usage of multithreading
+    """
+
+    async def async_fn_wrapper(*args, **kwargs):
+        loop = get_running_loop()
+
+        # Run fn in default ThreadPoolExecutor (CPU * 5 threads)
```
olivakar: >Run fn in default ThreadPoolExecutor (CPU * 5 threads) [](start = 10, length = 54)

better to mention this somewhere as doc string.

cartertinney: No. Implementation details are not recommended for inclusion in docstrings

---
In reply to: [246960811](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246960811) [](ancestors = 246960811)

olivakar: my mistake...i did not intend to mention the CPU * 5 threads part as doc string...but was referring to the fact that we are using `None` as the executor parameter for `run_in_executor`.  That can be explained as we are using default executor and not a processpool or threadpool..

---
In reply to: [246964302](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246964302) [](ancestors = 246964302,246960811)

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -0,0 +1,111 @@
...
+        await loop.run_in_executor(None, fn, *args, **kwargs)
```
olivakar: >loop.run_in_executor(None, fn, *args, **kwargs) [](start = 14, length = 47)

I see that the python doc pages say that this method is called like `loop.run_in_executor(executor, func, *args)` ....Should it have `kwargs`  then ?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/async_adapter.py
olivakar: >MQTTAsyncTransport [](start = 6, length = 18)

Will this `async_adapter` file have all transports ? or can this be renamed to mqtt_async_adapter ?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/async_adapter.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Asynchronous implementation of MQTTTransport for communication via MQTT protocol.
+    """
+
+    async def connect(self, callback=None):
+        logger.info("async connecting to transport")
+        connect_async = asyncio_compat.emulate_async(super(MQTTAsyncTransport, self).connect)
```
olivakar: >super(MQTTAsyncTransport, self) [](start = 53, length = 31)

This module will only come to play when we are at python 3 right ? May be we can start using `super().methoName(args)` as python 3 supports this as opposed to `super(subClass, instance).method(args)`

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
olivakar: I would prefer having it as `GenericClientSync` and `DeviceClientSync`..  

---
In reply to: [246951779](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246951779) [](ancestors = 246951779,246597321)

azure-iot-common/azure/iot/common/asyncio_compat.py
cartertinney: Potentially. We'll see during testing.

---
In reply to: [246982067](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246982067) [](ancestors = 246982067)

azure-iot-common/azure/iot/common/asyncio_compat.py
cartertinney: Since our exact implementation of emulate_async may change (see: limitations of our current model), I don't consider us tied to using the loop's default executor. 

---
In reply to: [246979670](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246979670) [](ancestors = 246979670,246964302,246960811)

azure-iot-common/azure/iot/common/asyncio_compat.py
olivakar: I am not sure if this understanding is implicit or not. But for me I think it will be really clear if there is a description of the param coro saying that it is a coroutine call or name the param like coro_call or something along those lines.

---
In reply to: [246957640](https://github.com/Azure/azure-iot-sdk-python-preview/pull/39#discussion_r246957640) [](ancestors = 246957640,246957140)

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -0,0 +1,111 @@
+import asyncio
+import concurrent.futures
```
olivakar: >concurrent.futures [](start = 7, length = 18)

unused

PR #43, opened 2019-01-24
Refactored + added tests for Async Adapters

cartertinney:
- Reorganized aysnc helper code. 
- Added tests for async helper code.
- Added tests for the MQTT Transport Adapter
- Added enforced pre-commit to setup rules, adjusted build tasks to account for it

NB: Clients are currently only covered by generic client tests. This should be restructured in a future PR to test individual user level clients, including async ones. That will be a separate PR.


azure-iot-common/tests/conftest.py
```diff
@@ -0,0 +1,8 @@
+import sys
+
+collect_ignore = []
+
+# Ignore Async tests if below Python 3.5
+if sys.version_info < (3, 5):
+    collect_ignore.append("test_async_adapter.py")
```
pierreca: >test_async_adapter.py [](start = 27, length = 21)

nit/for later: can we use a wildcard and a naming pattern for tests?

azure-iot-common/tests/test_asyncio_compat.py
```diff
@@ -0,0 +1,79 @@
...
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="Requires Python 3.7+")
+async def test_get_running_loop(mocker, event_loop):
+    spy_get_running_loop = mocker.spy(asyncio, "get_running_loop")
+    result = asyncio_compat.get_running_loop()
+    assert result == event_loop
+    assert spy_get_running_loop.call_count == 1
+    assert spy_get_running_loop.call_args == mocker.call()
+
+
+@pytest.mark.skipif(sys.version_info >= (3, 7), reason="Requires Python 3.6 or below")
```
pierreca: >sys.version_info >= (3, 7) [](start = 20, length = 26)

did python 3.7 _break_ this? or is it working but we're just not taking this path?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -192,16 +192,6 @@ def callback():
...
 
-    def _emit_connection_status(self):
-        """
-        The connection status is emitted whenever the client on the module gets connected or disconnected.
-        """
-        logger.info("emit_connection_status: {}".format(self.state))
-        if self.on_connection_state:
-            self.on_connection_state(self.state)
-        else:
-            logger.warn("No callback defined for sending state")
```
pierreca: this doesn't seem to be related to just moving tests. + there are still calls to this API from other parts of the file

azure-iot-common/tests/conftest.py
cartertinney: That would be bad here, given that we want to specifically forbid these files, not collect them. Pytest handles collections automatically.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
cartertinney: It's a duplicate. This functionality is provided elsewhere, and got duplicated during a merge

---
In reply to: [250847278](https://github.com/Azure/azure-iot-sdk-python-preview/pull/43#discussion_r250847278) [](ancestors = 250847278)

azure-iot-common/azure/iot/common/asyncio_compat.py
```diff
@@ -13,6 +12,8 @@ def get_running_loop():
         loop = asyncio.get_running_loop()
     except AttributeError:
         loop = asyncio.get_event_loop()
```
olivakar: Should we use `loop = asyncio._get_running_loop()` after todays meeting ?

azure-iot-common/tests/test_asyncio_compat.py
```diff
@@ -0,0 +1,79 @@
...
+async def test_get_running_loop_compat(mocker, event_loop):
```
olivakar: >test_get_running_loop_compat [](start = 10, length = 28)

these 2 tests are testing the compat functionality of get running loop for differing versions. is it also possible to include the py version in the test name?

azure-iot-common/tests/conftest.py
```diff
@@ -0,0 +1,8 @@
+import sys
+
+collect_ignore = []
```
BertKleewein: >collect_ignore [](start = 0, length = 14)

nice!  I didn't know about this feature.  I might have done some things differently in the past had I known this.  Thanks!

BertKleewein:
>     but get_event_loop works well in simple cases (e.g. there is only one loop).

minor: docstring is now out of date -- both the fallback function name and the justification for using the fallback (I assume) is different.

---
Refers to: azure-iot-common/azure/iot/common/asyncio_compat.py:9 in 01d34a1. [](commit_id = 01d34a172120418f86bc74a544452a50bfd879dc, deletion_comment = False)

azure-iot-common/tests/test_asyncio_compat.py
```diff
@@ -0,0 +1,79 @@
...
+async def test_get_running_loop_compat_no_running_loop(mocker):
+    mocker.patch.object(asyncio, "_get_running_loop", return_value=None)
+    with pytest.raises(RuntimeError, message="Expecting Runtime Error"):
+        asyncio_compat.get_running_loop()
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="Requires Python 3.7+")
+async def test_create_task(mocker, dummy_coroutine):
+    spy_create_task = mocker.spy(asyncio, "create_task")
+    coro_obj = dummy_coroutine()
```
BertKleewein: >dummy_coroutine [](start = 15, length = 15)

the word "dummy" will eventually get flagged by one of the code scanners as being potentially offensive.  Use "fake" instead.

azure-iot-common/tests/test_asyncio_compat.py
```diff
@@ -0,0 +1,79 @@
...
+async def test_get_running_loop_compat(mocker, event_loop):
+    spy_get_event_loop = mocker.spy(asyncio, "_get_running_loop")
+    result = asyncio_compat.get_running_loop()
+    assert result == event_loop
+    assert spy_get_event_loop.call_count == 1
+    assert spy_get_event_loop.call_args == mocker.call()
+
+
+@pytest.mark.skipif(sys.version_info >= (3, 7), reason="Requires Python 3.6 or below")
+async def test_get_running_loop_compat_no_running_loop(mocker):
```
BertKleewein: >test_get_running_loop_compat_no_running_loop [](start = 10, length = 44)

Why is there no 3.7 version?  If this module guarantees some behavior for all versions, why are you only testing it for some versions?

cartertinney: There's no case for it in 3.7. In 3.6 we have to raise the error ourselves, thus the test. In 3.7, asyncio throws it for us. 

---
In reply to: [251153162](https://github.com/Azure/azure-iot-sdk-python-preview/pull/43#discussion_r251153162) [](ancestors = 251153162)

azure-iot-common/tests/test_async_adapter.py
```diff
@@ -0,0 +1,56 @@
...
+    mock_fn = mocker.MagicMock(return_value=dummy_value)
+    mock_fn.__doc__ = "docstring"
+    return mock_fn
+
+
+async def test_emulate_async(mocker, mock_function, dummy_value):
+    async_fn = async_adapter.emulate_async(mock_function)
+    result = await async_fn(dummy_value)
+    assert inspect.iscoroutinefunction(async_fn)
+    assert async_fn.__doc__ == mock_function.__doc__  # verify docstring carries over
```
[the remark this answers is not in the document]
cartertinney: Too specific an implementation detail IMO. There's other ways to make it happen other than functools.wrap, and I'd rather just test that it has the effect we want, rather than the method by which we achieve it.

---
In reply to: [251156400](https://github.com/Azure/azure-iot-sdk-python-preview/pull/43#discussion_r251156400) [](ancestors = 251156400)

azure-iot-common/tests/test_asyncio_compat.py
```diff
@@ -0,0 +1,79 @@
...
+@pytest.fixture
+def dummy_coroutine():
+    async def coro():
+        return
+
+    return coro
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="Requires Python 3.7+")
+async def test_get_running_loop(mocker, event_loop):
```
BertKleewein: >test_get_running_loop [](start = 10, length = 21)

suggestion: Tests are easier to understand if the expected behavior as part of the function name.  Otherwise, I can't see what the test is doing without reading the code and thinking through the logic.  If I'm looking at a list of tests, or if I see that this test fails, I have to read through and think through a bunch of code before I can even guess what's going wrong.  I can guess that it tests the get_running_loop function, but so do a bunch of other tests in this file.  What makes this test important?
e.g: 
test_py37_get_running_loop_calls_asyncio_get_running_loop
test_py36_and_older_get_running_loop_calls_asyncio_underscore_get_running_loop
test_py36_and_older_get_running_loop_throws_if_asyncio_underscore_get_running_loop_returns_none
test_py37_create_task_calls_asyncio_create_task
test_py36_and_older_create_task_calls_asyncio_ensure_future
test_py352_and_newer_create_future_calls_event_loop_create_future
test_py351_and_older_create_future_creates_asyncio_future_object

azure-iot-common/tests/test_asyncio_compat.py
```diff
@@ -0,0 +1,79 @@
+import pytest
```
BertKleewein: >pytest [](start = 7, length = 6)

what is the "event_loop" parameter in your calls.  It looks like a fixture, but I don't see if defined anywhere in here and I don't see how it could be imported.  What crazy magic is this?

cartertinney: pytest-asyncio

---
In reply to: [251159459](https://github.com/Azure/azure-iot-sdk-python-preview/pull/43#discussion_r251159459) [](ancestors = 251159459)

vsts/build.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
   - script: 'python -m pip install --upgrade pip'
     displayName: 'Update pip'
 
-  - script: 'python dev_setup.py'
+  - script: 'python env_setup.py --no_dev'
```
BertKleewein: I'm curious why you needed to add this.  Was pre-commit causing problems?

cartertinney: Yup. Not entirely sure why, but also didn't have much problem coding around it - VSTS has no reason to install pre-commit anyway. So now there's a "--no_dev" mode that's just for running and testing.

---
In reply to: [251161237](https://github.com/Azure/azure-iot-sdk-python-preview/pull/43#discussion_r251161237) [](ancestors = 251161237)

azure-iot-common/tests/test_asyncio_compat.py
cartertinney: I edited that comment a few times and somehow some of it got lost (and different parts of different edits got combined)

I've since added the test. My rationale had been that since I'd be forcing the error with my mock there was no actual test - I was testing test code, not actual implementation code. However, I realized that just because the underlying function call throws an exception doesn't mean ours does - proving that a raised exception bubbles up has some value.

tl;dr I added the test

---
In reply to: [251155632](https://github.com/Azure/azure-iot-sdk-python-preview/pull/43#discussion_r251155632) [](ancestors = 251155632,251153867,251153162)

azure-iot-common/tests/test_asyncio_compat.py
cartertinney: I've made some adjustments that are not quiet the same as these, but should help

---
In reply to: [251158481](https://github.com/Azure/azure-iot-sdk-python-preview/pull/43#discussion_r251158481) [](ancestors = 251158481)

PR #45, opened 2019-01-31
Async Naming/Namespacing Adjustments + Tests

cartertinney:
*Adjusted client naming/namespacing as per Python guidelines
*Added Async Client tests
*Refactored existing client tests
*Added bugfixes


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/authentication_provider.py
```diff
@@ -22,6 +22,7 @@ def __init__(self, hostname, device_id, module_id=None):
     def get_current_sas_token(self):
         pass
 
+    # @abc.abstractstaticmethod
```
cartertinney: Ideally it would be @abc.abstractstaticmethod, since parse is not a regular method, but a static method (does not have self, or cls as a parameter). That is to say, as is, with the @abc.abstractmethod, this is incorrect. Unfortunately, @abc.abstractstaticmethod does not exist in 2.7. At some we'll have to fix this, but it seemed low priority, so I left it alone for now.

azure-iot-hub-devicesdk/samples/receive_c2d_message.py
```diff
@@ -41,7 +41,7 @@ def c2d_message_handler(c2d_message):
 
 while True:
     selection = input("Press Q: Quit for exiting\n")
-    if selection is "Q" or selection is "q":
+    if selection == "Q" or selection == "q":
```
cartertinney: VSTS Flake8 plugin seems to have updated as of Jan 30th such that this change had to be made in order to pass static code analysis.

azure-iot-hub-devicesdk/tests/test_sync_clients.py
```diff
@@ -0,0 +1,130 @@
...
+    xfail_notimplemented = pytest.mark.xfail(raises=NotImplementedError, reason="Unimplemented")
+
+    client_classes = [DeviceClient, ModuleClient]
+
+    @pytest.fixture(params=client_classes)
+    def client(self, request, transport):
+        client_class = request.param
+        return client_class(transport)
+
+    @pytest.mark.parametrize(
```
olivakar: >@pytest.mark.parametrize( [](start = 4, length = 25)

The verbose output looks good and we should change the build step for pytest to have verbose in my opinion, otherwise its pretty difficult to understand if all cases were covered or not.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/auth/authentication_provider.py
olivakar: i tried this. It did not work  for all versions of python. I think I tried another way also. Using 2 decorators `@staticmethod`
 and `@abstractmethod` but that too did not work. I think both of these work in python 3.2  up and above.

---
In reply to: [252901525](https://github.com/Azure/azure-iot-sdk-python-preview/pull/45#discussion_r252901525) [](ancestors = 252901525)

PR #49, opened 2019-02-15
C2D/Input Message (Sync Client)

cartertinney:
This PR contains the implementation and tests for C2D and Input Message functionality for the Sync Client ONLY.

Async Client support will come in a later PR.


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,142 @@
...
+        super(MessageQueue, self).__init__()
+        self._enabled = False
+        self._on_enable = on_enable
+        self._on_disable = on_disable
+
+    def enable(self):
+        self._enabled = True
+        self._on_enable()
+
+    def disable(self):
```
cartertinney: Enable and disable have to be able to be done on a MessageQueue because it's possible to disable receiving messages for specific inputs on the module client. You can't just enable/disable the whole feature

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @enabled.setter
+    def enabled(self, v):
+        if v:
+            self.enable()
+        else:
+            self.disable()
+
+
+class MessageQueueManager(object):
```
cartertinney: Perhaps MessagingManager might be a better name for what this does (although it doesn't currently, and probably shouldn't, handle sending them).

Think of this as a modular, self-contained piece of the client itself. Thus why it gets the transport (which it needs in order to intelligently enable/disable the c2d/input messages)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def _input_message_queue_on_disable(self):
+        # Disable input messages as a feature if all input queues are disabled
+        if all(not queue.enabled for queue in self.input_message_queues.values()):
+            self._disable_feature("input")
+
+    def _c2d_message_queue_on_disable(self):
+        self._disable_feature("c2d")
+
+    def _enable_feature(self, feature_name):
```
cartertinney: I'm considering moving these back to the client, since they'll be different depending on sync/async client, and it would be ideal to have this manager be sync/async agnostic

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,142 @@
...
+        if v:
+            self.enable()
+        else:
+            self.disable()
+
+
+class MessageQueueManager(object):
+    def __init__(self, transport):
+        self.c2d_message_queue = None
+        self.input_message_queues = {}
```
cartertinney: No use of weak references (for now). We still could as a helpful feature to ensure we don't eat up a ton of memory due to user error, but I figured we can add that later since it's not necessary for the core function of making this work properly. Make it work first, then make it right and all that.

azure-iot-hub-devicesdk/samples/receive_c2d_message.py
```diff
@@ -18,26 +19,26 @@
...
-# Register the connection state callback with the client...
-device_client.on_connection_state = connection_state_callback
 
-# ... and connect the client.
-device_client.connect()
-
-# enable the device to receive c2d messages
-device_client.enable_feature("c2d", c2d_message_handler)
+# Run a listener thread in the background
+listen_thread = threading.Thread(target=c2d_listener, args=(c2d_message_queue,), daemon=True)
```
BertKleewein: >daemon=True [](start = 81, length = 11)

cool.  I didn't know about the daemon parameter.  This is new for 3.3.  For 2.7 compat, it looks like there's a setDaemon function

BertKleewein:
> # Licensed under the MIT License. See License.txt in the project root for license information.

Instead of renaming, you could follow [https://stackoverflow.com/questions/42511879/pytest-skip-entire-module-file-python-2-and-3](https://stackoverflow.com/questions/42511879/pytest-skip-entire-module-file-python-2-and-3):

(my favorite answer if it works)
if pytest.__version__ < "3.0.0":
  pytest.skip()
else:
  pytestmark = pytest.mark.skip 


---
Refers to: azure-iot-hub-devicesdk/tests/skip_test_generic_client_sync_LEGACY.py:3 in 68febba. [](commit_id = 68febba27025a99f1215f73b8d51627d9d56288f, deletion_comment = False)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
BertKleewein: Agree that this object shouldn't have anything to do with sending messages.  

---
In reply to: [257085321](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r257085321) [](ancestors = 257085321)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,142 @@
...
+        enable_complete = Event()
+
+        def callback():
+            enable_complete.set()
+
+        if feature_name == "input":
+            self._transport.enable_input_messages(callback)
+        elif feature_name == "c2d":
+            self._transport.enable_c2d_messages(callback)
+        else:
```
BertKleewein: I think the transport will eventually have a similar pattern - self._transport.enable_feature(feature_name -- this might affect your decision to move this back into the client or not.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,142 @@
...
+            if queue.enabled:
+                queue.put(message_received)
+                logger.info("Input message delivered to {} queue".format(input_name))
+            else:
+                logger.warn("Input message queue for {} is disabled".format(input_name))
+
+    def receive_c2d_message(self, message_received):
+        # No check on if the queue exists - it should.
+        # If for some reason it doesn't, exception will be raised
```
BertKleewein: >exception will be raised [](start = 41, length = 24)

This shouldn't throw.  This is very possible -- for MQTT at least, subscription state survives a disconnect/reconnect.  So, if you are subscribed, then you reboot your device and reconnect, you're still subscribed even though the caller might never have called any enable functions.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,142 @@
...
+            )
+            self.c2d_message_queue = queue
+        else:
+            queue = self.c2d_message_queue
+
+        if queue.enabled is not True:
+            queue.enable()
+        return queue
+
+    def receive_input_message(self, input_name, message_received):
```
BertKleewein: >receive_input_message [](start = 8, length = 21)

don't worry too much about this, but the pattern for this function (and get_input_message_queue) will be repeated at least twice more (for twin updates and for method calls).  Both of those are like input messages in that we'll have multiple queues for a single feature.

azure-iot-hub-devicesdk/samples/receive_c2d_message.py
cartertinney: Oh, good catch, I didn't know it was a different API in 2.7

---
In reply to: [257345281](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r257345281) [](ancestors = 257345281)

cartertinney:
> # Licensed under the MIT License. See License.txt in the project root for license information.

Actually, this file is supposed to be deleted entirely. It's still kicking around due to a merge I think.

---
In reply to: [464154816](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#issuecomment-464154816) [](ancestors = 464154816)

---
Refers to: azure-iot-hub-devicesdk/tests/test_generic_client_sync_LEGACY.py:3 in 68febba. [](commit_id = 68febba27025a99f1215f73b8d51627d9d56288f, deletion_comment = False)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,138 @@
...
+    def _c2d_message_queue_on_enable(self):
+        self._enable_feature("c2d")
+
+    def _input_message_queue_on_disable(self):
+        # Disable input messages as a feature if all input queues are disabled
+        if all(not queue.enabled for queue in self.input_message_queues.values()):
+            self._disable_feature("input")
+
+    def _c2d_message_queue_on_disable(self):
+        self._disable_feature("c2d")
```
pierreca: >c2d [](start = 31, length = 3)

nit: for strings, think about practicing the "rule of 3": if you re-use a value more than 3 times in the code, it belongs in a constant.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,138 @@
...
+        def callback():
+            enable_complete.set()
+
+        if feature_name == "input":
+            self._transport.enable_input_messages(callback)
+        elif feature_name == "c2d":
+            self._transport.enable_c2d_messages(callback)
+        else:
+            logger.error("Feature name has not been defined.Feature names can be 'input' or 'c2d'")
+            raise ValueError("Feature names can be only among 'input' or 'c2d'")
```
pierreca: >ValueError("Feature names can be only among 'input' or 'c2d'") [](start = 18, length = 62)

(opinion) I'm not sure  value (but a lot of potential bugs) in using strings as a way to discriminate which function to call on the transport... I would much rather call something like MessagingManager.c2d.disable() for example. c2d could be an "interface" (or an ABC) that has enable/disable methods

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,138 @@
...
+            self._transport.disable_input_messages(callback)
+        elif feature_name == "c2d":
+            self._transport.disable_c2d_messages(callback)
+        else:
+            logger.error("Feature name has not been defined.Feature names can be 'input' or 'c2d'")
+            raise ValueError("Feature names can be only among 'input' or 'c2d'")
+
+        disable_complete.wait()
+
+    def get_input_message_queue(self, input_name):
```
pierreca: >get_input_message_queue [](start = 8, length = 23)

if you wanted to commit fully to the use of strings/enums, shouldn't this be get_feature()?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -24,51 +29,17 @@ def __init__(self, transport):
...
         """
         self._transport = transport
-        self._transport.on_transport_connected = self._handle_transport_connected_state
-        self._transport.on_transport_disconnected = self._handle_transport_connected_state
+        self._transport.on_transport_connected = self._state_change
+        self._transport.on_transport_disconnected = self._state_change
 
         self.state = "initial"
+        self._message_queues = MessageQueueManager(
+            transport
```
pierreca: Have you considered using a simple dictionary (key/value pairs) with the queues? it seems like we're jumping through a lot of hoops to just enable/disable features when we could just be calling this on the queue objects directly.

azure-iot-hub-devicesdk/samples/receive_c2d_message.py
```diff
@@ -18,26 +19,27 @@
...
-    print("the data in the message received was ")
-    print(c2d_message.data)
-    print("custom properties are")
-    print(c2d_message.custom_properties)
 
+# define behavior for receiving a C2D message
+def c2d_listener(message_queue):
+    while True:
+        c2d_message = message_queue.get()  # blocking call
+        print("the data in the message received was ")
```
pierreca: I guess we could also use get_nowait() and be non-blocking right?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,108 @@
...
+            self.input_message_queues[input_name] = queue
+
+        if queue.enabled is not True:
+            queue.enable()
+        return queue
+
+    def get_c2d_message_queue(self):
+        queue = self.c2d_message_queue
+
+        if queue.enabled is not True:
```
olivakar: >if queue.enabled is no [](start = 8, length = 22)

will this not work ?
`if not queue.enabled:`

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,108 @@
...
+            queue.enable()
+        return queue
+
+    def receive_input_message(self, input_name, message_received):
+        delivered = False
```
olivakar: >delivered = False [](start = 8, length = 17)

Is it a convention to return Boolean variables ? can we not return true from the if and otherwise by default return false ?

cartertinney: I'd be okay with that, but generally the IoT team has preferred one exit methods, i.e. only one return statement, so I did it this way.

---
In reply to: [259134718](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259134718) [](ancestors = 259134718)

azure-iot-hub-devicesdk/samples/receive_c2d_message.py
cartertinney: That is correct

---
In reply to: [258282020](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r258282020) [](ancestors = 258282020)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
cartertinney: It was originally just a dictionary, yeah, but I expanded it to its own class since there was a lot of associated code cluttering up the client and I wanted to keep things clean and separated.

---
In reply to: [258281695](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r258281695) [](ancestors = 258281695)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
cartertinney: This method is private by convention, and is only to be used by us internally, so there's no concern about typos or user error. However the interface has further been adjusted so that this happens at transport level now, getting rid of this logic from client level

---
In reply to: [258258132](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r258258132) [](ancestors = 258258132)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
cartertinney: I'm not sure what you mean by this. This method returns a specified input message queue, not a feature.

---
In reply to: [258258677](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r258258677) [](ancestors = 258258677)

olivakar:
>         :param auth_provider: The authentication provider

unused

---
Refers to: azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py:28 in 3210b3f. [](commit_id = 3210b3fa72ba6e5978976b97e010a0f403f4cfa6, deletion_comment = False)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -10,6 +10,7 @@
 from .mqtt_provider import MQTTProvider
 from transitions.extensions import LockedMachine as Machine
 from azure.iot.hub.devicesdk.transport.abstract_transport import AbstractTransport
+from azure.iot.hub.devicesdk.transport import constant
```
olivakar: >constant [](start = 46, length = 8)

cant we declare constants here ? 
or else it is better tot take all constants from here to the constant module.

cartertinney: These constants apply to the entire transport package, so I figured they ought to be defined there, as they are used elsewhere as well.

When used elsewhere you import them the same way - from azure.iot.hub.devicesdk.transport import constant, then you refer to the constants as, say constant.C2D (see: tests, message_queue, etc.)

I find this syntax preferable to exposing them through the init by doing "from azure.iot.hub.devicesdk.transport import C2D" as the constant module adds clarity.

I'm open to alternatives if anyone has strong feelings about this though.

---
In reply to: [259174246](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259174246) [](ancestors = 259174246)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class MessageQueueManager(object):
+    def __init__(self, enable_feature_fn, disable_feature_fn):
+        self._enable_feature_fn = enable_feature_fn
+        self._disable_feature_fn = disable_feature_fn
+        self.c2d_message_queue = MessageQueue(
+            self._c2d_message_queue_on_enable, self._c2d_message_queue_on_disable
+        )  # not yet enabled
+        self.input_message_queues = {}
```
olivakar: >{} [](start = 36, length = 2)

if it is a dictionary may be use `dict()`

cartertinney: Python convention is to use literals + there is reduced performance from using constructors.

---
In reply to: [259175712](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259175712) [](ancestors = 259175712)

azure-iot-hub-devicesdk/tests/test_message_queue.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @pytest.fixture
+    def on_disable_stub(self, mocker):
+        return mocker.stub(name="on_disable_stub")
+
+    @pytest.fixture
+    def message_queue(self, on_enable_stub, on_disable_stub):
+        return MessageQueue(on_enable_stub, on_disable_stub)
+
+    def test_instantiates_unenabled(self, message_queue):
```
olivakar: >test_instantiates_unenabled [](start = 8, length = 27)

test instantiates not enabled

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,129 @@
...
+    @property
+    def enabled(self):
+        return self._enabled
+
+    @enabled.setter
+    def enabled(self, v):
+        if v:
+            self.enable()
+        else:
+            self.disable()
```
pierreca: I'm seeing 2 ways of doing the same thing (call enable/disable methods or change the property). I would rather only have one way. if we can't make this read-only, then I would make enable/disable private (underscore methods)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
pierreca: You have get_input_message_queue and get_c2d_message_queue (ie one function per feature) however you have "enable_feature" and "disable_feature" where I'm guessing "feature_name" could be either input or c2d. that's where the API is asymetrical.

---
In reply to: [259172314](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259172314) [](ancestors = 259172314,258258677)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,129 @@
...
+        delivered = False
+        try:
+            queue = self.input_message_queues[input_name]
+        except KeyError:
+            logger.warning("No input message queue for {} - dropping message".format(input_name))
+        else:
+            if queue.enabled:
+                queue.put(message_received)
+                delivered = True
+                logger.info("Input message delivered to {} queue".format(input_name))
```
pierreca: >delivered [](start = 43, length = 9)

nit: "delivered" is misleading. it sounds like the message has been sent on the network and the delivery ack has been received - when really it was just enqueued (the latter would be my suggestion)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,129 @@
...
+            else:
+                logger.warning(
+                    "Input message queue for {} is disabled - dropping message".format(input_name)
+                )
```
pierreca: note for later: i understand why we're doing this but i'm mildly uncomfortable with it. This is one of the hypothesis we should validate during the user study.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
```diff
@@ -0,0 +1,129 @@
...
+        return delivered
+
+    def receive_c2d_message(self, message_received):
+        delivered = False
+        if self.c2d_message_queue.enabled:
+            self.c2d_message_queue.put(message_received)
+            delivered = True
+            logger.info("C2D message delivered to queue")
```
pierreca: >delivered [](start = 37, length = 9)

same here

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/message_queue.py
cartertinney: enable() and disable() are the preferred ways to do this, so should not be private. This setter is more of an insurance policy - the "enabled" property is not able to be private, as it's important to check the enabled status of a MessageQueue. However, because it's public, it could potentially be overwritten, which would result in unexpected and hard to trace behaviour when the enabled property is set without triggering the on_enable/on_disable handlers.

We could just throw an exception on an attempt to set here instead, but I felt that this was a smoother solution

---
In reply to: [259534214](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259534214) [](ancestors = 259534214)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -218,13 +159,12 @@ class ModuleClient(GenericClientSync):
 
     def __init__(self, transport):
         super(ModuleClient, self).__init__(transport)
-        self._transport.on_transport_input_message_received = self._handle_input_message_received
+        self._transport.on_transport_input_message_received = (
+            self._message_queues.receive_input_message
```
pierreca: >receive_input_message [](start = 33, length = 21)

nit: naming: that sounds more like a "handle_input_message" or "enqueue_input_message"

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/constant.py
```diff
@@ -0,0 +1,2 @@
+C2D = "c2d"
+INPUT = "input"
```
pierreca: is this really better than an enum? long term?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
pierreca: I like having all constants in the same place. easier to debug/find them later. whether these should be in an enum or "floating" constants I'm not sure. pick the most pythonic way?

---
In reply to: [259175173](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259175173) [](ancestors = 259175173,259174246)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -384,21 +385,39 @@ def send_output_event(self, message, callback=None):
     def _on_shared_access_string_updated(self):
         self._trig_on_shared_access_string_updated()
 
-    def enable_input_messages(self, callback=None, qos=1):
+    def enable_feature(self, feature_name, callback=None, qos=1):
```
pierreca: >enable_feature [](start = 8, length = 14)

again I understand why we do this but i would like to validate this with our user study, because we're the only SDK doing this and it feels like additional hoops and if we have a way to auto-enable we should.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -384,21 +385,39 @@ def send_output_event(self, message, callback=None):
...
+            logger.error("Feature name {} is unknown".format(feature_name))
+            raise ValueError("Invalid feature name")
+
+    def disable_feature(self, feature_name, callback=None):
+        if feature_name == constant.INPUT:
+            self._disable_input_messages(callback)
+        elif feature_name == constant.C2D:
+            self._disable_c2d_messages(callback)
+        else:
+            logger.error("FEature name {} is unknown".format(feature_name))
```
pierreca: >FEature [](start = 26, length = 7)

nit: typo

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -384,21 +385,39 @@ def send_output_event(self, message, callback=None):
...
+            self._enable_input_messages(callback, qos)
+        elif feature_name == constant.C2D:
+            self._enable_c2d_messages(callback, qos)
+        else:
+            logger.error("Feature name {} is unknown".format(feature_name))
+            raise ValueError("Invalid feature name")
+
+    def disable_feature(self, feature_name, callback=None):
+        if feature_name == constant.INPUT:
+            self._disable_input_messages(callback)
```
pierreca: >_disable_input_messages [](start = 17, length = 23)

this big if/else statement may not grow well. fine for now, but if we get past 3 or 4 features that have enable/disable functions, we may want to establish a pattern here (dictionary or otherwise)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
cartertinney: Enum isn't really a thing in standard Python prior to 3.4. Python originally just used string literals for this purpose. For compat reasons I think that's the route we should continue, rather than depending on an external enum library.

---
In reply to: [259537743](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259537743) [](ancestors = 259537743,259175173,259174246)

cartertinney:
>         :param auth_provider: The authentication provider

Resolving in another PR

---
In reply to: [466225313](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#issuecomment-466225313) [](ancestors = 466225313)

---
Refers to: azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py:28 in 3210b3f. [](commit_id = 3210b3fa72ba6e5978976b97e010a0f403f4cfa6, deletion_comment = False)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/constant.py
cartertinney: Yeah, enums are relatively new to Python. This is standard practice to use string literals. The practice of a separate constants.py is less common, but makes sense given that it's specific to the package rather than a module.

---
In reply to: [259537543](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259537543) [](ancestors = 259537543)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
cartertinney: We are autoenabling as far as the client is concerned. Not sure what this comment means. Will discuss further at a later time.

---
In reply to: [259538058](https://github.com/Azure/azure-iot-sdk-python-preview/pull/49#discussion_r259538058) [](ancestors = 259538058)

PR #50, opened 2019-02-16
disable/unsubscribe with tests

olivakar:
Even though we disable the feature, and unsubscribe from mqtt , this does not stop the hub from sending messages. 
Still need to figure out how to do that.
Should we drop messages silently ?
Should we detach handler ?
Should we disconnect after disabling (not  a good option if the device still  wants telemetry)

We need to discuss regarding this options. So this is a basic disabling.
The sample to test does not exist in this branch. The sample and unused methods of client is in the branch `ok-disable-DONOT-DELETE`.

### DIFF
--- a/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
+++ b/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
@@ -34,6 +34,7 @@ class MQTTProvider(object):
         self.on_mqtt_disconnected = None
         self.on_mqtt_published = None
         self.on_mqtt_subscribed = None
+        self.on_mqtt_unsubscribed = None
 
         self._create_mqtt_client()
 
@@ -88,11 +89,21 @@ class MQTTProvider(object):
                 logger.error("Unexpected error calling on_mqtt_message_received")
                 logger.error(traceback.format_exc())
 
+        def on_unsubscribe_callback(client, userdata, mid):
+            logger.info("UNSUBACK received for %s", str(mid))
+            # TODO: how to do failure?
+            try:
+                self.on_mqtt_unsubscribed(mid)
+            except:  # noqa: E722 do not use bare 'except'
+                logger.error("Unexpected error calling on_mqtt_unsubscribed")
+                logger.error(traceback.format_exc())
+
         self._mqtt_client.on_connect = on_connect_callback
         self._mqtt_client.on_disconnect = on_disconnect_callback
         self._mqtt_client.on_publish = on_publish_callback
         self._mqtt_client.on_subscribe = on_subscribe_callback
         self._mqtt_client.on_message = on_message_callback
+        self._mqtt_client.on_unsubscribe = on_unsubscribe_callback
 
         logger.info("Created MQTT provider, assigned callbacks")
 
@@ -157,4 +168,15 @@ class MQTTProvider(object):
         """
         logger.info("subscribing")
         (result, mid) = self._mqtt_client.subscribe(topic, qos)
+        return mid
+
+    def unsubscribe(self, topic):
+        """
+        Unsubscribe the client from one topic.
+        :param topic: a single string which is the subscription topic to unsubscribe from.
+        :return: mid the message ID for the unsubscribe request.
+        Raises a ValueError if topic is None or has zero string length, or is not a string.
+        """
+        logger.info("unsubscribing")
+        (result, mid) = self._mqtt_client.unsubscribe(topic)
         return mid
--- a/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
+++ b/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
@@ -43,6 +43,10 @@ class MQTTTransport(AbstractTransport):
         self._connect_callback = None
         self._disconnect_callback = None
         self._subscribe_callback = None
+        self._unsubscribe_callback = None
+
+        self._c2d_topic = None
+        self._input_topic = None
 
         states = ["disconnected", "connecting", "connected", "disconnecting"]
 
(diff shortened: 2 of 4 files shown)


BertKleewein:
>> Even though we disable the feature, and unsubscribe from mqtt , this does not stop the hub from sending messages.
>> Still need to figure out how to do that.

Have you seen examples of this?  The whole point of unsubscribing is that it should stop the hub from sending.  If you have seen it, is it just an example of messages that are in flight that still come through or can you repro this with new messages?  Is this from iothub or from edgehub?

>> Should we drop messages silently ?

Yes.  There are a few different reasons to drop messages silently.
#1: our sessions don't set the clean flag to true, so if we reboot the device and reconnect, iothub will not reset previous subscriptions.  At this point, the client app will be out of sync with iothub because iothub will think the app is subscribed even though the app might not be subscribed.
#2: for input messages and methods, we subscribe for a topic that will send us all inputmessages and all methods even though the app might not be subscribed for all of them.  

>> Should we detach handler ?

No opinion.  Not sure if the implications here.  

>> Should we disconnect after disabling (not a good option if the device still wants telemetry)

Absolutely not.

cartertinney:
Agree with Bert on all this.

---
In reply to: [464854788](https://github.com/Azure/azure-iot-sdk-python-preview/pull/50#issuecomment-464854788) [](ancestors = 464854788)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
```diff
@@ -88,11 +89,21 @@ def on_message_callback(client, userdata, mqtt_message):
                 logger.error("Unexpected error calling on_mqtt_message_received")
                 logger.error(traceback.format_exc())
 
+        def on_unsubscribe_callback(client, userdata, mid):
+            logger.info("UNSUBACK received for %s", str(mid))
+            # TODO: how to do failure?
```
pierreca: ># TODO: how to do failure? [](start = 12, length = 26)

I don't think this can "fail". either you get an UNSUBACK and it worked, or you don't, right?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, client_id, hostname, username, ca_cert=None):
         self.on_mqtt_disconnected = None
         self.on_mqtt_published = None
         self.on_mqtt_subscribed = None
+        self.on_mqtt_unsubscribed = None
```
pierreca: maybe not for now but these are "mini-states" so I wonder if we need to track this at this level or in the transport state machine and if these can get out of sync.
Also, there will be multiple topics to publish and subscribe to, so while for now it's OK, this is going to have to change and we should start thinking about it now.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -120,6 +124,19 @@ def __init__(self, auth_provider):
...
+                "dest": None,
+                "after": "_call_unsubscribe",
+            },
+            {"trigger": "_trig_disable_receive", "source": "connecting", "dest": None},
+            {
+                "trigger": "_trig_disable_receive",
+                "source": "disconnected",
+                "dest": "connecting",
+                "after": "_call_provider_connect",
+            },
```
pierreca: I don't know about that. triggering a connection just so that we can unsubscribe seems a little painful. I guess we could discuss the nuance of what the user is really trying to achieve (stop receiving messages vs actually cleaning up a subscription to a topic).

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
olivakar: created on board

---
In reply to: [258251940](https://github.com/Azure/azure-iot-sdk-python-preview/pull/50#discussion_r258251940) [](ancestors = 258251940)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
olivakar: I think it is referring to except ? all these callbacks probably mean how to handle the exception block...I did not want to break the pattern hence kept the same TODO

---
In reply to: [258251395](https://github.com/Azure/azure-iot-sdk-python-preview/pull/50#discussion_r258251395) [](ancestors = 258251395)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
olivakar: I also felt this when I was trying to define the states this time. May be we should start thinking about the state machine in a different way. Bert was thinking through something. May be all subscriptions can happen during connection nd vice versa. As of now I will let this be.

---
In reply to: [258252555](https://github.com/Azure/azure-iot-sdk-python-preview/pull/50#discussion_r258252555) [](ancestors = 258252555)

olivakar:
>> Even though we disable the feature, and unsubscribe from mqtt , this does not stop the hub from sending messages.
>> Still need to figure out how to do that.

Have you seen examples of this?  The whole point of unsubscribing is that it should stop the hub from sending.  If you have seen it, is it just an example of messages that are in flight that still come through or can you repro this with new messages?  Is this from iothub or from edgehub?

----- kept on happening from Iothub for new messages

>> Should we drop messages silently ?

Yes.  There are a few different reasons to drop messages silently.
#1: our sessions don't set the clean flag to true, so if we reboot the device and reconnect, iothub will not reset previous subscriptions.  At this point, the client app will be out of sync with iothub because iothub will think the app is subscribed even though the app might not be subscribed.
#2: for input messages and methods, we subscribe for a topic that will send us all inputmessages and all methods even though the app might not be subscribed for all of them.  

>> Should we detach handler ?

No opinion.  Not sure if the implications here.  

---- There is an exception raised for a code path. We can handle that but I did not dive too much into this , so there could be other scenarios i did not think of.

>> Should we disconnect after disabling (not a good option if the device still wants telemetry)

Absolutely not. 

----- I also think so. I think I was trying to ask also if we should unsubscribe when we disconnect rather than exposing disable as a separate behavior

---
In reply to: [465262290](https://github.com/Azure/azure-iot-sdk-python-preview/pull/50#issuecomment-465262290) [](ancestors = 465262290,464854788)

PR #51, opened 2019-02-25
Added missing docs + fixed some existing docs

cartertinney:
Added missing docs at client layer. Fixed some existing docs as well. Added copyright header to all files.

- Transport still is missing some documentation.
- Not all documentation is consistently styled. (Will wait until we're integrated with the doc gen tool)


azure-iot-common/azure/iot/common/async_adapter.py
```diff
@@ -19,13 +28,11 @@ def emulate_async(fn):
 
 
 class AwaitableCallback(object):
-    """
-    A callback whose completion can be waited upon.
+    """A sync callback whose completion can be waited upon.
```
pierreca: >sync [](start = 9, length = 4)

nit (super nitpicky for later): prefer full words over abbreviated forms.

PR #52, opened 2019-02-26
Build Process Changes

cartertinney:
Will now build all packages as part of the VSTS process, not just devicesdk

- Added build_packages.py
- Modified setup.py files to have latest dependencies
- Improved README files


azure-iot-hub-devicesdk/README.md
```diff
@@ -1,111 +1,45 @@
...
+### Module Client
+* Supports Azure IoT Edge Hub and Azure IoT Hub
+* Send telemetry messages to a Hub or to another Module
+* Receive Input messages from a Hub or other Modules
 
-## Quick start
+These clients are synchronous and communicate using the MQTT protocol
 
+## Installation
+We currently do not provide a binary distribution of our package, which means you'll have to clone the [repository](https://github.com/Azure/azure-iot-sdk-python-preview) and install with:
```
pierreca: I assume this must be run from the package directory itself? can we call out where this must be run from for each SDK?

azure-iot-hub-devicesdk/samples/README.md
```diff
@@ -0,0 +1,12 @@
+# Samples for the Azure IoT Hub Device SDK
+
+This directory contains simple samples showing how to use the various features of the Microsoft Azure IoT Hub service from a device running Python
+
+## List of Samples
+
+* Send Telemetry from a Device
+* Receive Cloud-to-Device (C2D) messages
+* Send to Output from Module
```
pierreca: Receive on Input for modules

azure-iot-hub-devicesdk/README.md
cartertinney: It's meant to be in the same directory as this README since the setup.py is right there. But I can add clarification.

azure-iot-hub-devicesdk/samples/README.md
cartertinney: That would be inaccurate, because that sample does not actually exist yet.

PR #56, opened 2019-03-04
receive input messages and auth providers sample

olivakar:


azure-iot-hub-devicesdk/samples/authentication_providers.py
```diff
@@ -14,35 +14,52 @@
...
+    env_auth_provider = from_environment()
+    device_client_env = await DeviceClient.from_authentication_provider(env_auth_provider, "mqtt")
+    await device_client_env.connect()
 
 
 if __name__ == "__main__":
     loop = asyncio.get_event_loop()
-    loop.run_until_complete(main())
+    # Here we use option 1, but option 2 and 3 can be used interchangeably.
+    loop.run_until_complete(create_shared_access_sig_auth_provider())
```
cartertinney: Can we use asyncio.run() here instead? Anything that can abstract the event loop is better

pierreca:
@olivakar @cartertinney is there a reason for this to not be merged yet? should we update the branch and merge or are we waiting on something else?

PR #59, opened 2019-03-13
initial classes for security and client

olivakar:
I have given this PR only for initial review , so that I don't keep on continuing in the wrong direction.

1.  Please do not review the `mqtt_transport`. It does not have any state machine as of now.
2. The `query registration status` part is not implemented yet.
2. I tried having a sample only to check that it is working.

I will continue working on the state machine (transport) , query and if these initial files look okay then the tests.


azure-iot-provisioning-devicessdk/setup.py
```diff
@@ -0,0 +1,38 @@
...
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.4",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+    ],
+    install_requires=["six>=1.12.0,<2.0.0", "azure-iot-common"],
+    packages=find_packages(exclude=["tests"]),
```
cartertinney: exclude samples as well

cartertinney:
Why is this devicessdk instead of devicesdk?

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client.py
```diff
@@ -0,0 +1,43 @@
...
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class RegistrationClient(object):
+    """
+    Super class for any client that can be used to register devices to Device Provisioning Service.
+    :ivar str provisioning_host: Host running the Device Provisioning Service.
+    :ivar str security_client: Instance of Security client
+    :ivar str module_id: Instance of transport.
```
cartertinney: None of these exist in the abstract class.

And the ivar that DOES exist (transport) isn't listed.

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/sk_registration_client.py
```diff
@@ -0,0 +1,35 @@
...
+class SymmetricKeyRegistrationClient(RegistrationClient):
+    """
+    Client which can be used to run the registration of a device using Symmetric Key authentication.
+    """
+
+    def __init__(self, transport):
+        """
+        Initializer for the Symmetric Key Registration Client
+        """
+        RegistrationClient.__init__(self, transport)
```
cartertinney: Use super

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/security/sk_security_provider.py
```diff
@@ -0,0 +1,64 @@
...
+
+class SymmetricKeySecurityProvider(object):
+    """
+    A provider that is responsible for providing shared access tokens that will eventually establish
+    the authenticity of devices to Device Provisioning Service.
+    """
+
+    def __init__(self, registration_id, symmetric_key, id_scope):
+        """
+        Initialize the symmetric key security client.
```
cartertinney: >client [](start = 46, length = 6)

provider?

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/security/sk_security_provider.py
```diff
@@ -0,0 +1,64 @@
...
+        The registration ID is alphanumeric, lowercase string and may contain hyphens.
+        """
+        return self._registration_id
+
+    @property
+    def id_scope(self):
+        """
+        :return: Host running the Device Provisioning Service.
+        """
+        return self._id_scope
```
cartertinney: Why not just let these be public attributes? I don't see a reason they couldn't be dynamically modified.

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    def __init__(self, transport):
+        """
+        Initializes the registration client.
+        :param transport: Instance of the Transport object.
+        """
+        self.transport = transport
```
pierreca: >transport [](start = 13, length = 9)

should this be "private" (underscore?)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
```diff
@@ -0,0 +1,24 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class SymmetricKeyTransport(object):
```
cartertinney: Unclear as to the point of this abstract class.

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client_factory.py
```diff
@@ -0,0 +1,37 @@
...
+from .sk_registration_client import SymmetricKeyRegistrationClient
+
+
+def create_from_security_provider(provisioning_host, security_provider, transport_choice):
+    """
+    Creates different types of registration clients which can enable devices to communicate with Device Provisioning
+    Service based on parameters passed.
+    :param provisioning_host: Host running the Device Provisioning Service. Can be found in the Azure portal in the
+    Overview tab as the string Global device endpoint
+    :param security_provider: Instance of Security client object which can be either of Symmetric Key, TPM or X.509
```
pierreca: >security_provider [](start = 11, length = 17)

the param is called security provider but the comment talks about security client. other sdks talk about security client.. i would suggest to go with this until we decide another name is better?

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/sk_registration_client.py
```diff
@@ -0,0 +1,35 @@
...
+        self.transport.on_registration_complete = self.on_device_registration_complete
+
+    def register(self):
```
pierreca: >register(self) [](start = 8, length = 14)

nit: missing docstrings

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client.py
```diff
@@ -0,0 +1,43 @@
...
+This module provides an abstract interface representing clients which can communicate with the
+Device Provisioning Service.
+"""
+
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class RegistrationClient(object):
```
cartertinney: >RegistrationClient [](start = 6, length = 18)

You call this a Provisioning Client in the sample. Why is it called Registration Client here?

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/security/sk_security_provider.py
pierreca: let's make it client everywhere as other SDKs use that - then we can discuss what the better name is.

---
In reply to: [265317526](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265317526) [](ancestors = 265317526)

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,54 @@
...
+id_scope = os.getenv("PROVISIONING_IDSCOPE")
+registration_id = os.getenv("PROVISIONING_REGISTRATION_ID")
+symmetric_key = os.getenv("PROVISIONING_SYMMETRIC_KEY")
+
+
+symmetric_key_security_provider = SymmetricKeySecurityProvider(
+    registration_id, symmetric_key, id_scope
+)
+provisioning_client = create_from_security_provider(
+    provisioning_host, symmetric_key_security_provider, "mqtt"
```
cartertinney: Is there a semantic reason why the provisioning_host information can't be bundled with the security provider? I'm not saying that I necessarily think that's the right way to do it, but since they both get passed down to the transport, it makes me wonder why provisioning_host isn't part of the security_provider.

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/security/sk_security_provider.py
```diff
@@ -0,0 +1,64 @@
...
+        The registration ID is alphanumeric, lowercase string and may contain hyphens.
+        :param symmetric_key: The key which will be used to create the shared access signature token to authenticate
+        the device with the Device Provisioning Service. By default, the Device Provisioning Service creates
+        new symmetric keys with a default length of 32 bytes when new enrollments are saved with the Auto-generate keys
+        option enabled. Users can provide their own symmetric keys for enrollments by disabling this option within
+        16 bytes and 64 bytes and in valid Base64 format.
+        :param id_scope: The ID scope is used to uniquely identify the specific provisioning service the device will
+        register through. The ID scope is assigned to a Device Provisioning Service when it is created by the user and
+        is generated by the service and is immutable, guaranteeing uniqueness.
+        """
```
pierreca: this is VERY good. thanks so much for going through this effort. I would love to see all docs be as precise as this. good work!

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,54 @@
...
+def registration_status_callback(topic, payload):
+    result = str(payload)
+    print(result)
+    if "operationId" in result:
+        print("Device is registering")
+
+
+provisioning_client.on_registration_complete = registration_status_callback
+
+provisioning_client.register()
```
cartertinney: I'd like to see an async sample too. Not that it'd be much different. But still.

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def registration_status_callback(topic, payload):
+    result = str(payload)
+    print(result)
+    if "operationId" in result:
+        print("Device is registering")
+
+
+provisioning_client.on_registration_complete = registration_status_callback
```
cartertinney: >registration_status_callback [](start = 47, length = 28)

Given what this does, it's kind of difficult to figure out another way to do this, but I really, really, really hate callbacks, and we have to at some point figure out a way to replace this pattern.

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,54 @@
...
+from azure.iot.hub.devicesdk.auth.authentication_provider_factory import from_connection_string
+
+import time
+
+logging.basicConfig(level=logging.INFO)
+
+provisioning_host = os.getenv("PROVISIONING_HOST")
+id_scope = os.getenv("PROVISIONING_IDSCOPE")
+registration_id = os.getenv("PROVISIONING_REGISTRATION_ID")
+symmetric_key = os.getenv("PROVISIONING_SYMMETRIC_KEY")
```
cartertinney: Sample README should define what these are.

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/security/sk_security_provider.py
pierreca: There's no reason for them to change during execution. it's associated with a specific device and DPS instance. not sure how you want this to exist in Python, but we don't want the customers to think they can reuse the same client to register multiple devices. I don't think that's a scenario. 

---
In reply to: [265318495](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265318495) [](ancestors = 265318495)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/device_provisioning_constant.py
```diff
@@ -0,0 +1,23 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""
+This module contains constants related to the transport package.
+"""
+
+USER_AGENT = "azure-iot-provisioning-device" + "/" + "1.5.3"
```
pierreca: >azure-iot-provisioning-device" + "/" + "1.5.3" [](start = 14, length = 46)

probably want this to identify that it's the python preview SDK and reflect a proper version (right now this will look like node). I would suggest "<pip_package_name>/<pip_package_version>"

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
pierreca: IIRC not all protocols support all types of authentication. MQTT does not support TPM I think for example. it'd be nice to have a way to express this in the class definition itself, i'm guessing that's what Oliva is going for.

---
In reply to: [265319178](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265319178) [](ancestors = 265319178)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    A super class of transports that use shared access signature token to connect.
+    """
+
+    def __init__(self, symmetric_key_security_provider):
+        self.security_provider = symmetric_key_security_provider
```
pierreca: >security_provider [](start = 13, length = 17)

same here - should this be underscored/private?

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/security/sk_security_provider.py
cartertinney: I mean, these attributes only do something when "create_shared_access_signature" is called. That itself just draws on whatever the current value is - there's no inherent need for the values to stay the same, as they aren't being changed internally, or carrying any kind of intermediate value. No other internal or intermediate values are derived from or depend on them either.

If you change the registration id and and id scope, the object isn't invalidated. "create_shared_access_signature" still works as if the object was instantiated with different values. Nothing breaks. Thus, I don't see the need to go private by convention.

The only time the values should be private by convention is when changing them would put the application in an unexpected state IMO. Otherwise, the Python principle of developers being "responsible adults" should apply. Python is not a defensive language. Thus, unless there's a good reason related to the integrity of the application, I don't believe that attributes should be private by convention.

---
In reply to: [265323518](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265323518) [](ancestors = 265323518,265318495)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
cartertinney: Does this then mean that we end up with multiple transports per protocol, inheriting from a different authentication base?

i.e. AMQP that inherits from SymmetricKeyTransport, and another AMQP that inherits from a different type of authentication?

---
In reply to: [265325807](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265325807) [](ancestors = 265325807,265319178)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
cartertinney: As per my comment on the sk_security_provider, I don't really see a reason this has to be private by convention.

---
In reply to: [265325942](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265325942) [](ancestors = 265325942)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
pierreca: ideally in OO programming you'd use interface and implement multiple interfaces. not sure how you'd translate this in Python. if a class can inherit from only one ABC, then maybe the ABC is unnecessary and we just rely on duck-typing?

---
In reply to: [265328420](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265328420) [](ancestors = 265328420,265325807,265319178)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
pierreca: because public means it becomes a public API that we have to document for the users and that we have to support. trust me on this, you want everything private by default and open things up only if you really need to. all these objects are going to be public, you don't want customer using these public fields to mess with things.

---
In reply to: [265328937](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265328937) [](ancestors = 265328937,265325942)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
cartertinney: We could do some kind of multiple inheritance perhaps... but that can get a bit dicey in Python, or so I've heard (in practice I've never had an issue).

My recommendation is we don't tie ourselves into a particular hierarchy of abstraction before it's necessary.

---
In reply to: [265330148](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265330148) [](ancestors = 265330148,265328420,265325807,265319178)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
cartertinney: Why would there be a public API for transport, which isn't part of the exposed API to begin with?

---
In reply to: [265330521](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265330521) [](ancestors = 265330521,265328937,265325942)

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
pierreca: that's a valid question. i think originally the idea was that we would separate what constitutes "identity" (symmetric key, registration id) from what is specific to the DPS instance itself (Id scope and global device endpoint).
In the node SDK the idScope is passed to the registration client, not the security client. 

@Oliva please check with other SDKs what the design of the parameters for each object constructor is and let's see if there's a common rule everybody abides by. if there is, let's follow it. if there is not, let's figure out what we think is best as a team.

---
In reply to: [265321560](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265321560) [](ancestors = 265321560)

pierreca:
@olivakar great start!! :) this PR could get really big really quick. I would suggest to "stub" the transport, try to finalize the object model and initial set of tests, and check this in, then move on with adding transport features themselves. 
Great work, keep it up!

cartertinney:
+1

---
In reply to: [472608799](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#issuecomment-472608799) [](ancestors = 472608799)

azure-iot-provisioning-devicessdk/tests/test_dummy.py
```diff
@@ -0,0 +1,2 @@
+def test_dummy():
```
pierreca: >dummy [](start = 9, length = 5)

nit: dummy will fail the policheck (authorized vocabulary) test. suggest using stub, mock, fake, test... some other word.

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
olivakar: this mqtt from provisioning should be capable of handling symmetric auth and x509 auth. Some other transports also would have sas token functionality , hence I tried to go for multiple inheritance as many transports would take on this functionality. As far as i know it is possible in python.

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client.py
olivakar: changed in sample

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client.py
olivakar: changed to private

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client.py
olivakar: changed to private

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/registration_client_factory.py
olivakar: changed to security client

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/sk_registration_client.py
olivakar: I thought the 2 super formats for python 2 and python 3 are not same and this is the only one that works for both. I can find out more when unit tests are added.

Python 2 format for super is
`super().__init__(arg1, arg2)`

Python 3 super is
`super(ChildClass, self).__init__(arg1, arg2)`

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/sk_registration_client.py
olivakar: done

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/device_provisioning_constant.py
olivakar: I have changed to the format. however our package name does not contain the word "preview" ,

---
In reply to: [265324593](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265324593) [](ancestors = 265324593)

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
olivakar: I had already tried both ways(bundling the host inside security provider and keeping it separate) before giving this PR out as the same thing had confused me. The only reason I finally went with accepting 2 parameters (the host and security provider) separately is because  :-

1. the provisioning host is only needed in MQTT for the hostname part. There is no other need for this.
2. The security client on its own does not need this piece of information. It only needs id scope, registration id and the key.

If still we feel that the security client can accept this host and only have a getter defined out of it then I can make the change.

---
In reply to: [265334055](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265334055) [](ancestors = 265334055,265321560)

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
olivakar: wont fix right now.

---
In reply to: [265322440](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265322440) [](ancestors = 265322440)

azure-iot-provisioning-devicessdk/samples/register_symmetric_key.py
olivakar: i think we need async clients for it

---
In reply to: [265321914](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265321914) [](ancestors = 265321914)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
olivakar: Changed to private as of now as I don't quite remember why I did public.

---
In reply to: [265331363](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r265331363) [](ancestors = 265331363,265330521,265328937,265325942)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/device/sk_registration_client.py
cartertinney: That's backwards, it's Python 2 that is super(ChildClass, self).__init__(arg1, arg2). And that version is also compatible in 3. 

---
In reply to: [267876918](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r267876918) [](ancestors = 267876918)

azure-iot-provisioning-devicessdk/azure/iot/provisioning/devicessdk/transport/symmetric_key_transport.py
cartertinney: As per our discussion I'd like to see this class removed until the next PR.

---
In reply to: [267658451](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r267658451) [](ancestors = 267658451)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/device/sk_registration_client.py
```diff
@@ -0,0 +1,38 @@
...
+        RegistrationClient.__init__(self, transport)
+        self._transport.on_registration_complete = self.on_device_registration_complete
+
+    def register(self):
+        """
+        Register the device with the provisioning service.
+        """
+        self._transport.send_registration_request()
+
+    def cancel(self):
```
cartertinney: docstring

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/device/sk_registration_client.py
```diff
@@ -0,0 +1,38 @@
...
+        pass
+
+    def on_device_registration_complete(self, topic, payload):
```
cartertinney: this should be a convention private method, right?, i.e. with an underscore

azure-iot-provisioning-devicesdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,52 @@
...
+
+logging.basicConfig(level=logging.INFO)
+
+provisioning_host = os.getenv("PROVISIONING_HOST")
+id_scope = os.getenv("PROVISIONING_IDSCOPE")
+registration_id = os.getenv("PROVISIONING_REGISTRATION_ID")
+symmetric_key = os.getenv("PROVISIONING_SYMMETRIC_KEY")
+
+
+symmetric_key_security_provider = SymmetricKeySecurityClient(
```
cartertinney: mismatch in naming. Client or provider?

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/device/registration_client_factory.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+This module provides factory methods for creating clients which would
+communicate with Device Provisioning Service.
+"""
+from ..security.sk_security_client import SymmetricKeySecurityClient
+from ..transport.mqtt_transport import MQTTTransport
+from .sk_registration_client import SymmetricKeyRegistrationClient
+
+
+def create_from_security_provider(provisioning_host, security_client, transport_choice):
```
cartertinney: >security_provider [](start = 16, length = 17)

Mismatch here again. Security Provider, or Security Client? Param says client, function says provider.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/symmetric_key_transport.py
```diff
@@ -0,0 +1,24 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class SymmetricKeyTransport(object):
```
cartertinney: As per our discussion I'd like to see this class removed for the time being, and a solution for handling different methods of security client authentication added only when necessary, and the problem has been fully scoped. (the next PR with X509)

cartertinney:
Unclear why the package has three subpackages, but nothing in the package itself. I'd move everything from azure.iot.provisioning.devicesdk.device into azure.iot.provisioning.devicesdk. This will also eliminate relative imports that use ..

azure-iot-provisioning-devicesdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,52 @@
+import os
+import logging
+from azure.iot.provisioning.devicessdk.security.sk_security_client import SymmetricKeySecurityClient
+from azure.iot.provisioning.devicessdk.device.registration_client_factory import (
+    create_from_security_provider,
+)
+from azure.iot.hub.devicesdk import DeviceClient
+from azure.iot.hub.devicesdk.auth.authentication_provider_factory import from_connection_string
```
cartertinney: why are these imports here?

azure-iot-provisioning-devicesdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,52 @@
+import os
+import logging
+from azure.iot.provisioning.devicessdk.security.sk_security_client import SymmetricKeySecurityClient
+from azure.iot.provisioning.devicessdk.device.registration_client_factory import (
+    create_from_security_provider,
+)
```
cartertinney: importing a single function is generally a bad look. Ideally a module is imported, and you use module.function_name(). See how the latest Hub Device SDK samples use auth for an idea of what I'm talking about. But even making that change here won't be enough, because the module "registration_client_factory" is too unwieldly.

I have some ideas for adjustments to API shape, but don't worry about it now - once the featureset is more fleshed out, we can revisit the shape of the imports, and the exact instantiation process.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/device/registration_client_factory.py
```diff
@@ -0,0 +1,39 @@
...
+    Overview tab as the string Global device endpoint
+    :param security_client: Instance of Security client object which can be either of Symmetric Key, TPM or X.509
+    :param transport_choice: A string representing the transport the user wants
+    :return: A specific registration client based on parameters and validations.
+    """
+    if not provisioning_host:
+        raise ValueError("Provisioning Host must be provided.")
+    transport_choice = transport_choice.lower()
+    if transport_choice == "mqtt":
+        if isinstance(security_client, SymmetricKeySecurityClient):
```
cartertinney: >isinstance(security_client, SymmetricKeySecurityClient): [](start = 11, length = 56)

Given that Python is a duck typed language, any code that checks `isinstance()` is a red flag.

However, sometimes it's necessary, and I don't really see a better way to handle this.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/device_provisioning_constant.py
```diff
@@ -0,0 +1,23 @@
+# -------------------------------------------------------------------------
```
cartertinney: I'd rename this file to just be `constant.py`.

Then, constants can be imported and used like:

```python
from transport import constant

timeout = constant.DEFAULT_TIMEOUT
```

The key is that when defining a module that will actually be accessed and used inline as a module, the name should be short.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/device_provisioning_constant.py
```diff
@@ -0,0 +1,23 @@
...
+# --------------------------------------------------------------------------
+"""
+This module contains constants related to the transport package.
+"""
+
+USER_AGENT = "azure-iot-provisioning-devicesdk" + "/" + "0.0.1"
+"""
+Default interval for polling, to use in case service doesn't provide it to us.
+"""
+POLLING_INTERVAL = 2000
```
cartertinney: should this be `DEFAULT_POLLING_INTERVAL` since it's a default value, and for default timeout you've called it `DEFAULT_TIMEOUT`?

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/device_provisioning_constant.py
```diff
@@ -0,0 +1,23 @@
...
+
+"""
+api version to use while communicating with service.
+"""
+API_VERSION = "2018-11-01"
+"""
+Timeout to use when communicating with the service
```
cartertinney: specify default in docstring (nit)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/mqtt_transport.py
```diff
@@ -0,0 +1,82 @@
...
+import six.moves.urllib as urllib
+from . import device_provisioning_constant
+from azure.iot.hub.devicesdk.transport.mqtt.mqtt_provider import MQTTProvider
+from .symmetric_key_transport import SymmetricKeyTransport
+
+logger = logging.getLogger(__name__)
+
+
+class MQTTTransport(SymmetricKeyTransport):
+    def __init__(self, provisioning_host, security_provider):
```
cartertinney: > __init__( [](start = 7, length = 10)

set self.on_registration_complete to None in the init. While it won't change function, it's a good idea to have a list of all handlers defined in one place for clarity, instead of dynamically adding the attribute from the Registration Client.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/mqtt_transport.py
```diff
@@ -0,0 +1,82 @@
...
+        SymmetricKeyTransport.__init__(self, security_provider)
+        self._provisioning_host = provisioning_host
+        self._create_mqtt_provider()
+
+    def send_registration_request(self):
```
cartertinney: > send_registration_request [](start = 7, length = 26)

missing docs on all methods in this file

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/mqtt_transport.py
```diff
@@ -0,0 +1,82 @@
...
+    def _on_provider_disconnect_complete(self):
+        logger.info("_on_provider_disconnect_complete")
+
+    def _on_provider_publish_complete(self, mid):
+        logger.info("_on_provider_publish_complete")
+
+    def _on_provider_subscribe_complete(self, mid):
+        logger.info("_on_provider_subscribe_complete")
+
+    def on_provider_message_received_callback(self, topic, payload):
```
cartertinney: >def on_provider_message_received_callback [](start = 4, length = 41)

Why is this specific callback/handler the only one without a convention private with underscore?

Also why is it the only one with callback in the title?

(It's better without callback tbh)

azure-iot-provisioning-devicesdk/samples/register_symmetric_key.py
```diff
@@ -0,0 +1,52 @@
+import os
+import logging
+from azure.iot.provisioning.devicessdk.security.sk_security_client import SymmetricKeySecurityClient
```
cartertinney: >devicessdk [](start = 28, length = 10)

this namespace has not been updated

cartertinney: >security.sk_security_client [](start = 39, length = 27)

User should not be having to import from specific submodules and subpackages. Expose this all through the devicesdk `__init__.py`

azure-iot-provisioning-devicesdk/tests/device/test_registration_client_factory.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_raises_when_client_created_from_no_host(security_client):
+    with pytest.raises(ValueError, match="Provisioning Host must be provided."):
+        create_from_security_provider("", security_client, "mqtt")
+
+
+def test_raises_when_client_created_from_security_provider_with_not_symmetric_security():
+    with pytest.raises(
+        ValueError,
+        match="A symmetric key security provider must be provided for instantiating MQTT",
```
cartertinney: >"A symmetric key security provider must be provided for instantiating MQTT" [](start = 14, length = 75)

Open question as to if we want to be this specific about catching our exceptions in testing. I so far haven't been matching messages, but I'm open to it. I think it would require a much better formalization of the error story to be effective.

azure-iot-provisioning-devicesdk/tests/device/test_registration_client_factory.py
```diff
@@ -0,0 +1,63 @@
...
+    return "fakehost.com"
+
+
+@pytest.fixture
+def security_client():
+    return SymmetricKeySecurityClient("fake_registration_id", "fake_symmetric_key", "fake_id_scope")
+
+
+@pytest.mark.parametrize(
+    "protocol,expected_transport",
```
cartertinney: I'd also parametrize different types of security clients. Right now there's no test for matching protocols with security clients.

cartertinney:
Missing tests for:

device/sk_registration_client.py
transport/mqtt_transport.py

azure-iot-provisioning-devicesdk/tests/security/test_sk_security_client.py
```diff
@@ -0,0 +1,35 @@
...
+key_name = "registration"
+fake_provisioning_host = "beauxbatons.academy-net"
+fake_registration_id = "MyPensieve"
+module_id = "Divination"
+fake_id_scope = "Enchanted0000Ceiling7898"
+signature = "IsolemnlySwearThatIamuUptoNogood"
+expiry = "1539043658"
+
+
+def test_getters_after_create_security_client():
```
cartertinney: Test that the properties are not settable as well

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/mqtt_transport.py
olivakar: Also why is it the only one with callback in the title?
 -- got copied from the other transport. it is better without callback

---
In reply to: [270205914](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r270205914) [](ancestors = 270205914)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/symmetric_key_transport.py
olivakar: i am not using this anywhere. I have kept this file around 

---
In reply to: [269801835](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r269801835) [](ancestors = 269801835)

azure-iot-provisioning-devicesdk/tests/security/test_sk_security_client.py
olivakar: are u referring to renaming ?

---
In reply to: [270213951](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r270213951) [](ancestors = 270213951)

azure-iot-provisioning-devicesdk/samples/register_symmetric_key.py
olivakar: what kind of namespace updates are you referring ? are you referring to the ones that happened as part of discussion ?

---
In reply to: [270209007](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r270209007) [](ancestors = 270209007)

azure-iot-provisioning-devicesdk/tests/device/test_registration_client_factory.py
olivakar: right now since I have not introduced the other security clients I did not include that. as far as the transports are concerned we always know there are 3 types of transport.

---
In reply to: [270211088](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r270211088) [](ancestors = 270211088)

azure-iot-provisioning-devicesdk/setup.py
```diff
@@ -0,0 +1,46 @@
...
+
+with open("README.md", "r") as fh:
+    _long_description = fh.read()
+
+setup(
+    name="azure-iot-provisioning-devicesdk",
+    version="0.0.1",
+    description="Microsoft Azure IoT SDK Provisioning",
+    license="MIT License",
+    url="https://github.com/Azure/azure-iot-sdk-python",
```
BertKleewein: >url="https://github.com/Azure/azure-iot-sdk-python", [](start = 4, length = 52)

nit: wrong URL

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/registration_client_factory.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+This module provides factory methods for creating clients which would
+communicate with Device Provisioning Service.
+"""
+from .security.sk_security_client import SymmetricKeySecurityClient
+from .transport.mqtt_transport import MQTTTransport
+from .sk_registration_client import SymmetricKeyRegistrationClient
+
+
+def create_from_security_client(provisioning_host, security_client, transport_choice):
```
BertKleewein: >provisioning_host [](start = 32, length = 17)

this will also need to accept an id_scope (unless that concept is gone or the user passes it in some other way that I haven't come across yet).

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/sk_registration_client.py
```diff
@@ -0,0 +1,88 @@
...
+            logger.info("Successfully subscribed to Hub")
+
+        def callback_request():
+            send_request_complete.set()
+            logger.info("Successfully sent request to Hub")
+
+        self._transport.send_registration_request(
+            callback_subscribe=callback_subscribe, callback_request=callback_request
+        )
+        subscribe_complete.wait() and send_request_complete.wait()
```
BertKleewein: >subscribe_complete.wait() and send_request_complete.wait() [](start = 8, length = 58)

this flow is really a state machine thing.  Any ordering of operations like this belong there.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/sk_registration_client.py
```diff
@@ -0,0 +1,88 @@
...
+        def callback_disconnect():
+            disconnect_complete.set()
+            logger.info("Successfully disconnected from IoT Hub")
+
+        self._transport.disconnect(
+            callback_disconnect=callback_disconnect, callback_unsubscribe=callback_unsubscribe
+        )
+
+        # unsubscribe_complete.wait()
+        disconnect_complete.wait()
```
BertKleewein: Same comment about this belonging in the state machine.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/security/sk_security_client.py
```diff
@@ -0,0 +1,64 @@
...
+from azure.iot.common.sastoken import SasToken
+
+
+class SymmetricKeySecurityClient(object):
+    """
+    A client that is responsible for providing shared access tokens that will eventually establish
+    the authenticity of devices to Device Provisioning Service.
+    """
+
+    def __init__(self, registration_id, symmetric_key, id_scope):
```
BertKleewein: >id_scope [](start = 55, length = 8)

Ah, here's the id_scope.  I'm not exactly sure where the user should pass this in, but I think it belongs with the provisioning host.  It comes from exactly the same portal page as the provisioning host IIRC.

olivakar: so the `id_scope` is also needed by this class to produce the URI for `SasToken`. Since the host was only needed for the registration client I took the host parameter out and clubbed these 3 together..I understand what you mean by that the host and id_scope are available on portal page and the other 2 are available in enrollment details. I am open to suggestions if you feel we need to separate these out.

---
In reply to: [271431656](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r271431656) [](ancestors = 271431656)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/sk_registration_client.py
olivakar: will change state machines in another PR.

---
In reply to: [271430810](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r271430810) [](ancestors = 271430810)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/sk_registration_client.py
olivakar: same comment..will take care in another PR.

---
In reply to: [271430573](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r271430573) [](ancestors = 271430573)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/registration_client_factory.py
olivakar: id scope is inside the security client

---
In reply to: [271430125](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r271430125) [](ancestors = 271430125)

azure-iot-provisioning-devicesdk/tests/security/test_sk_security_client.py
cartertinney: No I'm referring to you need to add another test to test that the properties are not settable.

---
In reply to: [271106251](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r271106251) [](ancestors = 271106251,270213951)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/mqtt_transport.py
cartertinney: I realized I totally botched this comment. on_registration_complete is part of the SymmetricKeyRegistrationClient, and it gets set in the sample. 

What I meant to put here was to set on_registration_complete to None in the SymmetricKeyRegistrationClient

---
In reply to: [270205336](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r270205336) [](ancestors = 270205336)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/transport/mqtt_transport.py
cartertinney: oh you copied it from devicesdk, which has flawed docs. I see. Not your problem. 

---
In reply to: [270205541](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r270205541) [](ancestors = 270205541)

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/sk_registration_client.py
```diff
@@ -0,0 +1,88 @@
...
+        # TODO : Chnages to state machine inside mqtt
+        subscribe_complete.wait() and send_request_complete.wait()
+
+    def cancel(self):
+        """
+        Cancels the current registration process.
+        """
+        pass
+
+    def disconnect(self):
```
BertKleewein: >disconnect [](start = 8, length = 10)

What did we decide about including disconnect?  Does it make sense to have the disconnect be part of the registration flow?  in other words, does "register" do connect->register->disconnect like node does?

BertKleewein:
great start!  I'm signing off on the API shape and the rough flow, but I almost completely ignored the transport as we discussed.

azure-iot-provisioning-devicesdk/azure/iot/provisioning/devicesdk/sk_registration_client.py
olivakar: Disconnect does not make sense to be a part of register. But as of now the sample is working with disconnect, so I will leave it for now. Since we are revisiting the state machine we can change disconnect to be part of one of the polling or request response and delete it from client. 

---
In reply to: [272690807](https://github.com/Azure/azure-iot-sdk-python-preview/pull/59#discussion_r272690807) [](ancestors = 272690807)

PR #61, opened 2019-03-14
Device and Client Methods

cartertinney:
* Client implementation
* Samples (sync and async)
* Test stubs (Client)


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
```diff
@@ -0,0 +1,39 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module contains a class representing method calls that are received and responded to.
+"""
```
cartertinney: I'll probably combine this into a singular module with Message, called data_structures.py or something in the final version. Essentially, a module for user-facing objects.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
```diff
@@ -0,0 +1,39 @@
...
+        :param payload: The payload being sent with the call.
+        :param int response_timeout: The time in seconds before a response to the call is invalid.
+        """
+        self._name = name
+        self._payload = payload
+        self._response_timeout = response_timeout
+
+    @property
+    def name(self):
+        return self._name
```
BertKleewein: What is the value in making these properties instead of just attributes?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -151,6 +154,52 @@ def callback():
...
+            method_inbox = self._inbox_manager.get_named_method_call_inbox(method_name)
+        else:
+            method_inbox = self._inbox_manager.get_generic_method_call_inbox()
+
+        logger.info("Waiting for method call...")
+        method_call = method_inbox.get(block=block, timeout=timeout)
+        logger.info("Received method call")
+        return method_call
+
+    def send_method_response(self, method, result, status):
```
BertKleewein: I'm curious why this is on the client instead of putting it on the MethodCall object.  We've debated this a few times, and I always thought the MethodCall was a better place, but I don't know if those debates were ever finalized.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
cartertinney: The method call acts as a token to respond to the request with. Thus, it's semantically supposed to be a read-only object.

In truth the only field that's necessary, I believe, in order to respond is request ID (which I, of course, forgot to include on this structure >_>), but the idea is that this object works as a token, and the token should not be changed.

I do believe in trying to avoid enforcing read-only attributes whenever possible in Python, however, I think we can be a bit more cautious with user facing objects, especially ones that have semantic meaning like this one.

---
In reply to: [266089272](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266089272) [](ancestors = 266089272)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
cartertinney: They haven't really been explicitly finalized. This is as good a place as any to continue the discussion.

Some reasoning I chose to do it this way (all of which, by pure coincidence, begin with the letter 'S'):

- Simplicity of implementation. Passing callbacks through to MethodCall objects, as with the early version of Message objects (before the ack feature was scrapped), adds a lot of code, specifically code that involves passing around method pointers through multiple structures (Transport -> Client -> Inbox Manager -> MethodCall) in order to get transport functionality onto an external, user facing object like MethodCall. It's very ugly, and not particularly simple to follow along with.
- Symmetry. Receiving methods on the client, but sending the response from the MethodCall is kind of weird. 
- Semantics. Responding to a method call, is a client operation. It is the client that responds to the method, not the method itself. So, I think it's correct for the response to be done via the client. It makes clear that it is an I/O operation of the client, and that I/O is occurring.
- Separation of concern. This object represents the data of a method call. Combining that with actual call functionality creates an object with multiple purposes, which, I'm told, is not an ideal design. Objects should have one purpose.

---
In reply to: [266090705](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266090705) [](ancestors = 266090705)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/inbox_manager.py
```diff
@@ -78,3 +108,20 @@ def route_c2d_message(self, incoming_message):
         self.c2d_message_inbox._put(incoming_message)
         logger.info("C2D message sent to inbox")
         return True
```
pierreca: >return True [](start = 8, length = 11)

I feel funny about seeing return True in a lot of places and return False in only one for some reason.
I wonder if we should not return anything, and in the place where we return false, just rethrow the exception.

Feels kinda funny to generate internal exceptions only to end up using return codes.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
pierreca: I don't mind having smaller modules. one potentially desirable consequence would be the ability to cleanly isolate features into modules. Maybe for modularity purposes, maybe if there's such a thing as tree shaking in python? if this is all internal and has no bearing on the public interface though, we can always refactor later. your call.

---
In reply to: [265780208](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r265780208) [](ancestors = 265780208)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -88,6 +88,9 @@ def __init__(self, transport):
         """
         super(GenericClientSync, self).__init__(transport)
         self._inbox_manager = InboxManager(inbox_type=SyncClientInbox)
+        self._transport.on_transport_method_call_message_received = (
+            self._inbox_manager.route_method_call
+        )
```
pierreca: not familiar enough with python to understand why these parenthesis are necessary... educate me?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/inbox_manager.py
```diff
@@ -48,6 +54,30 @@ def get_c2d_message_inbox(self):
...
         return self.c2d_message_inbox
 
+    def get_generic_method_call_inbox(self):
+        """Retrieve the Inbox for generic method calls.
+
+        :returns: An Inbox for generic method calls.
+        """
+        return self.generic_method_call_inbox
+
+    def get_named_method_call_inbox(self, method_name):
```
pierreca: >method_name [](start = 42, length = 11)

why not have method_name = None and return either the generic or specific handler here? and have only one get_method_inbox() instead of 2?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
pierreca: this is the whole proxy/dto vs "smart" objects that we never really solved. historically, IoT SDKs lean towards having everythign on the client (that acts as "proxy" and having dumb DTO that only contain data. I think it could be a part of the user study feedback we want to ask. this has consequences:
- if we put everything on the client:
  - we also put accept/reject/abandon (C2D ack methods) on the client
  - we also put reported property updates of the twin of the client
- if we make "smarter" objects:
  - accept/reject/abandon have to live on the Message object
  - update reported properties have to live on the twin object
  - file upload notification have to live on a file upload request object

and so on. I for one could go either way, i just want consistency.

Another element for this debate:  if we go with a "pipeline" model for the client, then the proxy/dto thing might be easier to reason about. I would suggest talking to Cris Pop about this.

---
In reply to: [266138632](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266138632) [](ancestors = 266138632,266090705)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
```diff
@@ -151,6 +154,52 @@ def callback():
...
+        :param int status: The desired return status code for the method call.
+        """
+        logger.info("Sending method response to Hub...")
+        send_complete = Event()
+
+        def callback():
+            send_complete.set()
+            logger.info("Successfully sent method response to Hub")
+
+        # maybe consolidate method, result and status into a new object
```
pierreca: >maybe consolidate method, result and status into a new object [](start = 10, length = 61)

nit: if that is a thought for later please tag it "TODO"

azure-iot-hub-devicesdk/samples/receive_device_method.py
```diff
@@ -0,0 +1,83 @@
...
+
+# connect the client.
+device_client.connect()
+
+
+# define behavior for handling methods
+def method1_listener(device_client):
+    while True:
+        method_call = device_client.receive_method("method1")  # Wait for method1 calls
+        result = True  # set some result
```
pierreca: >True [](start = 17, length = 4)

I understand this might be hard to enforce in Python but I think to be valid a method response payload has to be valid json (to be confirmed). the samples should convey this

azure-iot-hub-devicesdk/samples/receive_device_method.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+generic_method_thread = threading.Thread(target=generic_method_listener, args=(device_client,))
+generic_method_thread.daemon = True
+generic_method_thread.start()
+
+
+# Wait for user to indicate they are done listening for messages
+while True:
+    selection = input("Press Q: Quit for exiting\n")
```
pierreca: >Quit for exiting [](start = 32, length = 16)

"Quit for exiting" sounds funny?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
```diff
@@ -0,0 +1,39 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module contains a class representing method calls that are received and responded to.
+"""
+
+
+class MethodCall(object):
```
pierreca: >MethodCall [](start = 6, length = 10)

on the device side i'd rather call this a MethodRequest (the "call" imho is on the service side). it's something we receive, not something we call.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, name, payload, response_timeout):
+        """Initializer for a MethodCall.
+
+        :param str name: The name of the method being called.
+        :param payload: The payload being sent with the call.
+        :param int response_timeout: The time in seconds before a response to the call is invalid.
+        """
+        self._name = name
+        self._payload = payload
+        self._response_timeout = response_timeout
```
pierreca: >self._response_timeout = response_timeout [](start = 8, length = 41)

the device does not see this (only the service that has a response timeout and a connect timeout). There is however a request_id that is used to match request and response.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
pierreca: >send_method_response [](start = 8, length = 20)

question: the success of methods depends on the device staying connected while it is processed by the user. if we get disconnected with stuff in the method inbox, how would we detect that and what should we do? probably clean up so that the user does not get notified of methods that are doomed to fail?
And then what happens if the user tries to respond to a method that has already failed because the device has been disconnected?

cartertinney: Yeah, I think for now the simplest thing to do is leave it on the client as is, and we can address how we want to deal with this in the (near) future. At the end of the day, I'm also on board for either approach as long as it's consistent.

---
In reply to: [266185870](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266185870) [](ancestors = 266185870,266138632,266090705)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
cartertinney: Yeah, request ID is missing, it was a (rather egregious) oversight on my part.

---
In reply to: [266186098](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266186098) [](ancestors = 266186098)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
cartertinney: These objects I'm talking about would all be user-facing, however, they wouldn't be exposed through this module, they'd be exposed through the package itself, so as far as how we organize them, that's just an internal implementation detail.

AFAIK there's no tree shaking tool or process for Python, at least, not that I've heard of.

---
In reply to: [266185461](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266185461) [](ancestors = 266185461,265780208)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/inbox_manager.py
cartertinney: Yeah, I also am not thrilled by this.

The reason we return a boolean rather than throw an exception is that this code is called in the transport, and the transport currently has no way to handle an exception in the routing of data to inboxes. And we definitely don't want to raise unhandled exceptions from the transport thread. Addressing this would require changes both in the client and the transport, so I'd rather do that in a separate PR tbh.

---
In reply to: [266185373](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266185373) [](ancestors = 266185373)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
cartertinney: This isn't my choice, it's black.

The line was over 100 characters. So black formatted it to use a different line. In doing so, it thus requires parenthesis  to keep it together (or explicit line breaks with '\', but black seems to shy away from those)

---
In reply to: [266185502](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266185502) [](ancestors = 266185502)

azure-iot-hub-devicesdk/samples/receive_device_method.py
cartertinney: We definitely can't enforce it, but if the expected payload is a json I can make the samples contain that.

---
In reply to: [266186000](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266186000) [](ancestors = 266186000)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/inbox_manager.py
cartertinney: Good call

---
In reply to: [266185617](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266185617) [](ancestors = 266185617)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
cartertinney: My issue with this is that we would then be passing the MethodRequest back into the client.send_method_response() API, which is bad semantics.

---
In reply to: [266186085](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266186085) [](ancestors = 266186085)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
cartertinney: I'm unclear what you mean by this. My understanding was as long as you've got a request ID, and the timeout hasn't occurred, that you'd be able to respond, as long as you had reconnected.  Can you clarify why items hanging in the inbox would be "doomed to fail"? If it's a question of timeouts, we ask the user to verify that the timeout hasn't already occurred.

---
In reply to: [266186191](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266186191) [](ancestors = 266186191)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
pierreca: technically the only thing we need to pass from the request into send_method_response is the request_id, not the full request object. we have to name this object for what it is, not what it gets passed to. You may go into other SDKs and check what they are doing and how they are naming things, but MethodCall doesn't feel nearly as accurate or used as MethodRequest IMHO.

---
In reply to: [266625788](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266625788) [](ancestors = 266625788,266186085)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
pierreca: still need to remove the response_timeout property. not closed.

---
In reply to: [266595247](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266595247) [](ancestors = 266595247,266186098)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
pierreca: sounds good (although i'd be careful with the use of the word "token" that has other meanings in the SDK. 

Also, process point: the commenter gets to close. if you don't want to change something, don't close it, wont-fix it instead please.

---
In reply to: [266114824](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266114824) [](ancestors = 266114824,266089272)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
pierreca: oh ok. thanks, i'm educated now :)

---
In reply to: [266598278](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266598278) [](ancestors = 266598278,266185502)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/sync_clients.py
pierreca: your understanding is incorrect. if a device is disconnected, methods immediately fail on the service side. if you reconnect and call with an old request id, it won't work.

---
In reply to: [266627364](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266627364) [](ancestors = 266627364,266186191)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/method_call.py
cartertinney: UPDATE: Instead of combining them into the same module, they've been combined into the same package. This helps with organization by not forcing them to be a part of the client layer, which was awkward because they were used in the transport as well (resulting in a somewhat circular-ish import)

---
In reply to: [266596563](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r266596563) [](ancestors = 266596563,266185461,265780208)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/data_structures/method_request.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    def __init__(self, request_id, name, payload):
+        """Initializer for a MethodRequest.
+
+        :param str request_id: The request id of the request.
+        :param str name: The name of the method to be invoked
+        :param payload: The payload being sent with the request.
+        """
+        self._request_id = request_id
```
cartertinney: This probably needs to be made public, but I'll make that change with the transport implementation when it becomes necessary.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -652,6 +662,27 @@ def _disable_c2d_messages(self, callback=None):
         self._trig_add_action_to_pending_queue(action)
         self.feature_enabled[constant.C2D_MSG] = False
 
+    def _enable_methods(self, callback=None, qos=1):
```
pierreca: >qos=1 [](start = 45, length = 5)

i don't think this should be a parameter. it should be "fixed" in a constant defined somewhere.
Also we should double-check with the service team that they expect qos 1 and not 0 here (ask Max Gortman)

azure-iot-hub-devicesdk/samples/receive_device_method.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+# define behavior for handling methods
+def method1_listener(device_client):
+    while True:
+        method_call = device_client.receive_method("method1")  # Wait for method1 calls
+        payload = json.dumps({"result": True, "data": "some data"})  # set response payload
+        status = 200  # set return status code
+        print("executed method1")
+        device_client.send_method_response(method_call, payload, status)  # send response
```
pierreca: >method_call [](start = 43, length = 11)

the more i see this the more i think this should be method_call.request_id rather than the full object.

azure-iot-hub-devicesdk/samples/receive_device_method_async.py
```diff
@@ -0,0 +1,88 @@
...
+                {"result": False, "data": "unknown method"}  # set response payload
+            )
+            status = 400  # set return status code
+            print("executed unknown method: " + method_call.name)
+            await device_client.send_method_response(method_call, payload, status)  # send response
+
+    # define behavior for halting the application
+    def stdin_listener():
+        while True:
+            selection = input("Press Q to exit\n")
```
pierreca: >exit [](start = 42, length = 4)

super super nitpicky: "Press Q to quit -> Quitting..." feels better than alternating between quitting or exiting. if you want to use "exit/exiting" everywhere you could also use spacebar or enter instead of q.

azure-iot-hub-devicesdk/tests/data_structures/test_method_request.py
```diff
@@ -0,0 +1,22 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+import pytest
+from azure.iot.hub.devicesdk.data_structures import MethodRequest
```
pierreca: >data_structures [](start = 29, length = 15)

ok you're gonna hate me for being nitpicky. "data structures" to me refers to "primary" data structures like arrays, lists, trees, etc.
I think what is contained in this package is types (I know this may sound egregious in python but it is what it is: a class definition creates a type).

azure-iot-hub-devicesdk/samples/receive_device_method.py
cartertinney: I'm not sure I can agree with that, but let's discuss after this PR.

---
In reply to: [267185229](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r267185229) [](ancestors = 267185229)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
cartertinney: I could get behind this, but only if it were transport wide.

---
In reply to: [267185061](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r267185061) [](ancestors = 267185061)

azure-iot-hub-devicesdk/tests/data_structures/test_method_request.py
cartertinney: I discussed this with Bert, and we think a better name is actually "common" so that it doesn't restrict us to just types. The whole purpose of this package is to provide common code for use between internal packages.

---
In reply to: [267185961](https://github.com/Azure/azure-iot-sdk-python-preview/pull/61#discussion_r267185961) [](ancestors = 267185961)

PR #63, opened 2019-03-15
 Refactor state machine to prepare for methods

BertKleewein:
This builds on top of my previous PR. (https://github.com/Azure/azure-iot-sdk-python-preview/pull/62) 

To prepare for methods, I'm doing the following:
1. Make subscribe and unsubscribe into generic functions
2. Make those functions into "Actions" which can be queued up in the state machine once the machine is connected.  
3. Refactor the previous "send-event" code to execute on any queued actions.  This currently includes subscribe, unsubscribe, send-event.  I have a stubbed-out send-method-response action.

This is pushing the state machine into a place where there are 4 states (disconnected, connecting, connected, and disconnecting), and all operations turn into actions that go into a queue that gets processed at the appropriate time.  This is a departure from how we did it with the node sdk, but I think it can solve the same goals (for MQTT at least) with much more simplicity.  For the record, I think we'll have to add new states when we implement Twin (because we will need subscribe to be blocking), but that's a story for another time.  

This gives Carter a base to build the Method code on top of.  It still has the following weaknesses:
1. thread safety is not guaranteed.  See my previous PR-62
2. Subscribe and unsubscribe are actions just like "send event" and "send method response" and, as such, can be overlapped with any other actions.  This overlap is OK for now, but won't be forever.
3. Subscribe and unsubscribe actions go into the queue as lifo operations.  They probably need to go first, so we know that subscribes are complete before we do anything else.
4. There are transitions missing for some edge-cases (send-event while disconnecting, etc), and error handling needs to be shored up.

### DIFF
--- a/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
+++ b/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
@@ -374,7 +374,7 @@ class MQTTTransport(AbstractTransport):
         self._mqtt_provider.on_mqtt_published = self._on_provider_publish_complete
         self._mqtt_provider.on_mqtt_subscribed = self._on_provider_subscribe_complete
         self._mqtt_provider.on_mqtt_unsubscribed = self._on_provider_unsubscribe_complete
-        self._mqtt_provider.on_mqtt_message = self._on_provider_message
+        self._mqtt_provider.on_mqtt_message_received = self._on_provider_message_received_callback
 
     def _get_telemetry_topic(self):
         topic = "devices/" + self._auth_provider.device_id
--- a/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
+++ b/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
@@ -1,4 +1,4 @@
-# -------------------------------------------------------------------------
+# --------------------------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
@@ -28,6 +28,35 @@ TOPIC_POS_MODULE = 6
 TOPIC_POS_INPUT_NAME = 5
 
 
+class TransportAction:
+    pass
+
+
+class SendMessageAction(TransportAction):
+    def __init__(self, message, callback):
+        self.message = message
+        self.callback = callback
+
+
+class SubscribeAction(TransportAction):
+    def __init__(self, topic, qos, callback):
+        self.topic = topic
+        self.qos = qos
+        self.callback = callback
+
+
+class UnsubscribeAction(TransportAction):
+    def __init__(self, topic, callback):
+        self.topic = topic
+        self.callback = callback
+
+
+class MethodReponseAction(TransportAction):
+    def __init__(self, method_response, callback):
+        self.method_response = method_response
+        self.callback = callback
+
+
 class MQTTTransport(AbstractTransport):
     def __init__(self, auth_provider):
         """
@@ -39,13 +68,10 @@ class MQTTTransport(AbstractTransport):
         self._mqtt_provider = None
         self.on_transport_connected = None
         self.on_transport_disconnected = None
-        self.on_event_sent = None
-        self._event_queue = queue.LifoQueue()
-        self._event_callback_map = {}
+        self._action_queue = queue.Queue()
+        self._callback_map = {}
         self._connect_callback = None
         self._disconnect_callback = None
-        self._subscribe_callback = None
-        self._unsubscribe_callback = None
 
         self._c2d_topic = None
         self._input_topic = None
@@ -64,7 +90,7 @@ class MQTTTransport(AbstractTransport):
                 "trigger": "_trig_provider_connect_complete",
                 "source": "connecting",
                 "dest": "connected",
-                "after": "_publish_events_in_queue",
+                "after": "_do_actions_in_queue",
             },
             {
                 "trigger": "_trig_disconnect",
@@ -83,22 +109,22 @@ class MQTTTransport(AbstractTransport):
                 "dest": "disconnected",
             },
             {
-                "trigger": "_trig_send_event",
+                "trigger": "_trig_queue_action",
                 "source": "connected",
-                "before": "_add_event_to_queue",
+                "before": "_add_action_to_queue",
                 "dest": None,
-                "after": "_publish_events_in_queue",
+                "after": "_do_actions_in_queue",
             },
             {
-                "trigger": "_trig_send_event",
+                "trigger": "_trig_queue_action",
                 "source": "connecting",
-                "before": "_add_event_to_queue",
+                "before": "_add_action_to_queue",
                 "dest": None,
             },
             {
-                "trigger": "_trig_send_event",
+                "trigger": "_trig_queue_action",
                 "source": "disconnected",
-                "before": "_add_event_to_queue",
+                "before": "_add_action_to_queue",
                 "dest": "connecting",
                 "after": "_call_provider_connect",
             },
@@ -112,32 +138,6 @@ class MQTTTransport(AbstractTransport):
                 "trigger": "_trig_on_shared_access_string_updated",
                 "source": ["disconnected", "disconnecting"],
                 "dest": None,
-            },
-            {
-                "trigger": "_trig_enable_receive",
-                "source": "connected",
-                "dest": None,
-                "after": "_call_subscribe",
-            },
-            {"trigger": "_trig_enable_receive", "source": "connecting", "dest": None},
-            {
-                "trigger": "_trig_enable_receive",
-                "source": "disconnected",
-                "dest": "connecting",
-                "after": "_call_provider_connect",
-            },
-            {
-                "trigger": "_trig_disable_receive",
-                "source": "connected",
-                "dest": None,
-                "after": "_call_unsubscribe",
-            },
-            {"trigger": "_trig_disable_receive", "source": "connecting", "dest": None},
-            {
-                "trigger": "_trig_disable_receive",
-                "source": "disconnected",
-                "dest": "connecting",
-                "after": "_call_provider_connect",
             },
         ]
 
@@ -234,22 +234,25 @@ class MQTTTransport(AbstractTransport):
         """
         Callback that is called by the provider when a publish operation is complete.
         """
-        if self.on_event_sent:
-            self.on_event_sent()
-        if mid in self._event_callback_map:
-            callback = self._event_callback_map[mid]
-            del self._event_callback_map[mid]
+        if mid in self._callback_map:
+            callback = self._callback_map[mid]
+            del self._callback_map[mid]
             callback()
+        else:
+            # TODO: tests for unkonwn MID cases
+            logger.warning("PUBACK received with unknown MID: %s", str(mid))
 
     def _on_provider_subscribe_complete(self, mid):
         """
         Callback that is called by the provider when a subscribe operation is complete.
         """
-        logger.info("_on_provider_subscribe_complete")
-        callback = self._subscribe_callback
-        if callback:
-            self._subscribe_callback = None
+        if mid in self._callback_map:
+            callback = self._callback_map[mid]
+            del self._callback_map[mid]
             callback()
+        else:
+            # TODO: tests for unkonwn MID cases
+            logger.warning("SUBACK received with unknown MID: %s", str(mid))
 
     def _on_provider_message_received_callback(self, topic, payload):
         """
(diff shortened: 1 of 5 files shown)


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -28,6 +28,35 @@
 TOPIC_POS_INPUT_NAME = 5
 
 
+class TransportAction:
```
cartertinney: Inherit from object (i.e. TransportAction(object)) for best compatibility

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -39,13 +68,10 @@ def __init__(self, auth_provider):
         self._mqtt_provider = None
         self.on_transport_connected = None
         self.on_transport_disconnected = None
-        self.on_event_sent = None
-        self._event_queue = queue.LifoQueue()
-        self._event_callback_map = {}
+        self._action_queue = queue.Queue()
```
cartertinney: Why the change from LIFO to FIFO? I agree that FIFO seems better, but I wonder why we were even using LIFO to begin with and what the motivation was.

BertKleewein: Lifo was a mistake.  It should have been fifo all along.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
...
+            topic = "TODO"
+            mid = self._mqtt_provider.publish(topic, action.method_response)
+            self._callback_map[mid] = action.callback
+
+        else:
+            logger.error("Removed unknown action type from queue.")
+
+    def _do_actions_in_queue(self, event_data):
         """
         Publish any events that are waiting in the event queue.  This function
```
cartertinney: >Publish any events that are waiting in the event queue. [](start = 8, length = 55)

Doc is outdated

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
             pass  # is there any other case
 
     def _on_provider_unsubscribe_complete(self, mid):
```
cartertinney: Missing doc

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
...
-        logger.info("Adding event to queue for later sending")
-        # TODO: throw here if event_data.args[0] is not an event/message
-        self._event_queue.put_nowait((event_data.args[0], event_data.args[1]))
+        action = event_data.args[0]
+        if isinstance(action, TransportAction):
+            self._action_queue.put_nowait(event_data.args[0])
+        else:
+            assert False
+
+    def _do_single_action(self, action):
```
cartertinney: missing doc (I know its pretty self explanatory, but consistency)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
...
                 return
-            logger.info("retrieved event from queue. publishing.")
-
-            encoded_topic = _encode_properties(message_to_send, self.topic)
 
-            mid = self._mqtt_provider.publish(encoded_topic, message_to_send.data)
-            self._event_callback_map[mid] = callback
+            self._do_single_action(action)
 
     def _create_mqtt_provider(self):
```
cartertinney: Missing doc

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -83,22 +109,22 @@ def __init__(self, auth_provider):
                 "dest": "disconnected",
             },
             {
-                "trigger": "_trig_send_event",
+                "trigger": "_trig_queue_action",
```
cartertinney: Naming. I'm not clear from this what it does, exactly.

From what I can tell from the code, and our previous discussion, it seems that this enqueues an action, then potentially pops an action off the queue to resolve depending on the state.

Perhaps a better name would just be _trig_action?

cartertinney:
>     def _on_provider_publish_complete(self, mid):

What is this parameter? It's undocumented

---
Refers to: azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py:233 in c351c16. [](commit_id = c351c16c962f8163ccbfe3300a1020b29cce83f2, deletion_comment = False)

cartertinney:
>     def _on_provider_publish_complete(self, mid):

since publish_event became do_action, are we still keeping the "publish" term around elsewhere? Seems asymmetrical.

---
Refers to: azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py:233 in c351c16. [](commit_id = c351c16c962f8163ccbfe3300a1020b29cce83f2, deletion_comment = False)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
...
+            self._action_queue.put_nowait(event_data.args[0])
+        else:
+            assert False
+
+    def _do_single_action(self, action):
+        if isinstance(action, SendMessageAction):
+            logger.info("running SendMessageAction")
+            message_to_send = action.message
+            base_topic = self._get_telemetry_topic()
+
```
cartertinney: I don't like the semantics of deriving a generic topic from a base of the telemetry topic.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
...
+            if isinstance(message_to_send, Message):
+                encoded_topic = _encode_properties(message_to_send, base_topic)
+            else:
+                encoded_topic = base_topic
+                message_to_send = Message(message_to_send)
```
cartertinney: Is this a thing we want to do in the transport, or in the client?

I've already implemented this Message wrapping in the client, but I could ditch it in favor of doing it here.

Alternatively, we could remove it here, and keep it in the client. What do you think the most appropriate place to do this is?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
...
+        if isinstance(action, TransportAction):
+            self._action_queue.put_nowait(event_data.args[0])
+        else:
+            assert False
+
+    def _do_single_action(self, action):
+        if isinstance(action, SendMessageAction):
+            logger.info("running SendMessageAction")
+            message_to_send = action.message
+            base_topic = self._get_telemetry_topic()
```
cartertinney: >base_topic [](start = 12, length = 10)

This isn't the base topic. It's the telemetry topic.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -349,44 +385,54 @@ def _get_telemetry_topic(self):
         topic += "/messages/events/"
         return topic
```
cartertinney: Refactor to use _get_base_topic, like how the other specific topics work

cartertinney:
> def transport(authentication_provider):

lets make this transport_device to make clear the distinction between this and transport_module

---
Refers to: azure-iot-hub-devicesdk/tests/transport/mqtt/test_mqtt_transport.py:73 in c351c16. [](commit_id = c351c16c962f8163ccbfe3300a1020b29cce83f2, deletion_comment = False)

cartertinney:
Overall, before I'm able to build off of this further, I'd like to see consistent documentation.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
pierreca: >TransportAction [](start = 6, length = 15)

all actions seem to have a callback - should that be part of the TransportAction contract?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
pierreca: sounds reasonable if the goal is to guarantee message order - is this why? 
A small comment about the design choice would protect us from asking again in the future.

---
In reply to: [266153947](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#discussion_r266153947) [](ancestors = 266153947)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
pierreca: also, it sounds like we are mapping actions to MQTT Packets types. I wonder if the naming should reflect that.

---
In reply to: [266183861](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#discussion_r266183861) [](ancestors = 266183861)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -234,22 +234,25 @@ def _on_provider_publish_complete(self, mid):
...
-        if mid in self._event_callback_map:
-            callback = self._event_callback_map[mid]
-            del self._event_callback_map[mid]
+        if mid in self._callback_map:
+            callback = self._callback_map[mid]
+            del self._callback_map[mid]
             callback()
+        else:
+            # TODO: tests for unkonwn MID cases
+            logger.warning("PUBACK received with unknown MID: %s", str(mid))
```
pierreca: >PUBACK [](start = 28, length = 6)

might be a SUBACK or an UNSUBACK now as well, right?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -273,45 +276,77 @@ def _on_provider_message_received_callback(self, topic, payload):
...
+        this queue, even if they're going to be run immediately.
         """
-        logger.info("Adding event to queue for later sending")
-        # TODO: throw here if event_data.args[0] is not an event/message
-        self._event_queue.put_nowait((event_data.args[0], event_data.args[1]))
+        action = event_data.args[0]
+        if isinstance(action, TransportAction):
+            self._action_queue.put_nowait(event_data.args[0])
+        else:
+            assert False
```
pierreca: >assert False [](start = 12, length = 12)

I'm not sure I understand how we could end up here. if the goal is to provoke a crash that will encourage a customer to publish a bug report, i think we should add a debug statement for that "Oops! looks like you found a bug - please submit a github issue <...instructions...>"

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
pierreca: quite helpful to every potential future contributor. let's not block on this but let's keep track of this. thanks for pushing for this Carter!

---
In reply to: [266157401](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#discussion_r266157401) [](ancestors = 266157401)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -349,44 +385,54 @@ def _get_telemetry_topic(self):
...
+    def _get_base_topic(self):
+        if self._auth_provider.module_id:
+            return (
+                "devices/"
+                + self._auth_provider.device_id
+                + "/modules/"
+                + self._auth_provider.module_id
+            )
+        else:
+            return "devices/" + self._auth_provider.device_id
```
pierreca: nit: this feels a little convoluted/repetitive. how about:
```
base_topic = "devices/" + self._auth_provider.device_id
if self._auth_provider.module_id:
    base_topic += "/modules/" + self._auth_provider.module_id 
```

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -349,44 +385,54 @@ def _get_telemetry_topic(self):
...
 
     def disconnect(self, callback=None):
+        logger.info("disconnect called")
         self._disconnect_callback = callback
         self._trig_disconnect()
 
     def send_event(self, message, callback=None):
-        self._trig_send_event(message, callback)
+        action = SendMessageAction(message, callback)
+        self._trig_queue_action(action, self._action_queue)
```
pierreca: >self._action_queue [](start = 40, length = 18)

here you have a second parameter but not below. leftovers? i don't see code that changes based on a second parameter?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -39,13 +68,10 @@ def __init__(self, auth_provider):
         self._mqtt_provider = None
         self.on_transport_connected = None
         self.on_transport_disconnected = None
-        self.on_event_sent = None
-        self._event_queue = queue.LifoQueue()
-        self._event_callback_map = {}
+        self._action_queue = queue.Queue()
+        self._callback_map = {}
```
pierreca: If i understand correctly, _action_queue is for "pending" actions (ie not on the wire yet) and _callback_map is to store the callbacks of actions "in flight" (ie the packet has been sent and the ACK hasn't been received). I wonder:
- if the naming could convey this idea (provided that my understanding is correct)
- (less important) if we need one additional "action processor" data structure that groups together the pending action queue, the callback map, and the methods that act on those (_add_to_queue, _do_single_action and _do_actions_in_queue). would it make this easier to test?

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
BertKleewein: added callback.
Also, this is not a 1:1 mapping to packet types.  This is a higher level.  MethodResponse and SendMessage both result in PUBLISH operations.

---
In reply to: [266184101](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#discussion_r266184101) [](ancestors = 266184101,266183861)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
BertKleewein: Awesome!  I totally agree on both points, but I want to refactor the "action processor" in a separate commit.

---
In reply to: [266185036](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#discussion_r266185036) [](ancestors = 266185036)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
cartertinney: Adding onto this, why don't we raise an Exception here instead? After all, "assert False" raises an AssertionError, which isn't terribly informative (and is intended largely for testing) - we might as well choose what gets raised.

---
In reply to: [266184228](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#discussion_r266184228) [](ancestors = 266184228)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
BertKleewein: I think the client is the right place.  It has to happen for all transports.  I removed it from here.

---
In reply to: [266169151](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#discussion_r266169151) [](ancestors = 266169151)

BertKleewein:
>     def _on_provider_publish_complete(self, mid):

yes, "PUBLISH" is still an MQTT operation, so we do need to keep it around, but maybe in a more limited context.  Many of our actions will involve sending a PUBLISH to the hub.

---
In reply to: [473463623](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#issuecomment-473463623) [](ancestors = 473463623)

---
Refers to: azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py:233 in ac29e41. [](commit_id = ac29e41a97938e7fcd1192de7e581e1c571aaeae, deletion_comment = False)

BertKleewein:
> def transport(authentication_provider):

I like that idea, but that is more test churn than I want to deal with right now.  Next time we're poking around in here, we should change it.

---
In reply to: [473467073](https://github.com/Azure/azure-iot-sdk-python-preview/pull/63#issuecomment-473467073) [](ancestors = 473467073)

---
Refers to: azure-iot-hub-devicesdk/tests/transport/mqtt/test_mqtt_transport.py:73 in ac29e41. [](commit_id = ac29e41a97938e7fcd1192de7e581e1c571aaeae, deletion_comment = False)

PR #64, opened 2019-03-19
handle case where paho calls PUBACK, SUBACK, and UNSUBACK callbacks w…

BertKleewein:
…ith MID before returning that MID to us

This is a workaround for a particular behavior of Paho.  We previously coded for this order:
1. transport calls provider.subscribe
2. provider sends SUBSCRIBE package.
3. provider returns message ID (MID) for SUBSCRIBE operation to transport
4. provider received SUBACK and calls transport._on_provider_subscribe_complete(MID).
5. transport calls client callback for completed subscribe operation.

Sometimes, the SUBACK comes back so quickly that Paho swaps step 3 & 4 and does this:
1. transport calls provider.subscribe
2. provider sends SUBSCRIBE package.
3. provider received SUBACK and calls transport._on_provider_subscribe_complete(MID).
4. provider returns message ID (MID) for SUBSCRIBE operation to transport

Before this change, we would drop the SUBACK in step 3 because it happened before provider.subscribe() returned the MID to us.

This code adds a dictionary to keep track of "unkonwn" operations.  When we get a SUBACK for a MID that we haven't seen yet (because it hasn't been returned to us), we store it in the dictionary.  Then, when the MID is returned, we look in the dictionary to see if the operation is complete and call the callback if necessary.


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/abstract_transport.py
```diff
@@ -48,7 +48,7 @@ def disconnect(self, callback):
         pass
 
     @abc.abstractmethod
-    def enable_feature(self, feature_name, callback=None, qos=1):
```
pierreca: >, qos=1 [](start = 56, length = 7)

I agree with this (this should be decided at the feature level)  but I just want to make sure it's intentional given it seems unrelated.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -340,6 +345,9 @@ def _on_provider_publish_complete(self, mid):
         else:
             # TODO: tests for unkonwn MID cases
```
pierreca: >unkonwn [](start = 30, length = 7)

nit: typo (repeats below)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
```diff
@@ -156,6 +156,11 @@ def __init__(self, auth_provider):
         # used to call back into the caller to indicate that an action is complete.
         self._in_progress_actions = {}
 
+        # Map of responses we receive with a MID that is not in the _in_progress_actions map.
+        # We need this because sometimes a SUBSCRIBE or a PUBLISH will complete before the call
+        # to subscribe() or publish() returns.
+        self._responses_with_unknown_mid = {}
```
pierreca: I feel a bit weird about having a dictionary where the key equals the value just for the sake of finding it later (and I do get the O(1) niceness of it). it does feel like we should be able to handle everything with a modified in_progress_actions structure. 

here's a rough, pseudo-codish idea that takes the idea of an action, an inprogress collection and a way to deal with an action no matter what comes first. it's slightly less efficient because it uses an array (how big do we expect this to become though?) but it feels that it describes the intent a little more naturally (IMHO).

You could also do this with 2 internal dictionaries like you're doing here, but encapsulated in the InProgressActions class and still get the benefit of having a try_complete_action_with_mid() and try_complete_action_with_ack() methods. These 2 methods would save you a lot of repetition around lines 445-ish where you systematically have to test the unkown_mid dictionary key existence.

Also: i wonder if this (my pseudocode and your code) should be more try/except based (with KeyNotFound errors) than if..else-ish (i legitimately do not know what would be more pythonic)

```python
class Action:
  def __init__():
    mid: None;
    callback: None;
    ack: None;

class InProgressActions:
  def __init__():
    self.actions = []

  def try_complete_action_with_mid(self, mid, callback):
    for action in self.actions:
      if action.ack && action.ack.mid == mid:
        self.actions.remove(action)
        callback()
      else:
        self.actions.append(Action(mid, callback))
  
  def try_complete_action_with_ack(self, ack):
    for action in self.actions:
      if action.mid == ack.mid:
        self.actions.remove(action)
        action.callback()
      else:
        self.actions.append(Action.from_ack(ack))


def class Client:
  def __init__():
    self.in_progress = InProgressActions()
    self.mqtt_client = ...
    self.mqtt_client.on_ack_received = self.on_ack_received


  def send(self, msg, callback):
    mid = self.mqtt_client.publish(msg)
    in_progress.try_complete_action_with_mid(mid, callback)

  def on_ack_received(self, ack):
    in_progress.try_complete_action_with_ack(ack)
```

pierreca:
code seems to work so let's not block - i do feel the pattern could be improved a bit - please consider my suggestions. if not implemented now, should we revisit this later? if yes, let's add a work item.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/abstract_transport.py
BertKleewein: This was intentional, but for a slightly different reason -- QOS is an MQTT concept and this is an interface that is common for all transports.

---
In reply to: [267173737](https://github.com/Azure/azure-iot-sdk-python-preview/pull/64#discussion_r267173737) [](ancestors = 267173737)

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_transport.py
BertKleewein: Good idea, but not now.  added post-it to "make it right" section of board.

---
In reply to: [267178965](https://github.com/Azure/azure-iot-sdk-python-preview/pull/64#discussion_r267178965) [](ancestors = 267178965)

PR #70, opened 2019-03-22
raise exceptions on paho failures and enable paho logs

BertKleewein:
Paho does not always throw on failure like you would expect, so i added explicit throws.  I also added a logger so we could see Paho logs -- though there's not much.
This doesn't fix our random disconnect, but it does make debugging more pleasant.

### DIFF
--- a/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
+++ b/azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
@@ -48,6 +48,8 @@ class MQTTProvider(object):
 
         self._mqtt_client = mqtt.Client(self._client_id, False, protocol=mqtt.MQTTv311)
 
+        self._mqtt_client.enable_logger(logging.getLogger("paho"))
+
         def on_connect_callback(client, userdata, flags, result_code):
             logger.info("connected with result code: %s", str(result_code))
             # TODO: how to do failed connection?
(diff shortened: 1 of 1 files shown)


azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
```diff
@@ -48,6 +48,8 @@ def _create_mqtt_client(self):
 
         self._mqtt_client = mqtt.Client(self._client_id, False, protocol=mqtt.MQTTv311)
 
+        self._mqtt_client.enable_logger(logging.getLogger("paho"))
```
pierreca: >self._mqtt_client.enable_logger(logging.getLogger("paho")) [](start = 8, length = 58)

while doing this let's keep an eye out to make sure paho doesn't leak credentials in the logs.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
```diff
@@ -157,6 +159,8 @@ def publish(self, topic, message_payload):
         """
         logger.info("sending")
         message_info = self._mqtt_client.publish(topic=topic, payload=message_payload, qos=1)
+        if message_info.rc != mqtt.MQTT_ERR_SUCCESS:
+            raise Exception(mqtt.error_string(message_info.rc))
```
cartertinney: >Exception [](start = 18, length = 9)

Can we define a custom Exception for this? Otherwise, there's no good way to catch it.

olivakar:
Are we planning on catching these exceptions ?

BertKleewein:
That’s a good point.  I’ll have to take a good look at the stack and make sure we catch these properly so they get bubbled up to the user.  I added a sticky to the board for this.

azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/mqtt/mqtt_provider.py
BertKleewein: Good point.  I added a sticky to define a custom exception for this.

PR #77, opened 2019-03-29
Namespace adjustments

cartertinney:
* azure-iot-hub-devicesdk -> azure-iot-device
* azure-iot-common has been made into a subpackage of azure-iot-device (for now)


azure-iot-device/azure/iot/device/__init__.py
```diff
@@ -0,0 +1,7 @@
+""" Azure IoT Device Library
+
+This library provides clients and associated models for communicating with Azure IoT services
+from an IoT device.
+"""
+
+from .iothub import *
```
BertKleewein: We seem to be inconsistent about where we manually set __all__ in __init__.py.  I know that we don't set it on internal modules.  It looks like we only set it on modules that expose public namespaces.  Why don't we set it here?

olivakar:
> class AMQPWrapper(object):

may be this got left behind...could you delete this before release ?

---
Refers to: azure-iot-device/azure/iot/device/iothub/transport/amqp/amqp_wrapper.py:8 in ffd4700. [](commit_id = ffd4700e5537021cb6351e809734bbe64a082150, deletion_comment = False)

azure-iot-device/azure/iot/device/__init__.py
cartertinney: I thought this might come up. Bear with me here, there's quite a bit of explanation.

`__all__`  is really protection against anyone who does an `import *` to prevent the exposure of internal stuff. `import *` is generally a bad idea, but users *love* doing it. It's especially bad for us, since without `__all__` protection, 2.7 users will import the `aio` subpackage and then hit a runtime error. I also realize that I'm saying `import *` is  a bad idea while in this very file doing it, but I have a reason here!

The `azure.iot.device` namespace represents all code that is run on the device-side of the IoT relationship. We've structured the namespace from the perspective of a cloud application developer, choosing to bundle all device-side stuff together in one namespace, since, from the perspective of a cloud application, that's all just stuff that makes the products work (if that makes sense). The final namespaces will be (approximately):

`azure.iot.hub`
`azure.iot.provisioning`
`azure.iot.device` 
`azure.iot.digitaltwin`

Note how the service clients get their own namespaces, but the device stuff is all bundled together.

However, there's no getting around that there are distinct subpackages within azure.iot.device. Sure, stuff is shared through `azure.iot.device.common` (which isn't exposed), but by and large, `azure.iot.device.iothub` is separate from DPS which would be in `azure.iot.device.provisioning`. Each of them will expose their own API that's unique to that client subpackage. HOWEVER, we also don't want users having to go to `azure.iot.device.iothub` and `azure.iot.device.provisioning`, so we're exposing each of those through the `__init__` of azure.iot.device (i.e. this file).

Since the `__init__` of `azure.iot.device.iothub` already exposes the full IoTHubClient API though, it doesn't make sense to replicate all the imports from there, because:

A) it creates additional maintenance, AND maintenance that is easy to forget
B) it involves going through specific internal modules of a subpackage, which kind of defeats the whole purpose of them being a subpackage
C) especially in the `azure.iot.device.aio` namespace, you'd be referencing specific modules of not a subpackage, but a subpackage of a sibling package (azure.iot.device.iothub.aio). Really messy!!!

Thus, the philosophy to the structure here is to bottleneck the imports as soon as it makes sense to do so, in order to keep complexity and duplication as low as possible. And the best place to bottleneck is on the subpackages. Thus, with the `__all__` safely defined for each architecturally unique subpackage (`azure.iot.device.iothub` and `azure.iot.device.provisioning`), it is simplest, and totally safe to just do an `import *` on those subpackages here.

I hope that all makes sense. Feel free to ask me any questions in person if there's any confusion whatsoever. I put.... a lot of thought into this.

---
In reply to: [275532204](https://github.com/Azure/azure-iot-sdk-python-preview/pull/77#discussion_r275532204) [](ancestors = 275532204)

cartertinney:
> class AMQPWrapper(object):

This has been around forever. But it probably does make sense to remove, yeah. Let me check with Pierre

---
In reply to: [483425424](https://github.com/Azure/azure-iot-sdk-python-preview/pull/77#issuecomment-483425424) [](ancestors = 483425424)

---
Refers to: azure-iot-hub-devicesdk/azure/iot/hub/devicesdk/transport/amqp/amqp_wrapper.py:8 in ffd4700. [](commit_id = ffd4700e5537021cb6351e809734bbe64a082150, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -9,15 +9,15 @@
...
 from .transport import constant
-from .common import Message
+from .models import Message
 from .inbox_manager import InboxManager
 from .sync_inbox import SyncClientInbox
 
 logger = logging.getLogger(__name__)
 
-__all__ = ["DeviceClient", "ModuleClient"]
+__all__ = ["IoTHubDeviceClient", "IoTHubModuleClient"]
```
olivakar: is this all and the other one in "init" both needed ?

cartertinney: removed

---
In reply to: [275560460](https://github.com/Azure/azure-iot-sdk-python-preview/pull/77#discussion_r275560460) [](ancestors = 275560460)

PR #82, opened 2019-04-22
Transport/Provider refactor

cartertinney:
* Moved all reference to Message ID (a Paho specific concept) to the Provider
* Added callbacks parameters to some Provider APIs
* Improved test coverage in the Provider
* Added exception handling to callbacks in Provider
* Updated pytest-testdox requirement due to bugfix
* Improved documentation in Provider

### DIFF
--- a/azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_provider.py
+++ b/azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_provider.py
@@ -14,16 +14,23 @@ logger = logging.getLogger(__name__)
 
 class MQTTProvider(object):
     """
-    A wrapper over the actual implementation of mqtt message broker which will eventually connect to an mqtt broker
-    to publish/subscribe messages.
+    A wrapper class that provides an implementation-agnostic MQTT message broker interface.
+
+    :ivar on_mqtt_connected: Event handler callback, called upon establishing a connection.
+    :type on_mqtt_connected: Function
+    :ivar on_mqtt_disconnected: Event handler callback, called upon a disconnection.
+    :type on_mqtt_disconnected: Function
+    :ivar on_mqtt_message_received: Event handler callback, called upon receiving a message.
+    :type on_mqtt_message_received: Function
     """
 
     def __init__(self, client_id, hostname, username, ca_cert=None):
         """
         Constructor to instantiate a mqtt provider.
-        :param client_id: The id of the client connecting to the broker.
-        :param hostname: hostname or IP address of the remote broker.
-        :param ca_cert: Certificate which can be used to validate a server-side TLS connection.
+        :param str client_id: The id of the client connecting to the broker.
+        :param str hostname: Hostname or IP address of the remote broker.
+        :param str username: Username for login to the remote broker.
+        :param str ca_cert: Certificate which can be used to validate a server-side TLS connection (optional).
         """
         self._client_id = client_id
         self._hostname = hostname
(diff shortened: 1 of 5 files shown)


azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_transport.py
```diff
@@ -329,84 +323,32 @@ def _on_provider_disconnect_complete(self):
...
         a method invocation.  This function needs to decide what kind of message it is based on the topic name and
         take the correct action.
 
-        :param topic: MQTT topic name that the message arrived on
-        :param payload: Payload of the message
+        :param str topic: MQTT topic name that the message arrived on
+        :param bytes payload: Payload of the message as a bytestring
         """
-        logger.info("Message received on topic %s", topic)
+        logger.info("Message received on topic {}".format(topic))
```
BertKleewein: >logger.info("Message received on topic {}".format(topic)) [](start = 8, length = 57)

I'm curious why you changed this.  Is this personal preference or do you have a specific reason?

azure-iot-device/tests/iothub/transport/mqtt/test_mqtt_provider.py
```diff
@@ -8,164 +8,1120 @@
...
+
+@pytest.fixture
+def provider(mock_mqtt_client):
+    # Implicitly imports the mocked Paho MQTT Client from mock_mqtt_client
+    return MQTTProvider(client_id=fake_device_id, hostname=fake_hostname, username=fake_username)
+
+
+@pytest.mark.describe("MQTT Provider - Instantiation")
+class TestInstantiation(object):
+    @pytest.mark.it("Creates an instance of the Paho MQTT Client")
```
BertKleewein: >@pytest.mark.it("Creates an instance of the Paho MQTT Client") [](start = 4, length = 62)

Awesome!  I love this.  using the pytest.mark.it decorator is so much better than having the test name generated from the name of the test function.

azure-iot-device/tests/iothub/transport/mqtt/test_mqtt_provider.py
```diff
@@ -8,164 +8,1120 @@
...
+
+    @pytest.mark.it("Reconnects with Paho")
+    def test_calls_paho_reconnect(self, mocker, mock_mqtt_client, provider):
+        provider.reconnect(fake_password)
+
+        assert mock_mqtt_client.reconnect.call_count == 1
+        assert mock_mqtt_client.reconnect.call_args == mocker.call()
+
+    @pytest.mark.it(
+        "Triggers on_mqtt_connected event handler callback upon user-driven reconnect completion"
```
BertKleewein: >upon user-driven reconnect completion [](start = 59, length = 37)

I'm not sure if the phrase "user-driven" belongs in here.  The test is simulating a reconnect completion, but I wouldn't call that "user-driven".

BertKleewein: ah-ha,  it should say "upon completion of user-driven reconnect" to show that the reconnect is user-driven.  With the existing wording, I though the completion was user-driven

---
In reply to: [277847664](https://github.com/Azure/azure-iot-sdk-python-preview/pull/82#discussion_r277847664) [](ancestors = 277847664)

azure-iot-device/tests/iothub/transport/mqtt/test_mqtt_provider.py
```diff
@@ -8,164 +8,1120 @@
...
+
+    @pytest.mark.it("Stops MQTT Network Loop")
+    def test_calls_loop_stop(self, mocker, mock_mqtt_client, provider):
+        provider.disconnect()
+
+        assert mock_mqtt_client.loop_stop.call_count == 1
+        assert mock_mqtt_client.loop_stop.call_args == mocker.call()
+
+    @pytest.mark.it(
+        "Triggers on_mqtt_disconnected event handler callback upon user-driven disconnect completion"
```
BertKleewein: >user-driven [](start = 67, length = 11)

same comment on wording.  I suggest searching the entire file for "user-driven" and "externally-driven" if you're going to change this.  I won't add any more comments on the matter.

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_provider.py
```diff
@@ -132,52 +154,108 @@ def connect(self, password):
...
 
     def disconnect(self):
         """
-        This method disconnects the mqtt provider. This should be called from the upper transport
-        when it wants to disconnect from the mqtt provider.
+        Disconnect from the MQTT broker.
         """
         logger.info("disconnecting transport")
         self._mqtt_client.disconnect()
+        self._mqtt_client.loop_stop()
```
BertKleewein: >loop_stop [](start = 26, length = 9)

Did you have a reason for adding this back in?  Is this required or did it just seem like a good idea?  
I'm asking because I've seen this hang before (it waits for all IO to flush before returning).  I think I fixed the deadlock that was causing the hang, but I'm not sure what I think about adding this back in.  I do know that the loop runs in a daemon thread, and I do know that it's harmless to call loop_start multiple times.
I think we should maybe leave it in because I think I fixed the deadlock, but we should be aware that it might hang.

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_transport.py
cartertinney: .format is considered to be the more modern way to do it. It's more flexible and avoids a lot of problems related to the %s approach. In this context, there's no real difference, it was more of a consistency thing. Since I learned the convention is now to use format, I've been trying to switch over to that exclusively.

---
In reply to: [277841533](https://github.com/Azure/azure-iot-sdk-python-preview/pull/82#discussion_r277841533) [](ancestors = 277841533)

azure-iot-device/tests/iothub/transport/mqtt/test_mqtt_provider.py
cartertinney: Shout out to the two guys maintaining pytest-testdox for pushing through features and fixes when I ask.

---
In reply to: [277844525](https://github.com/Azure/azure-iot-sdk-python-preview/pull/82#discussion_r277844525) [](ancestors = 277844525)

azure-iot-device/tests/iothub/transport/mqtt/test_mqtt_provider.py
cartertinney: Good point. I'll make this change here and everywhere.

---
In reply to: [277848527](https://github.com/Azure/azure-iot-sdk-python-preview/pull/82#discussion_r277848527) [](ancestors = 277848527,277847664)

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_provider.py
cartertinney: I didn't add it back in for any reason other than, it seemed missing based on the docs I was looking at, and there were references to it in docs/tests.

I'm open to re-removing it, but I haven't seen any issues. I say let's keep it for now, and be aware of the potential issue.

---
In reply to: [277871470](https://github.com/Azure/azure-iot-sdk-python-preview/pull/82#discussion_r277871470) [](ancestors = 277871470)

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_provider.py
```diff
@@ -33,97 +40,112 @@ def __init__(self, client_id, hostname, username, ca_cert=None):
...
+        Connect to the MQTT broker, using hostname and username set at instantiation.
+
         This method should be called as an entry point before sending any telemetry.
+
+        :param str password: The password for connecting with the MQTT broker.
         """
         logger.info("connecting to mqtt broker")
 
-        ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
+        ssl_context = ssl.SSLContext(protocol=ssl.PROTOCOL_TLSv1_2)
```
olivakar: >ssl_context = ssl.SSLContext(protocol=ssl.PROTOCOL_TLSv1_2) [](start = 8, length = 59)

Are we going to follow this pattern of specifying parameter name when calling methods with 1 required positional param ?
The guidelines say to do this only with more than 2.

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_transport.py
```diff
@@ -329,84 +323,32 @@ def _on_provider_disconnect_complete(self):
...
         message_received = Message(payload)
         # TODO : Discuss everything in bytes , need to be changed, specially the topic
```
olivakar: ># TODO : Discuss everything in bytes , need to be changed, specially the topic [](start = 8, length = 78)

u can remove this TODO. I kept it because of the `_topic` mistaken use.

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_provider.py
cartertinney: I honestly don't know why I did this. You're right, this isn't the pattern.

---
In reply to: [278271073](https://github.com/Azure/azure-iot-sdk-python-preview/pull/82#discussion_r278271073) [](ancestors = 278271073)

PR #86, opened 2019-04-27
Direct Methods

cartertinney:
* Implemented receive_method_request and send_method_response
* Added testing **except for transport tests re: sending method response**
    * Blocked on pending pipeline test PR #89 
* Improved testing for Inbox Manager (and was able to unify the tests wrt SyncClientInboxes and AsyncClientInboxes!!!!!)


azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
```diff
@@ -334,17 +345,24 @@ def _on_provider_message_received_callback(self, topic, payload):
...
-            _extract_properties(topic_parts[TOPIC_POS_MODULE], message_received)
+            _extract_message_properties(topic_parts[TOPIC_POS_MODULE], message_received)
             self.on_transport_input_message_received(input_name, message_received)
         elif _is_c2d_topic(topic):
-            _extract_properties(topic_parts[TOPIC_POS_DEVICE], message_received)
+            message_received = Message(payload)
+            _extract_message_properties(topic_parts[TOPIC_POS_DEVICE], message_received)
             self.on_transport_c2d_message_received(message_received)
+        elif _is_method_topic(topic):
+            rid = topic_parts[TOPIC_POS_REQUEST_ID].split("=")[1]
```
pierreca: >TOPIC_POS_REQUEST_ID [](start = 30, length = 20)

it's dangerous to function on position for properties going after the '?' character. in fact we just had a bug about that in the node DPS sdk.
what needs to happen is everything after the '?' must be treated as key/value pairs that could appear in any order.
I would suggest adding a fix for this before merging, or fix it immediately after.

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
```diff
@@ -488,6 +508,20 @@ def _get_input_topic_for_subscribe(self):
         """
         return self._get_topic_base() + "/inputs/#"
 
+    def _get_method_topic_for_subscribe(self):
+        """
+        :return: The topic for ALL incoming methods. It is of the format
+        "$iothub/methods/POST/#"
+        """
+        return "$iothub/methods/POST/#"
```
pierreca: >$iothub/methods/POST/# [](start = 16, length = 22)

there's little repetition so i'm not too worried about this but i think it'd be nice to have all topics grouped in a set of constants co-located together. for when we need to add/change/check a topic it's easier to know where it's at and how it compares to other topics. also a nice self-documenting thing.

azure-iot-device/samples/advanced-hub-scenarios/receive_direct_method.py
```diff
@@ -31,35 +31,42 @@
...
+            payload = {"result": True, "data": "some data"}  # set response payload
             status = 200  # set return status code
             print("executed method1")
-            await device_client.send_method_response(
-                method_request, payload, status
-            )  # send response
+            method_response = MethodResponse.create_from_method_request(
+                method_request, status, payload
+            )
+            await device_client.send_method_response(method_response)  # send response
```
pierreca: seeing this, i think having to have 2 steps there is a little jarring (i apologize if it was my idea and i'm changing my mind...). 
i'd rather send_method_response take just request_id, status and payload. 

i understand having a send_response method on an object called request might be weird too (and might create its own set of weird dependency/garbage collection problems)

maybe a source of inspiration would be to see how http servers serve http request/responses?

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
cartertinney: The issue is not all are constants. Most are dynamically created. That said, I think there's an argument for defining constant format strings with the dynamic keywords, as in the function below.

---
In reply to: [280650160](https://github.com/Azure/azure-iot-sdk-python-preview/pull/86#discussion_r280650160) [](ancestors = 280650160)

azure-iot-device/samples/advanced-hub-scenarios/receive_direct_method.py
cartertinney: There at least needs to be a Method Response object under the covers so the data can be wrapped up together to be used by lower layers of the stack, but I'm open to changing the API. I'll take a look at how some other libraries work, and we can revisit this before the merge.

I'm fairly against having "smart objects" at the moment, given the significant complexity that results from implementing them. But if we do decide to eventually go down that route, it would have to be an across the board change, we can't just do it for Methods, it has to be done for Messages, and any other objects in the future (Twin?).

---
In reply to: [280651175](https://github.com/Azure/azure-iot-sdk-python-preview/pull/86#discussion_r280651175) [](ancestors = 280651175)

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
```diff
@@ -0,0 +1,787 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
```
BertKleewein: It looks like this was a merge mistake.  This file moved and changed significantly from this version.

azure-iot-device/azure/iot/device/iothub/models/methods.py
```diff
@@ -0,0 +1,70 @@
...
+    :ivar int status: The status of the execution of the MethodRequest.
+    :ivar dict payload: The JSON payload to be sent with the response.
+    """
+
+    def __init__(self, request_id, status, payload=None):
+        """Initializer for MethodResponse.
+
+        :param str request_id: The request id of the MethodRequest being responded to.
+        :param int status: The status of the execution of the MethodRequest.
+        :param dict payload: The JSON payload to be sent with the response. (OPTIONAL)
```
BertKleewein: >dict [](start = 15, length = 4)

json != dict.  json can also be an array, string, number, "true", "false", or "null".  ref: json.org

azure-iot-device/azure/iot/device/iothub/transport/pipeline_ops_iothub.py
```diff
@@ -97,3 +97,26 @@ def __init__(self, message, callback=None):
         super(SendOutputEvent, self).__init__(callback=callback)
         self.message = message
         self.needs_connection = True
+
+
+class SendMethodResponse(PipelineOperation):
+    """
```
BertKleewein: >   """ [](start = 1, length = 6)

odd indents in this function

azure-iot-device/azure/iot/device/iothub/transport/mqtt/pipeline_stages_iothub_mqtt.py
```diff
@@ -72,6 +73,17 @@ def _run_op(self, op):
                 new_op=pipeline_ops_mqtt.Publish(topic=topic, payload=op.message.data),
             )
 
+        elif isinstance(op, pipeline_ops_iothub.SendMethodResponse):
+            # Sending a Method Response gets translated into an MQTT Publish operation
+            topic = mqtt_topic.get_method_topic_for_publish(
+                op.method_response.request_id, str(op.method_response.status)
+            )
+            payload = json.dumps(op.method_response.payload)
+            # TODO: Why is there no callback on the continue?
```
BertKleewein: ># TODO: Why is there no callback on the continue? [](start = 12, length = 49)

it uses the callback in the original_op

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_topic.py
```diff
@@ -46,6 +46,24 @@ def get_input_topic_for_subscribe(device_id, module_id):
     return _get_topic_base(device_id, module_id) + "/inputs/#"
 
 
+def get_method_topic_for_publish(request_id, status):
+    """
+    :return: The topic for publishing method responses. It is of the format
+    "$iothub/methods/res/<status>/?$rid=<requestId>
+    """
+    return "$iothub/methods/res/{status}/?$rid={rid}".format(
+        status=urllib.parse.quote(status), rid=urllib.parse.quote(request_id)
```
BertKleewein: >status=urllib.parse.quote(status), rid=urllib.parse.quote(request_id) [](start = 8, length = 69)

you might be right about uri-encoding here, but this makes me wonder.  status is an integer, and request_id needs to be mirrored from the request.  encoding status is, at worst, an unnecessary function call, but encoding request_id _might_ be broken if urllib.parse.quote isn't an exact inverse of unquote_plus (which is used to extract the rid from the request).  (this assumes request id's can be arbitrary, but they're defined by the iothub service, so it may never _need_ encoding/decoding)

azure-iot-device/azure/iot/device/iothub/models/methods.py
cartertinney: Yeah, sometimes it wouldn't be a dictionary. The type is whatever the result of json.loads is. I'll correct the docs

---
In reply to: [282743221](https://github.com/Azure/azure-iot-sdk-python-preview/pull/86#discussion_r282743221) [](ancestors = 282743221)

azure-iot-device/azure/iot/device/iothub/transport/pipeline_ops_iothub.py
cartertinney: Good catch. My indenting got automatically messed up due to the double spaces you added on continued docstring lines elsewhere in the file. I believe the MS docgen convention is to only use one space for that. I'll fix this, and also the other docs that caused it.

---
In reply to: [282743332](https://github.com/Azure/azure-iot-sdk-python-preview/pull/86#discussion_r282743332) [](ancestors = 282743332)

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_topic.py
cartertinney: I'll switch them to both to use the quote_plus variant. 

I agree that the field is probably just fine on it's own (rids I think are just integers) and there's probably no need to ever do a "quote" on it, but since it's being decoded with unquote_plus due to being in a function that has to be more generic, it makes sense to keep it all symmetrical.

We could **probably** drop the quote on status, but I also just generally think url encoding/decoding everything going into a URI is generally good practice - not just from a defensive coding perspective, but also because it makes it easier to notice when url encoding is missing, and correct it.

i.e. If I'm scanning a file, and I notice something in a uri isn't being encoded/decoded, I know it should be fixed - I don't have to think about "should this particular field I may not know about be encoded/decoded?". This prevents any bugs with things that DO need to be encoded/decoded elsewhere from slipping through.

Open to doing it differently though, let me know.

---
In reply to: [282744164](https://github.com/Azure/azure-iot-sdk-python-preview/pull/86#discussion_r282744164) [](ancestors = 282744164)

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
cartertinney: Good catch, I'll scrap it. I knew something like this would happen.

---
In reply to: [282742575](https://github.com/Azure/azure-iot-sdk-python-preview/pull/86#discussion_r282742575) [](ancestors = 282742575)

azure-iot-device/samples/advanced-hub-scenarios/receive_direct_method.py
cartertinney: Follow up: I've been looking into alternatives in other libraries, but the issue is that how we handle request response is a bit weird due to how the IoTHub implementation of methods is - it doesn't map particularly well to more traditional request/response paradigms, especially since a lot of those are built around fundamental concepts of HTTP.

I think we should punt this down the road a bit. Everyone had signed off on this API over the various other samples that were presented as options, and if we want to revisit the design, I think it's a decision I'd like everyone to be a part of.

---
In reply to: [281713995](https://github.com/Azure/azure-iot-sdk-python-preview/pull/86#discussion_r281713995) [](ancestors = 281713995,280651175)

PR #87, opened 2019-04-29
convert mqtt transport into a pipeline

BertKleewein:
Lots of changes here.  Details coming in email.


azure-iot-device/azure/iot/device/iothub/auth/sk_authentication_provider.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         shared_access_key,
         shared_access_key_name=None,
         gateway_hostname=None,
+        connection_string=None,
```
BertKleewein: >connection_string=None, [](start = 8, length = 23)

this needs to be removed.  It is no longer necessary

azure-iot-device/samples/simple_send_telemetry.py
```diff
@@ -4,11 +4,14 @@
...
 # --------------------------------------------------------------------------
 
+import logging
 import os
 import asyncio
 from azure.iot.device.aio import IoTHubDeviceClient
 from azure.iot.device import auth
 
+logging.basicConfig(level=logging.INFO)
+
```
BertKleewein: all changes in this file need to be removed

azure-iot-device/azure/iot/device/common/transport/mqtt/test.py
```diff
@@ -0,0 +1,43 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import os
```
BertKleewein: This entire file needs to be removed from the commit

azure-iot-device/azure/iot/device/iothub/auth/base_renewable_token_authentication_provider.py
```diff
@@ -161,6 +161,7 @@ def timerfunc():
             self.generate_new_sas_token()
 
         self._token_update_timer = Timer(seconds_until_update, timerfunc)
+        self._token_update_timer.daemon = True
```
BertKleewein: >self._token_update_timer.daemon = True [](start = 8, length = 38)

as a side-effect, we can remove the disconnect function from this class.  The only reason for disconnect to exist was to kill the timer, but now we don't have to worry about this.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+from . import pipeline_events
+from . import mqtt_topic_helper
+from azure.iot.device.iothub.models import Message
+from azure.iot.device.common.transport.mqtt.mqtt_provider import MQTTProvider
+from azure.iot.device.common.transport import constant
+
+logger = logging.getLogger(__name__)
+
+# TODO: standardize on {} instead of %s
+# TODO: standardize logs with {}({}) prefix
```
cartertinney: Potentially also consider to perhaps use {keywords} instead of just {}. It's not necessary, but it could be more readable. I don't do this too often, but it is nice when there's no surrounding context to draw from.

e.g.

```
"{name}({operation}) handling".format(name=self.name, operation=op.name)
```

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        self.pipeline_root = None
+
+    def run_op(self, op):
+        """
+        Run the given operation.  This is the public function that outside callers would call to run an
+        operation.  Derived classes should override the _run_op function to impliment stage-specific behavior.
+        """
+        try:
+            self._run_op(op)
+        except Exception as e:
```
cartertinney: >except Exception as e: [](start = 8, length = 22)

Use a bare `except` statement instead. This only catches things derived from Exception, not any raisable.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        """
+        Run the operations passed in *args in a serial manner, such that each operation waits for the
+        previous operation to be completed before running.  If an operation fails, the operations that
+        follow will not be run, but the operation passed using the finally_op will be run.  After
+        all operations and the finally_op (if provided) are run, then the function passed in the
+        callback parameter will be called, passing the last operation (in the event of success)
+        or the failing operation (in the event of failure).
+        """
+        worklist = deque(args)
+        finally_op = kwargs.get("finally_op", None)
```
cartertinney: >finally_op = kwargs.get("finally_op", None) [](start = 8, length = 43)

Why not just let this be the last arg provided in *args? I'm not understanding the point of it as a keyword

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        callback = kwargs.get("callback", None)
+
+        if len([x for x in kwargs.keys() if x not in ["finally_op", "callback"]]):
+            raise TypeError("finally_op and callback are the only allowed keyword args")
```
cartertinney: >raise TypeError("finally_op and callback are the only allowed keyword args") [](start = 12, length = 76)

I'm not sure about this check. I think the expected behavior of extraneous kwargs is just doing nothing with them rather than raising an error.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+    def run_ops_serial(self, *args, **kwargs):
+        """
+        Run the operations passed in *args in a serial manner, such that each operation waits for the
+        previous operation to be completed before running.  If an operation fails, the operations that
+        follow will not be run, but the operation passed using the finally_op will be run.  After
+        all operations and the finally_op (if provided) are run, then the function passed in the
+        callback parameter will be called, passing the last operation (in the event of success)
+        or the failing operation (in the event of failure).
+        """
+        worklist = deque(args)
```
cartertinney: >deque(args) [](start = 19, length = 11)

Interested - was there a particular reason you used a deque and it's .popleft method instead of using a list, and it's .pop method?

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+                    finally_op.error = last_op.error
+                callback(finally_op)
+
+            if finally_op:
+                finally_op.callback = finally_done
+                self.continue_op(finally_op)
+            else:
+                callback(last_op)
+
+        def single_op_callback(completed_op):
```
cartertinney: As we've discussed earlier, I think we should avoid the use of the term callback when defining the function that is used as a callback, since it only becomes a callback once provided to another object/layer/function/whatever to use.

i.e. the term "callback" should only refer to something already passed to the current layer by an upper layer, to call when required.

cartertinney: >single_op_callback(completed_op): [](start = 12, length = 33)

I also don't think "single_op" is a great name for this. I get why you've called it that, but given that it implicitly chains a series of singular operations, it isn't really a single op.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        callback = kwargs.get("callback", None)
+
+        if len([x for x in kwargs.keys() if x not in ["finally_op", "callback"]]):
+            raise TypeError("finally_op and callback are the only allowed keyword args")
+
+        def last_op_done(last_op):
+            def finally_done(finally_op):
+                if not finally_op.error:
+                    finally_op.error = last_op.error
+                callback(finally_op)
```
cartertinney: This (and the one below on line 78) should be in try/except blocks.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        operation.  Derived classes should override the _run_op function to impliment stage-specific behavior.
+        """
+        try:
+            self._run_op(op)
+        except Exception as e:
+            logger.error(msg="Error in _run_op call", exc_info=e)
+            op.error = e
+            self.complete_op(op)
+
+    def run_ops_serial(self, *args, **kwargs):
```
cartertinney: >run_ops_serial [](start = 8, length = 14)

I'm sure I'm missing something, but I'm not clear on why run_ops_serial is necessary, when we already have ops that chain, what with the .next attribute and the continue_op() method and all.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+
+    def complete_op(self, op):
+        """
+        Helper function to complete an operation by calling it's callback function thus
+        returning the result of the operation back up the pipeline.
+        """
+        logger.info("{}({}): completing".format(self.name, op.name))
+        op.callback(op)
+
+    def continue_with_translated_op(self, original_op, translated_op):
```
cartertinney: >continue_with_translated_op [](start = 8, length = 27)

Translated Operation not defined as a term here - I have no idea what this does from the docs.

Additionally, while I can infer what it does based on my knowledge, this concept of application specific term (enable C2D) vs. mqtt specific operation (subscribe), is elsewhere referred to as "conversion", which is a confusing mix of terms (if they are the same thing)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+
+    def _handle_pipeline_event(self, event):
+        if self.on_pipeline_event:
+            # TODO: try/catch
+            self.on_pipeline_event(event)
+        else:
+            logger.warning("incoming pipeline event with no handler.  dropping.")
+
+
+class IotHub_UseSkAuthProvider(PipelineStage):
```
cartertinney: >IotHub_UseSkAuthProvider [](start = 6, length = 24)

Here and everywhere - don't use underscores in class names.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+from azure.iot.device.common.transport import constant
+
+logger = logging.getLogger(__name__)
+
+# TODO: standardize on {} instead of %s
+# TODO: standardize logs with {}({}) prefix
+
+
+@six.add_metaclass(abc.ABCMeta)
+class PipelineStage(object):
```
cartertinney: Given that (some of) these are high level, protocol-agnostic concepts, does it make sense for them to all be in transport.mqtt? Shouldn't some of these be in transport rather than transport.mqtt?

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+                translated_op.name,
+                original_op.name,
+            )
+            original_op.error = translated_op.error
+            self.complete_op(original_op)
+
+        translated_op.callback = translated_op_complete
+        self.continue_op(translated_op)
+
+    # TODO: connected and disconnect migth become pipeline events.  maybe.
```
cartertinney: ># TODO: connected and disconnect migth become pipeline events.  maybe. [](start = 4, length = 70)

At some point we should really figure out some consistent story for connect and disconnect. Connect and Disconnect have been the bane of my existence at the transport layer, they're so inconsistent and hard to deal with. Plus I think we have a whole bunch of undefined edge cases surrounding them.

But I think that's a thing to deal with later.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+
+    def _run_op(self, op):
+        """
+        run the operation.  At the root, the only thing to do is to pass the operation
+        to the next stage.
+        """
+        logger.info("{}({}) handling".format(self.name, op.name))
+        self.continue_op(op)
+
+    def add_next_stage(self, new_next_stage):
```
cartertinney: >add_next_stage [](start = 8, length = 14)

terminology nit: couldn't we just call this 'append'? That term implies a lot of the detail of this, and you could then remove all references the verbose "new_next_stage". Plus "append" is the term used in the list primitive, so it's familiar

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+            # TODO: try/catch
+            self.on_pipeline_event(event)
+        else:
+            logger.warning("incoming pipeline event with no handler.  dropping.")
+
+
+class IotHub_UseSkAuthProvider(PipelineStage):
+    """
+    PipelineStage which handles a Shared Key Authentication Provider.  This is
+    likely a temporary stage if we're going to get rid of the two-stage
```
cartertinney: Maybe? We are definitely getting rid of two-stage client creation, but I'm not sure we're getting rid of auth providers. They may just become internal.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        # the connect is complete.
+        elif op.name is pipeline_ops.names.Iot_Connect:
+            if self.connected:
+                logger.info("transport is already conencted.  completing early.")
+                self.complete_op(op)
+            else:
+                self._do_connect(op)
+
+        # If we get a request to disconnect, we either complete the request immediately
+        # (if we're already disconencted) or we pass the disconnect request down.
```
cartertinney: I'm glad we finally have this. The transport crashing every time a disconnect was triggered was... not good.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        elif op.name is pipeline_ops.names.Iot_Disconnect:
+            if not self.connected:
+                logger.info("transport is already disconencted.  completing early.")
+                self.complete_op(op)
+            else:
+                self.continue_op(op)
```
cartertinney: Admittedly though, it is a bit semantically weird to be handling disconnects in the EnsureConnection stage. Of course, you have to handle disconnects in ensure connection, but I'd like it to clearly trigger a reconnect here. I'm guessing that's a side effect of disconnect, but it isn't clear from the scope of this class, and thus the semantics are odd to me.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+
+        # Any other operation that requires a connection can trigger a connection if i
+        # we're not connected.
+        elif op.needs_connection and not self.connected:
```
cartertinney: >elif op.needs_connection [](start = 8, length = 24)

Perhaps related to the above? Is this where reconnect happens? If so, it could be made clearer.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+    Note: this stage will likely be replaced by a more full-featured stage to handle
+    other "block while we're setting something up" operations, such as subscribing to
+    twin responses.  That is another example where we want to ensure some state and block
+    requests until that state is achieved.
+    """
+
+    def __init__(self):
+        PipelineStage.__init__(self)
+        self.connected = False
+        self.queue = deque()
```
cartertinney: >deque() [](start = 21, length = 7)

why use this instead of a Queue? Do we really need to generalize Stack/Queue behavior?

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+    def _do_connect(self, op=None):
+        """
+        Start connecting the transport in response to some operation (which may or may not be an Iot_Connect operation)
+        """
+        # first, we block the queue while we're connecting
+        self._block(op=op)
+
+        # If we're connecting as a side-effect of some other operation (that is not Iot_Connect), then we queueu
+        # that operation to run after the connection is complete.
+        if op.name is not pipeline_ops.names.Iot_Connect:
```
cartertinney: Perhaps do a type check instead?

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
```diff
@@ -0,0 +1,342 @@
...
+    A base class for data objects representing operations that travels down the pipeline.
+
+    Each PipelineOperation object represents a single asyncroneous operation that is performed
+    by the pipeline. The PipelineOperation objects travel through "stages" of the pipeline,
+    and each stage has the opportunity to act on each specific operation that it
+    receives.  If a stage does not handle a particular operation, it needs to pass it to the
+    next stage.  If the operation gets to the end of the pipeline without being handled
+    (completed), then it is treated as an error.
+
+    :ivar name: The name of the operation.  This, rather than the `isinstance` function, is used by stages to decide if they want to handle on a particular operation object.
```
cartertinney: >This, rather than the `isinstance` function [](start = 44, length = 43)

why

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+            logger.info("{}({}) queueing until connection complete".format(self.name, op.name))
+            self.queue.append(op)
+
+        # callback that gets called after we're connected.
```
cartertinney: >callback [](start = 10, length = 8)

as always, you know how i feel about things referred to as callbacks before they are provided to a consumer

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+
+        if op.name == pipeline_ops.names.IotHub_SetAuthProviderArgs:
+            # if we get auth provider args from above, we save some, use some to build topic names,
+            # and always pass it down because we know that the Mqtt_Provider stage will also want
+            # to receive these args.
+            self._set_topic_names(device_id=op.device_id, module_id=op.module_id)
+            client_id = op.device_id
+            if op.module_id:
+                client_id += "/" + op.module_id
+            username = op.hostname + "/" + client_id + "/" + "?api-version=2018-06-30"
```
cartertinney: >op.hostname + "/" + client_id + "/" + "?api-version=2018-06-30" [](start = 23, length = 63)

I've been thinking lately that string arithmetic should be used sparingly in favor of format strings.

```
username = "{hostname}/{client_id}/?api-version=2018-06-30".format(hostname=op.hostname, client_id=client_id)
```
is just more readable IMO.

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_topic_helper.py
```diff
@@ -0,0 +1,171 @@
+# --------------------------------------------------------------------------------------------
```
cartertinney: Mixed feelings about this module.

It definitely cleans up the code, so that's nice, but I'm not sure all of this should be maintained in the same place?

I'm good with it for now though.

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_topic_helper.py
```diff
@@ -0,0 +1,171 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cartertinney: lets drop it "helper" from the name though.

mqtt_topic is a good enough module name.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+            self.continue_op(op)
+
+    def _set_topic_names(self, device_id, module_id):
+        """
+        Build topic names based on the device_id and module_id passed.
+        """
+        self.telemetry_topic = mqtt_topic_helper.get_telemetry_topic_for_publish(
+            device_id, module_id
+        )
+        self.subscription_map = {
```
cartertinney: >subscription_map [](start = 13, length = 16)

I think this is a bad name. Map of what? Doesn't tell me that it has topics in it.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+
+class IotHub_MqttConverter(PipelineStage):
+    """
+    PipelineStage which converts other Iot and IotHub operations into Mqtt operations.  This stage also
+    converts mqtt pipeline events into Iot and IotHub pipeline events.
+    """
+
+    def __init__(self):
+        PipelineStage.__init__(self)
+        self.subscription_map = {}
```
cartertinney: >subscription_map [](start = 13, length = 16)

I think this is a bad name. Doesn't make clear that it contains topic strings.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+        elif op.name == pipeline_ops.names.Iot_SetSasToken:
+            # When we get a sas token from above, we just save it for later
+            logger.info("{}({}) got password".format(self.name, op.name))
+            self.sas_token = op.sas_token
+            self.complete_op(op)
+
+        elif op.name == pipeline_ops.names.Iot_Connect:
+            logger.info("{}({}) conneting".format(self.name, op.name))
+
+            def connect_callback():
```
cartertinney: >callback [](start = 24, length = 8)

on_connect or something instead. And everywhere there's a definition with "callback" in the name.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
```diff
@@ -0,0 +1,645 @@
...
+                self.provider.on_mqtt_connected = self.on_connected
+                self.on_connected()
+                self.complete_op(op)
+
+            self.provider.on_mqtt_connected = connect_callback
```
cartertinney: >on_mqtt_connected [](start = 26, length = 17)

You reset the provider handler every time you call connect? Why?

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+class Mqtt_Publish(PipelineOperation):
+    """
+    A PipelineOperation object which contains arguments used to publish a specific payload on a specific topic using the Mqtt protocol.
+
+    This operation is in the group of "Mqtt" operations becauase its attributes are very specific to the Mqtt protocol.
+    """
+
+    def __init__(self, topic=None, payload=None, qos=1, callback=None):
```
cartertinney: >qos=1 [](start = 49, length = 5)

Why let this be settable? Do we ever want to have a specific Publish with a unique QoS? Why not just hardcode the Publish QoS as 1?

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
```diff
@@ -0,0 +1,342 @@
...
+    pass
+
+
+# build a names object which contains names of all valid PipelineObject classes.
+# We use this to get some runtime code validation by comparing PipelineOperation.name values with attributes of this object (which are known-good PipelineOperation class names) instead of static strings (which might not be valid PipelineOperation class names)
+names = ValidPipelineOperations()
+o = None
+for o in locals():
+    if o.startswith("Iot_") or o.startswith("Mqtt_") or o.startswith("IotHub_"):
+        setattr(names, o, o)
```
cartertinney: IMO put this at the top of the file so it's more visible

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
```diff
@@ -0,0 +1,342 @@
...
+        Constructor to instantiate a `Mqtt_Subscribe` object.
+
+        :param callback Function: The function that gets called when this operation is complete or has failed.  The `callback` function must accept A PipelineOperation object which indicates the specific call which has completed or failed.
+        """
+        PipelineOperation.__init__(self, callback=callback)
+        self.topic = topic
+        self.qos = qos
+
+
+class Mqtt_Unsubscribe(PipelineOperation):
```
cartertinney: >_ [](start = 10, length = 1)

Here and everywhere, no underscores in class names

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
```diff
@@ -0,0 +1,342 @@
...
+    A PipelineOperation object which contains arguments used to unsubscribe from a specific Mqtt topic using the Mqtt protocol.
+
+    This operation is in the group of "Mqtt" operations becauase its attributes are very specific to the Mqtt protocol.
+    """
+
+    def __init__(self, topic=None, callback=None):
+        """
+        Constructor to instantiate a `Mqtt_Unsubscribe` object.
+
+        :param callback Function: The function that gets called when this operation is complete or has failed.  The `callback` function must accept A PipelineOperation object which indicates the specific call which has completed or failed.
```
cartertinney: >:param callback Functio [](start = 8, length = 23)

Here and everywhere the initializers are missing parameters in the docs.

Also nit, but these aren't constructors, they're initializers. The constructor is something else (that we never touch)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
```diff
@@ -0,0 +1,342 @@
...
+class Mqtt_SetConnectionArgs(PipelineOperation):
+    """
+    A PipelineOperation object which contains arguments used to connect to a server using the Mqtt protocol.
+
+    This operation is in the group of "Mqtt" operations becauase its attributes are very specific to the Mqtt protocol.
+    """
+
+    def __init__(
+        self,
+        client_id=None,
```
cartertinney: >client_id=None [](start = 8, length = 14)

Do we really want all these to be optional? Creates a lot of edge cases.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
```diff
@@ -0,0 +1,342 @@
...
+    This operation is in the group of "IotHub" operations becauase it is very specific to the IotHub client
+    """
+
+    def __init__(self, message=None, callback=None):
+        """
+        Constructor to instantiate a `IotHub_SendOutputEvent` object.
+
+        :param callback Function: The function that gets called when this operation is complete or has failed.  The `callback` function must accept A PipelineOperation object which indicates the specific call which has completed or failed.
+        """
+        PipelineOperation.__init__(self, callback=callback)
```
cartertinney: >PipelineOperation.__init__ [](start = 8, length = 26)

User super instead

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_events.py
```diff
@@ -0,0 +1,79 @@
...
+    have the opportunity to tie a PipelineEvent to a PipelineOperation object
+    if they are waiting for a response for that particular operation.
+    """
+
+    def __init__(self):
+        self.name = self.__class__.__name__
+        self.error = None
+
+
+class Mqtt_IncomingMessage(PipelineEvent):
```
cartertinney: >_ [](start = 10, length = 1)

Here and everywhere, no underscores

cartertinney: >Mqtt [](start = 6, length = 4)

I think we're doing MQTT by convention in names. ALL CAPS

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_events.py
```diff
@@ -0,0 +1,79 @@
...
+    A PipelineEvent object which represents an incoming Mqtt message on some Mqtt topic
+    """
+
+    def __init__(self, topic, payload):
+        PipelineEvent.__init__(self)
+        self.topic = topic
+        self.payload = payload
+
+
+class IotHub_C2dMessage(PipelineEvent):
```
cartertinney: >C2d [](start = 13, length = 3)

either c2d or C2D.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_events.py
```diff
@@ -0,0 +1,79 @@
...
+    pass
+
+
+# build a names object which contains names of all valid PipelineEvent classes.
+# We use this to get some runtime code validation by comparing PipelineEvent.name values with attributes of this object (which are known-good PipelineEvent class names) instead of static strings (which might not be valid PipelineEvent class names)
+names = ValidPipelineEvents()
+o = None
+for o in locals():
+    if o.startswith("Iot_") or o.startswith("Mqtt_") or o.startswith("IotHub_"):
+        setattr(names, o, o)
```
cartertinney: Put at top of file

azure-iot-device/tests/common/transport/mqtt/test_mqtt_transport.py
```diff
@@ -11,7 +11,7 @@
 from azure.iot.device.common.transport.mqtt.mqtt_transport import MQTTTransport
 from azure.iot.device.common.transport import constant
 from azure.iot.device.iothub.auth.authentication_provider_factory import from_connection_string
-from mock import MagicMock, patch
+from mock import MagicMock, patch, ANY
```
cartertinney: lets stop using mock altogether tbh. the mocker fixture from pytest-mock is syntactically smoother.

cartertinney:
> # license information.

Goes without saying that this whole file more or less needs to be scrapped in favor of layer specific unittests

---
Refers to: azure-iot-device/tests/common/transport/mqtt/test_mqtt_transport.py:4 in c3df954. [](commit_id = c3df9547bc9e6d162f0c912ace4a98c9fb88ce82, deletion_comment = False)

cartertinney:
Overall really happy with the way this is looking. Lots of suggested changes, but the overall design is pretty solid. I'm signing off now, since I'm not going to be reviewing again until I begin working again next week, but I'd still like to see those changes before a merge.

Not sure if you want to bother trying to merge this with the methods PR or not. Ideally I would have gotten it in before you were ready to push this, but alas.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: yes, absolutely.  splitting these files is one of my next steps


---
In reply to: [279665186](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279665186) [](ancestors = 279665186)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: Different concepts.  The .next attribute is the next stage in the pipeline.  continue_op() passes the operation down to the next stage.  Both of these affect how a single op goes through the pipeline.

run_ops_serial is about running multiple ops through the pipeline.  It takes an array of ops.  It passes the first one down, waits for it to travel through the pipeline and eventually complete.  After the first op completes, run_ops_serial passes the second op down the pipeline, and waits for it to complete, and so on.  We could do this with callbacks on each operation, but I found myself repeating that same pattern over and over so I made a helper.

---
In reply to: [279659606](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279659606) [](ancestors = 279659606)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: I've changed names and improved the doc.  I'm not sure if I'm ready to introduce a new subclass yet.  Conversion is the most likely reason to use this function, but I think we'll find other uses.

---
In reply to: [279661803](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279661803) [](ancestors = 279661803)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: No technical reason.  It was used in some code I was reading and I was curious to use it.  Any reason to change it to a list?

---
In reply to: [279654358](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279654358) [](ancestors = 279654358)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: It's like try/except/finally.  a failure in the chain will skip the following operations, but the finally op will always be run.

---
In reply to: [279651953](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279651953) [](ancestors = 279651953)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: I did this because, If you make a function (say def foo(bar=1):) and call it with an extra arg (say foo(bar=20, blah=1)), you get TypeError: foo() got an unexpected keyword argument 'blah'.  I wanted to do the same thing.  
I would have just used named args, but py27 wouldn't allow "def run_ops_serial(*args, finally_op=None, callback=None)" which is what I really wanted.

---
In reply to: [279652802](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279652802) [](ancestors = 279652802)

BertKleewein:
> # license information.

Agree 100%

---
In reply to: [487894714](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#issuecomment-487894714) [](ancestors = 487894714)

---
Refers to: azure-iot-device/tests/common/transport/mqtt/test_mqtt_transport.py:4 in c3df954. [](commit_id = c3df9547bc9e6d162f0c912ace4a98c9fb88ce82, deletion_comment = False)

azure-iot-device/tests/common/transport/mqtt/test_mqtt_transport.py
BertKleewein: OK

---
In reply to: [279687589](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279687589) [](ancestors = 279687589)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: We always want self.on_connected to be called (it informs other stages that we're connected), but, if we're handling an Iot_Connect op, we also want to complete the op.  The easiest way to do that is to change on_mqtt_connected while we're waiting for the connection to complete.  Maybe there's a better way, I'll have to think about this.

---
In reply to: [279684136](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279684136) [](ancestors = 279684136)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: I did it for a reason, but that reason no longer makes sense.  I"m going to sweep through and use isinstance.  

---
In reply to: [279677901](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279677901) [](ancestors = 279677901)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: Because I have a brand new hammer and this is very clearly a nail, or a screw, or, OK, I have no problem changing it.

---
In reply to: [279677490](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279677490) [](ancestors = 279677490)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
BertKleewein: old reasons.  keeping the name field (for logging), switching to isinstance

---
In reply to: [279678034](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279678034) [](ancestors = 279678034)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops_base.py
```diff
@@ -0,0 +1,156 @@
...
+        self.callback = callback
+        self.needs_connection = False
+        self.error = None
+
+
+class Connect(PipelineOperation):
+    """
+    A PipelineOperation object which tells the pipeline to connect to whatever service it needs to connect to.
+
+    This operation is in the group of "base" operations becauase connecting is a common operation that many clients might need to do.
```
olivakar: >becauase [](start = 56, length = 8)

spell

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops_base.py
```diff
@@ -0,0 +1,156 @@
...
+    This operation is in the group of "base" operations becauase disconnecting is a common operation that many clients might need to do.
+
+    Even though this is an "base" operation, it will most likely be handled by a more specific stage (such as an IotHub or Mqtt stage).
+    """
+
+    def __init__(self, feature_name=None, callback=None):
+        """
+        Constructor to instantiate a `Iot_EnableFeature` object.
+
+        :param callback Function: The function that gets called when this operation is complete or has failed.  The `callback`
```
olivakar: >:param callback Function [](start = 8, length = 24)

it should be :param Function callback

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: I think reconnecting is in a different not-yet-implemented retry stage.  

This stage is going to be heavily refactored into the stage we discussed at the whiteboard that ensures connection and also ensures subscriptions.   Ops are going to have an array of "demands" or 'requirements" that they need before they can run, and this stage will enforce those demands.  "connection" is one requirement for operations that require a connection, other requirements will all be subscriptions, at least to start., so an op could say that it requires a connection and a subscription to the twin topics before it can run.

---
In reply to: [279675524](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279675524) [](ancestors = 279675524)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_ops.py
BertKleewein: We should.  we carry this around and don't need to.

---
In reply to: [279685064](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279685064) [](ancestors = 279685064)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_base.py
```diff
@@ -0,0 +1,411 @@
...
+            self.complete_op(op)
+
+    def run_ops_serial(self, *args, **kwargs):
+        """
+        Run the operations passed in *args in a serial manner, such that each operation waits for the
+        previous operation to be completed before running.
+
+        In normal operation, the first operation will be passed down to the next stage to run.  That
+        stage will process it, pass it down, or whatever needs to be done to handle that operation.
+        After that operation travels all the way down the pipeline and the result is returned to the i
```
olivakar: > i [](start = 100, length = 2)

not needed

olivakar:
>         self._token_update_timer = None

can this be cancelled and instantiated again for the `_schedule_token_update` function instead of a temp timer variable ?

and if this is being done for any reason then may be i should do this too

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/base_renewable_token_authentication_provider.py:141 in ae2c1c7. [](commit_id = ae2c1c75b8ee29806f5949ca72511a59d1024278, deletion_comment = False)

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
```diff
@@ -5,138 +5,17 @@
...
-    TransportAction object used to send a method response back to the service.
-    """
-
-    def __init__(self, method_response, callback):
-        TransportAction.__init__(self, callback)
-        self.method_response = method_response
-
 
 class MQTTTransport(AbstractTransport):
     def __init__(self, auth_provider):
```
olivakar: >auth_provider [](start = 23, length = 13)

I think for DPS this needs to be a security client if we are going to use the same transport.
Since the stage is always using UseSkAuthProvider may we should use some other name for this stage and do another isinstance check

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
```diff
@@ -145,358 +24,67 @@ def __init__(self, auth_provider):
...
-            pass  # is there any other case
-
-    def _add_action_to_queue(self, event_data):
-        """
-        Queue an action for running later.  All actions that need to run while connected end up in
-        this queue, even if they're going to be run immediately.
-
-        This is called by the state machine as part of a state transition
+            else:
+                logger.warning("Dropping unkonwn pipeline event %s", event.name)
```
olivakar: >unkonwn [](start = 41, length = 7)

spell

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_base.py
```diff
@@ -0,0 +1,411 @@
...
+    def __init__(self):
+        self.name = self.__class__.__name__
+        self.next = None
+        self.previous = None
+        self.pipeline_root = None
+
+    def run_op(self, op):
+        """
+        Run the given operation.  This is the public function that outside callers would call to run an
+        operation.  Derived classes should override the _run_op function to impliment stage-specific behavior.
```
olivakar: >_run_op [](start = 56, length = 7)

add private like the `private _handle_pipeline_event`  on line 126

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages.py
BertKleewein: OK, I found the reason: [https://stackoverflow.com/questions/21639888/is-it-possible-to-convert-list-to-queue-in-python](https://stackoverflow.com/questions/21639888/is-it-possible-to-convert-list-to-queue-in-python)
popleft is O(n) for list of size n.  not a big deal here since our lists are small.
making a Queue from a list is slow (probably O(n)) compared to making a deque from a list (>300 times faster according to some comment in the stackoverflow discussion).
I think I'll just use Queue, but the efficiency tradeoff is still interesting.

---
In reply to: [279901078](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r279901078) [](ancestors = 279901078,279654358)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_base.py
```diff
@@ -0,0 +1,411 @@
...
+            logger.error(msg="Error in _handle_pipeline_event call", exc_info=e)
+            self.pipeline_root.unhandled_error_handler(e)
+
+    def _handle_pipeline_event(self, event):
+        """
+        Function which actually handles an event.  If a stage wants to handle events,
+        it should override this function.
+        """
+        if self.previous:
+            self.previous._handle_pipeline_event(event)
```
olivakar: >_handle_pipeline_event [](start = 26, length = 22)

should this be the public `handle_pipeline_event` ?

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_transport.py
BertKleewein: This transport is going to move into the iothub module since it's very iothub-specific.  DPS will need a different transport object (or whatever we want to call it) that will do the same thing with a DPS-specific pipeline and ops.

---
In reply to: [280164482](https://github.com/Azure/azure-iot-sdk-python-preview/pull/87#discussion_r280164482) [](ancestors = 280164482)

PR #89, opened 2019-05-07
pipeline doc and test

BertKleewein:
First stage of documentation and tests for pipeline objects.  I'm moving on to more tests, but I'd like to get this reviewed or merging into master.

### DIFF
--- a/azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
+++ b/azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
@@ -16,8 +16,7 @@ class PipelineOperation(object):
     next stage.  If the operation gets to the end of the pipeline without being handled
     (completed), then it is treated as an error.
 
-    :ivar name: The name of the operation.  This, rather than the `isinstance` function, is
-      used by stages to decide if they want to handle on a particular operation object.
+    :ivar name: The name of the operation.  This is used primarily for logging
     :type name: str
     :ivar callback: The callback that is called when the operation is completed, either
       successfully or with a failure.
@@ -102,6 +101,8 @@ class EnableFeature(PipelineOperation):
         """
         Initializer for EnableFeature objects.
 
+        :param str feature_name: Name of the feature that is being enabled.  The meaning of this
+          string is defined in the stage which handles this operation.
         :param Function callback: The function that gets called when this operation is complete or has
           failed.  The callback function must accept A PipelineOperation object which indicates
           the specific operation which has completed or failed.
@@ -128,6 +129,8 @@ class DisableFeature(PipelineOperation):
         """
         Initializer for DisableFeature objects.
 
+        :param str feature_name: Name of the feature that is being disabled.  The meaning of this
+          string is defined in the stage which handles this operation.
         :param Function callback: The function that gets called when this operation is complete or has
           failed.  The callback function must accept A PipelineOperation object which indicates
           the specific operation which has completed or failed.
@@ -151,6 +154,8 @@ class SetSasToken(PipelineOperation):
         """
         Initializer for SetSasToken objects.
 
+        :param str sas_token: The token string which will be used to authenticate with whatever
+          service this pipeline connects with.
         :param Function callback: The function that gets called when this operation is complete or has
           failed.  The callback function must accept A PipelineOperation object which indicates
           the specific operation which has completed or failed.
--- a/azure-iot-device/azure/iot/device/common/transport/pipeline_stages_base.py
+++ b/azure-iot-device/azure/iot/device/common/transport/pipeline_stages_base.py
@@ -20,11 +20,54 @@ class PipelineStage(object):
     Base class representing a stage in the processing pipeline.  Each stage is responsible for receiving
     PipelineOperation objects from the top, possibly processing them, and possibly passing them down.  It
     is also responsible for receiving PipelineEvent objects from the bottom, possibly processing them, and
-    possibly passing them up.  In both cases (PipelineOperation and PipelineEvent), the default behavior
-    should be to pass along an operation or event that you don't handle.
+    possibly passing them up.
+
+    Each PipelineStage in the pipeline, is expected to act on some well-defined set of PipelineOperation
+    types and/or some set of PipelineEvent types.  If any stage does not act on an operation or event, it
+    should pass it to the next stage (for operations) or the previous stage (for events).  In this way, the
+    pipeline implements the "chain of responsibility" design pattern (Gamma, et.al. "Design Patterns".
+    Addison Wesley. 1995), with each stage being responsible for implementing some "rule" or "policy" of the
+    pipeline, and each stage being ignorant of the stages that are before or after it in the pipeline.
+
+    Each stage in the pipeline should act on the smallest set of rules possible, thus making stages small
+    and easily testable.  Complex logic should be the exception and not the rule, and complex stages should
+    operate on the most generic type of operation possible, thus allowing us to re-use complex logic for
+    multiple cases.  The best way to do this is with "converter" stages that convert a specific operation to
+    a more general one and with other converter stages that convert general operations to more specific ones.
+
+    An example of a specific-to-generic stage is UseSkAuthProvider which takes a specific operation
+    (use an auth provider) and converts it into something more generic (here is your device_id, etc, and use
+    this SAS token when connecting).
+
+    An example of a generic-to-specific stage is IotHubMQTTConverter which converts IotHub operations
+    (such as SendTelemetry) to MQTT operations (such as Publish).
+
+    Each stage should also work in the broadest domain possible.  For example a generic stage (say
+    "EnsureConnection") that initiates a connection if any arbitrary operation needs a connection is more useful
+    than having some MQTT-specific code that re-connects to the MQTT broker if the user calls Publish and
+    there's no connection.
+
+    One way to think about stages is to look at every "block of functionality" in your code and ask yourself
+    "is this the one and only time I will need this code"?  If the answer is no, it migth be worthwhile to
+    implement that code in it's own stage in a very generic way.
+
+
+    :ivar name: The name of the stage.  This is used primarily for logging
+    :type name: str
+    :ivar next: The next stage in the pipeline.  Set to None if this is the last stage in the pipeline.
+    :type next: PipelineStage
+    :ivar previous: The previous stage in the pipeline.  Set to None if this is the first stage in the pipeline.
+    :type previous: PipelineStage
+    :ivar pipeline_root: The first stage (root) of the pipeline.  This is useful if a stage wants to
+      submit an operation to the pipeline starting at the root.  This type of behavior is uncommon but not
+      unexpected.
+    :type pipeline_root: PipelineStage
     """
 
     def __init__(self):
+        """
+        Initializer for PipelineStage objects.
+        """
         self.name = self.__class__.__name__
         self.next = None
         self.previous = None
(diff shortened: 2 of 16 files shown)


azure-iot-device/tests/common/transport/mqtt/test_pipeline_events_mqtt.py
```diff
@@ -0,0 +1,20 @@
...
+import pytest
+from azure.iot.device.common.transport.mqtt import pipeline_events_mqtt
+
+fake_topic = "__fake_topic__"
+fake_payload = "__fake_payload__"
+
+
+@pytest.mark.describe("IncomingMessage object")
+class TestIncomingMessage(object):
+    @pytest.mark.it("Sets arguments correctly")
```
cartertinney: >("Sets arguments correctly") [](start = 19, length = 28)

This isn't really a specific requirement, in a doc requirement style. What does "sets arguments correctly" mean?

I recommend breaking things like this up into logical groups that translate to actual requirements.

i.e.

1) "Sets polymorphic class name" (is that the right term for this pattern where the name is in the class? I can't remember)
2) "Sets topic and payload from message"

cartertinney:
> class PipelineEvent(object):

Should this perhaps be an AbstractBaseClass? Can this ever be instantiated by itself?

---
Refers to: azure-iot-device/azure/iot/device/common/transport/pipeline_events_base.py:8 in bceea68. [](commit_id = bceea68614ea45ea7c9af434ecf2609a5797aef5, deletion_comment = False)

azure-iot-device/tests/common/transport/test_pipeline_events_base.py
```diff
@@ -0,0 +1,15 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import pytest
+from azure.iot.device.common.transport import pipeline_events_base
+
+
+@pytest.mark.describe("PipelineEvent object")
+class TestPipelineEvent(object):
+    @pytest.mark.it("Sets arguments correctly")
```
cartertinney: >"Sets arguments correctly" [](start = 20, length = 26)

As in the other comment, this is too vague to be a requirement IMO (and everywhere it occurs)

azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
```diff
@@ -16,8 +16,7 @@ class PipelineOperation(object):
     next stage.  If the operation gets to the end of the pipeline without being handled
     (completed), then it is treated as an error.
 
-    :ivar name: The name of the operation.  This, rather than the `isinstance` function, is
-      used by stages to decide if they want to handle on a particular operation object.
+    :ivar name: The name of the operation.  This is used primarily for logging
```
cartertinney: >This is used primarily for logging [](start = 44, length = 34)

Why not just log with .__class__.__name__ then? Syntactical ease? I could understand that - it's pretty obscene to have to type that out every time we log (which is hopefully a lot)

azure-iot-device/tests/common/transport/test_pipeline_ops_base.py
```diff
@@ -0,0 +1,112 @@
...
+
+def assert_all_base_defaults(obj):
+    assert obj.name is obj.__class__.__name__
+    assert obj.needs_connection is False
+    assert obj.error is None
+
+
+@pytest.mark.describe("PipelineOperation object")
+class TestPipelineOperation(object):
+    @pytest.mark.it("Sets required and default arguments correctly")
```
cartertinney: Okay, so since this seems to be everywhere I won't keep bringing it up after this, but I'd like to see this as a more specific thing that reads like a requirement.

azure-iot-device/azure/iot/device/common/transport/pipeline_stages_base.py
```diff
@@ -34,7 +77,15 @@ def run_op(self, op):
         """
         Run the given operation.  This is the public function that outside callers would call to run an
         operation.  Derived classes should override the private _run_op function to implement
-        stage-specific behavior.
+        stage-specific behavior.  When run_op returns, that doesn't mean that the operation has executed
+        to completion.  Rather, it means that the pipeline has done something that will cause the
+        operation to eventually execute to completion.  That migth mean that something was sent over
```
cartertinney: >migth [](start = 61, length = 5)

typo

azure-iot-device/tests/common/transport/test_pipeline_stages_base.py
```diff
@@ -0,0 +1,544 @@
...
+    @pytest.mark.it("Sets required and default arguments correctly")
+    def test_initializer(self):
+        obj = PipelineStage()
+        assert_default_attributes(obj)
+
+
+@pytest.mark.describe("PipelineStage RunOp function")
+class TestPipelineStageRunOp(object):
+    @pytest.mark.it("calls _run_op")
+    def test_1(self, stage, op):
```
cartertinney: >test_1 [](start = 8, length = 6)

Even though we have the `pytest.mark.it` to tell us what the test actually is testing for, I think the testname should still be something that's legible, for a couple reasons:

1) numbering tests gets out of control quickly as test are added and removed.
2) I think there's value to the `pytest.mark.it` telling us what the requirement is, while the test name tells us what the test __does___ (since those are not exactly 1:1)

azure-iot-device/tests/common/transport/test_pipeline_stages_base.py
```diff
@@ -0,0 +1,544 @@
...
+    assert obj.name is obj.__class__.__name__
+    assert obj.next is None
+    assert obj.previous is None
+    assert obj.pipeline_root is None
+
+
+# because PipelineStage is abstract, we need something concrete
+class PipelineStage(pipeline_stages_base.PipelineStage):
+    def _run_op(self, op):
+        pass
```
cartertinney: I know that the EnsureConnection tests aren't in yet, but just as an idea, have you considered parametrizing some of these pipeline stage tests so that they will test all subclasses of PipelineStage as well?

After all, the basics (and complexity) of how a PipelineStage works is inherited, so a more specific PipelineStage should pass all the tests the generic PipelineStage does, right?

Instead of hacking a PipelineStage object to instantiate even though it's an AbstractBaseClass, why not ONLY test the child classes? After all, the parent class is NEVER instantiated in practice as it is in this test file, it's only actually done through the children. By parametrizing your objects under test among the various PipelineStage child classes, you can use one set of tests to cover all PipelineStage children.

This accomplishes 2 things:

1) It tests the classes as they are actually used in practice.
2) It gives us assurance that these tests pass for all child classes. 

After all, testing the parent class only tells us that those rules are valid for the children by convention - we take for granted the children don't muck it up somehow.

I hope that all makes sense.

azure-iot-device/tests/iothub/transport/test_pipeline_events_iothub.py
```diff
@@ -0,0 +1,29 @@
...
+# license information.
+# --------------------------------------------------------------------------
+import pytest
+from azure.iot.device.iothub.transport import pipeline_events_iothub
+
+fake_message = "__fake_messsage__"
+fake_input_name = "__fake_input_name__"
+
+
+@pytest.mark.describe("C2DMessage object")
```
cartertinney: >C2DMessage object [](start = 23, length = 17)

This is probably going to be confusing since the event is called C2DMessage, which makes it sound like a more specific version of the Message class.

Perhaps events should have "Event" in the title, i.e. C2DMessageEvent to avoid confusion?

azure-iot-device/tests/iothub/transport/test_pipeline_ops_iothub.py
```diff
@@ -0,0 +1,106 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
cartertinney: Comments from previous files also apply to this one

cartertinney:
I've signed off, but I've flagged the changes I still would like to see first (if possible - some may not yet be feasible, perhaps I'm missing something). If they can't be, we should discuss some of these approaches to testing at a later date.

azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
BertKleewein: Syntactical ease, and it also gives us the option to be more explicit about names in the future.  I'm thinking about cases where we have 2 ops with the same name discriminated by namespace.  This allows us to add the namespace later if we want.

---
In reply to: [282122987](https://github.com/Azure/azure-iot-sdk-python-preview/pull/89#discussion_r282122987) [](ancestors = 282122987)

BertKleewein:
> class PipelineEvent(object):

Maybe, but:
1. it's handy for testing.  I like your idea on test refactor to test base functionality in derived classes, but I don't want to bite that off right now.
2. can you have an ABC where nothing is actually abstract?  There's nothing that _must_ be implemented in a derived class, so I'm not sure if that makes this actually abstract or not.

---
In reply to: [490534307](https://github.com/Azure/azure-iot-sdk-python-preview/pull/89#issuecomment-490534307) [](ancestors = 490534307)

---
Refers to: azure-iot-device/azure/iot/device/common/transport/pipeline_events_base.py:8 in bceea68. [](commit_id = bceea68614ea45ea7c9af434ecf2609a5797aef5, deletion_comment = False)

azure-iot-device/tests/common/transport/test_pipeline_stages_base.py
BertKleewein: That does make sense.  I can see your argument, but I'm not sure if I agree.  I think that kind of testing _could_ have value, and it could save us some copy/paste in our tests, but I'm not sure how much value it would actually add.  

I think you're asking us to really think hard about what the word "unit" in "unit tests" means.  We usually define "unit" as a single piece of code (an object or a function), and we mock everything outside of that unit.  So, what is the "unit" we are testing here?  

As a thought exercise, think about using composition instead of inheritance for these objects and how we would test that?  In other words, if an Mqtt pipeline stage _contained_ a base pipeline object to do the base work (a "has-a" relationship, as in "an MQTT stage has a base object that it can use to do work"), it's _almost_ exactly the same as our case where an mqtt pipeline stage inherits the behavior of the the base stage (an "is-a" relationship, as in "an MQTT stage is-a base stage").  In the "has-a" example, we would mock the base object and only test that the MQTT object called into the base object.  We wouldn't actually test the base object.  Why is this different?

Another way to think about it is to think about how waterproof our assurances can be is to think about how we test the derived classes.  

First of all, if a derived object never calls run_ops_serial on itself, then we don't get anything extra by verifying that run_ops_serial works correctly in the derived class.  Sure, we've verified that this function is available in the derived class and that it behaves correctly, but it's never called, so who cares?   

However, if a derived class uses run_ops_serial, it's using it to do something specific, and that specific thing is part of the derived class' requirements.  For example, if a derived class uses run_ops_serial to run 3 ops in a row, our derived class tests had better verify the behavior in running those 3 ops.  (e.g. the first one runs, if the first one succeeds, the second one runs, etc.). The requirements on that derived class should specify that the 3 ops run in a row, and it should specify how to handle errors and whatnot, and the derived class tests should validate that.  The fact that the derived class uses run_ops_serial is an implementation detail.  

Testing the full range of run_ops_serial does not give us any advantage here either because we only really care about the subset of functionality that the derived class actually uses.  This is the same as the "derived class never calls run_ops_serial" case, except it's more nuanced.  If we verify that there's a run_ops_serial function in the derived class that works correctly with some set of parameters, but the derived class never actually calls it with those parameters, what value do we get in testing the derived class with those parameters?

To draw a visual analogy, say we're testing a button on some screen and that button has some text on it.  The base implementation for that button also supports buttons with pictures on them.  Do we need to test how this button will behave with a picture on it, even if this button will never have a picture on it?  Sure, we could, but it's a slippery slope.  

---
In reply to: [282134493](https://github.com/Azure/azure-iot-sdk-python-preview/pull/89#discussion_r282134493) [](ancestors = 282134493)

PR #90, opened 2019-05-10
CR feedback from pipeline doc and test 

BertKleewein:


azure-iot-device/tests/common/transport/test_pipeline_stages_base.py
```diff
@@ -6,6 +6,7 @@
 import logging
 import pytest
 import functools
+from unittest import mock
```
cartertinney: Please use the mocker fixture of pytest instead

azure-iot-device/tests/common/transport/test_pipeline_ops_base.py
```diff
@@ -18,95 +18,109 @@ def assert_all_base_defaults(obj):
 
 @pytest.mark.describe("PipelineOperation object")
 class TestPipelineOperation(object):
-    @pytest.mark.it("Sets required and default arguments correctly")
+    @pytest.mark.it("Sets name attribute on instantiation")
```
cartertinney: >@pytest.mark.it("Sets name attribute on instantiation") [](start = 4, length = 55)

As discussed, these multiple `pytest.mark.it`s will not actually result in multiple printed outputs during test.

I don't care too much if this is changed now or later, but we need to formalize (further) how we write requirements (what's in the wiki is kind of just a starting point and up for debate still), and adjust to follow a coherent strategy. Perhaps that will mean splitting these into multiple tests, rewriting the requirement, or asking for multiple mark support from pytest-testdox.

Up to you if you want to change this to a single requirement for the time being or not. I think in the short term that would result in better output, but also we aren't really using the test output as a requirements document (yet) since it isn't up to snuff across the board yet.

cartertinney:
>         pass

As I said in the last PR #89, on this same line, I don't think we should be testing AbstractClasses. You had a long follow up that raised some other interesting questions as well, but we should continue that discussion offline, and with others, and then formalize the results for a wiki article, and then implement the changes (if any) in a separate PR.

I'm making this comment purely for historical documentation reasons.

---
Refers to: azure-iot-device/tests/common/transport/test_pipeline_stages_base.py:27 in add0a4d. [](commit_id = add0a4d2b20fe25c151ef3ee96fc02c12762cde0, deletion_comment = False)

azure-iot-device/tests/common/transport/test_pipeline_ops_base.py
BertKleewein: waiting until we decide what we want to do.

---
In reply to: [283494972](https://github.com/Azure/azure-iot-sdk-python-preview/pull/90#discussion_r283494972) [](ancestors = 283494972)

PR #92, opened 2019-05-17
[BUGFIX] Timing/Threading bug in MQTT Provider

cartertinney:
* Fixed a bug where depending on the timing of operations in different threads, tracking of request/responses for Provider operations could end up in an unexpected state and cause an operation to never complete.

* Refactored operations management into a helper class in the Provider


BertKleewein:
Making this lock-free was a bold choice.  Without a lock, you have to be able tp prove that any possible interleaving of operations is valid and that's not easy.  

Here's an interleaving that breaks:
Thread A is calling publish (calls establish_operation)
Thread B is where the PUBACK comes back  (calls complete_operation)

Thread A:
calls PUBLISH

Thread B:
received PUBACK
calls complete_operation
checks the pending queue, this returns False since thread A hasn't added it yet.
<-- thread switch -->

Thread A:
PUBLISH operation returns MID
calls establish_operation
adds MID to pending queue
checks unknown queue, this returns False,
function is done

Thread B:
(continuing where it left off, right after B decided this wasn't in the pending queue)
adds to unknown queue
function is done

azure-iot-device/azure/iot/device/common/transport/mqtt/mqtt_provider.py
```diff
@@ -217,14 +209,48 @@ def publish(self, topic, payload, qos=1, callback=None):
...
+    def establish_operation(self, mid, callback=None):
+        """Establish a pending operation identified by MID, and store its completion callback.
 
-    def _set_operation_callback(self, mid, callback):
-        if mid in self._unknown_operation_responses:
-            # If response already came back, trigger the callback
+        If the operation has already been completed, the callback will be triggered.
+        """
+        # Manually acquire the lock instead of using a handler so it can be released
+        # at different times in the logical branch
```
BertKleewein: This is OK, but what happens if there's an exception after you acquire and before you release?  It's not terribly likely, but there's no way to recover if it does happen.  I would recommend using a try/finally or a context manager and add a variable to pull the callback out.

like:
do_callback=False
with(lock):
  if blah:
    do_callback = True
if do_callback:
  callback()

I don't like having to add an the extra variable, but having a guarantee that the lock gets released is golden.

BertKleewein: Plus, my fragile little mind doesn't like having to trace all the logic paths just to make sure it always gets released.  Better to let the language do the work. You know, context managers good, blah blah blah.  

---
In reply to: [287356662](https://github.com/Azure/azure-iot-sdk-python-preview/pull/92#discussion_r287356662) [](ancestors = 287356662)

cartertinney: This is a good point - I didn't think about exceptions, since this code clearly cannot raise exceptions between the acquire and release of the lock, and it would be fairly trivial to prove it.

But that could always change, and thus the best practice is definitely to use the context manager. 

---
In reply to: [287356662](https://github.com/Azure/azure-iot-sdk-python-preview/pull/92#discussion_r287356662) [](ancestors = 287356662)

PR #94, opened 2019-05-19
more pipeline tests

BertKleewein:
This is the last round of pipeline tests.  I found a few bugs along the way.  Sorry for the gigantic commit.  There was lots of code to cover.


azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_mqtt.py
```diff
@@ -58,7 +58,27 @@ def on_connected():
                 self.complete_op(op)
 
             self.provider.on_mqtt_connected = on_connected
-            self.provider.connect(self.sas_token)
+            try:
+                self.provider.connect(self.sas_token)
+            except BaseException:
```
olivakar: >BaseException [](start = 19, length = 13)

is this exception only possible during connect reconnect and disconnect ? and NOT during publish or any thing else ?
if so may be we need to update docs in the provider functions for connect , disconnect and reconnect saying it may raise exceptions .

azure-iot-device/azure/iot/device/iothub/transport/mqtt/pipeline_stages_iothub_mqtt.py
```diff
@@ -137,7 +140,8 @@ def _handle_pipeline_event(self, event):
                 self._handle_pipeline_event(pipeline_events_iothub.MethodRequest(method_received))
 
             else:
-                logger.warning("Warning: dropping message with topic {}".format(topic))
+                logger.info("Warning: unknown topic {} passing up to next handler".format(topic))
```
olivakar: >info [](start = 23, length = 4)

we can make this logger.warn if this needs to be warning

azure-iot-device/tests/common/pipeline_test_fixtures.py
```diff
@@ -0,0 +1,25 @@
...
+    return mocker.Mock()
+
+
+@pytest.fixture
+def fake_error():
+    return pipeline_test.get_fake_error()
+
+
+@pytest.fixture
+def event():
```
olivakar: >event [](start = 4, length = 5)

can this be named to mock event or test event ?

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_mqtt.py
cartertinney: Excepting on `BaseException` seems dangerous since it catches all kinds of stuff that you probably don't want (KeyboardInterrupt).

Also, in looking this up it turns out all the places we've been using bare `except` statements are implicitly doing this... so that's also probably a problem in those places. Perhaps instead of addressing this now, we can address it with a future PR that improves the exception handling story, and for now just use a bare `except` for consistency (which does the same thing as `except BaseException`)

---
In reply to: [285815533](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r285815533) [](ancestors = 285815533)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_mqtt.py
```diff
@@ -58,7 +58,27 @@ def on_connected():
...
+                self.provider.on_mqtt_connected = self.on_connected
```
cartertinney: Also I'm just generally rather confused by what's going on this whole area. I'd appreciate an inline comment explaining.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_mqtt.py
```diff
@@ -58,7 +58,27 @@ def on_connected():
...
+                raise
+
+        elif isinstance(op, pipeline_ops_base.Reconnect):
+            logger.info("{}({}): reconneting".format(self.name, op.name))
```
cartertinney: typo: reconnecting

azure-iot-device/azure/iot/device/iothub/transport/pipeline_stages_iothub.py
```diff
@@ -37,7 +37,7 @@ def _run_op(self, op):
             self.run_ops_serial(
                 pipeline_ops_iothub.SetAuthProviderArgs(
                     device_id=auth_provider.device_id,
-                    module_id=auth_provider.module_id,
+                    module_id=getattr(auth_provider, "module_id", None),
```
cartertinney: >module_id=getattr(auth_provider, "module_id", None), [](start = 20, length = 52)

This is odd, why is this default to `None` necessary? The value for `module_id` alerady defaults to `None` in the auth provider class, why do we need it here as well?

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_topic.py
```diff
@@ -64,24 +64,24 @@ def get_method_topic_for_subscribe():
     return "$iothub/methods/POST/#"
 
 
-def is_c2d_topic(topic):
+def is_c2d_topic(topic, device_id):
```
cartertinney: I think this is a positive change, but is the function name still relevant? This now not just tells us if the topic is a c2d_topic, but also if it matches our device.

azure-iot-device/tests/common/pipeline_test_fixtures.py
cartertinney: I disagree, it isn't a mock event. Nor is it really a "test" event. It just is a pipeline event.

---
In reply to: [285817618](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r285817618) [](ancestors = 285817618)

azure-iot-device/tests/common/pipeline_test.py
```diff
@@ -0,0 +1,117 @@
...
+    assert obj.name is obj.__class__.__name__
+    assert obj.next is None
+    assert obj.previous is None
+    assert obj.pipeline_root is None
+
+
+# because PipelineStage is abstract, we need something concrete
+class ConcretePipelineStage(pipeline_stages_base.PipelineStage):
+    def _run_op(self, op):
+        self.continue_op(op)
```
cartertinney: Are we still doing this?

azure-iot-device/tests/common/pipeline_test.py
```diff
@@ -0,0 +1,117 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
```
cartertinney: This file, if I understand it, contains helper functions for use in various pipeline tests. So I have two follow up questions:

1) Could we perhaps name it pipeline_test_helpers.py instead to make it clearer?
2) Why is this located in the root of the tests.common directory? Shouldn't this be in tests.common.pipeline?

azure-iot-device/tests/common/pipeline_test_fixtures.py
```diff
@@ -0,0 +1,25 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
```
cartertinney: Again, I'm not really sure why this is in tests.common rather than tests.common.pipeline, given that it's specific for pipeline tests.

Is it because you expose it through conftest.py? If so, did you know that you can have a conftest in every directory/subdirectory that applies to everything from that level down?

i.e. it seems like this pipeline stuff in tests.common should all be in tests.common.pipeline, including a new conftest.py

azure-iot-device/tests/common/transport/mqtt/test_pipeline_stages_mqtt.py
```diff
@@ -0,0 +1,389 @@
...
+    @pytest.mark.it("Sets name attribute on instantiation")
+    @pytest.mark.it("Sets next attribute to None on instantiation")
+    @pytest.mark.it("Sets previous attribute to None on instantiation")
+    @pytest.mark.it("Sets pipeline_root attribute to None on instantiation")
+    def test_initializer(self):
+        obj = pipeline_stages_mqtt.Provider()
+        assert_default_stage_attributes(obj)
+
+
+@pytest.mark.describe("MQTT Provider pipeline stage _runOp function with unhandled operations")
```
cartertinney: >function [](start = 59, length = 8)

method?

azure-iot-device/tests/common/transport/mqtt/test_pipeline_stages_mqtt.py
```diff
@@ -0,0 +1,389 @@
...
+    def test_restores_handler_on_failure(
+        self, params, stage, create_provider, op, mocker, fake_error, provider_function_fails
+    ):
+        handler_before = getattr(stage.provider, params["provider_handler"])
+        stage.run_op(op)
+        handler_after = getattr(stage.provider, params["provider_handler"])
+        assert handler_before == handler_after
+
+
+@pytest.mark.describe("MQTT Provider pipeline stage events from transport library")
```
cartertinney: >transport library [](start = 64, length = 17)

transport library?

azure-iot-device/tests/iothub/pipeline_test.py
```diff
@@ -0,0 +1,20 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
```
cartertinney: Same comments as in common,

1) this should be pipeline_test_helpers, if it has to exist at all (consider maybe putting it in the conftest.py?)
2) this and all other pipeline specific config/fixtures/etc. should be scoped to only apply to places where we are testing those things

cartertinney:
I'm admittedly a little confused about the overall testing picture for the pipeline stuff since it's spread across multiple PRs + there's the incoming reorganization of files in PR  #95 still coming.

I'll definitely want to revisit some of this stuff in more depth in the coming week once it's all merged in.

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_mqtt.py
BertKleewein: It's not that the functions _can_ raise exceptions.  _Any_ function can raise an exception.  This is here because we need to something _special_ if an exception is raised.  I added this comment to the code.  Hopefully it explains it well.

A note on exceptions handling in Connect, Disconnct, and Reconnet:
 
All calls into self.provider can raise an exception, and this is OK.
The exception handler in PipelineStage.run_op() will catch these errors
and propagate them to the caller.  This is an intentional design of the
pipeline, that stages, etc, don't need to worry about catching exceptions
except for special cases.

The code right below this comment is This is a special case.  In addition
to this "normal" exception handling, we add another exception handler
into this class' Connect, Reconnect, and Disconnect code.  We need to
do this because provider.on_mqtt_connected and provider.on_mqtt_disconnected
are both _handler_ functions instead of _callbacks_.

Because they're handlers instead of callbacks, we need to change the
handlers while the connection is established.  We do this so we can
know when the provider is connected so we can move on to the next step.
Once the connection is established, we change the handler back to its
old value before finishing.

The exception handling below is to reset the handler back to its original
value in the case where provider.connect raises an exception.  Again,
this extra exception handling is only necessary in the Connect, Disconnect,
and Reconnect case because they're the only cases that use handlers instead
of callbacks.


---
In reply to: [287157069](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287157069) [](ancestors = 287157069,285815533)

azure-iot-device/azure/iot/device/common/transport/mqtt/pipeline_stages_mqtt.py
BertKleewein: Not sure what you mean by "this whole area".  Hopefully my comment above explains it for you.  If not, please be more explicit about what's missing or confusing you.

---
In reply to: [287157524](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287157524) [](ancestors = 287157524)

BertKleewein:
agreed.  lets talk next week.  

---
In reply to: [495817460](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#issuecomment-495817460) [](ancestors = 495817460)

BertKleewein:
regarding try/except:  let's come up with a consistent story here.  At one point, you commented that "catch Exception" around external code wasn't catching enough, and I can see that argument.  But now it looks like "catch BaseException" is too much, so let's figure out what is enough when catching generic errors.

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_topic.py
BertKleewein: I would like to revisit this whole module.  I'm going to add this as a topic when we talk next week.

---
In reply to: [287540129](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287540129) [](ancestors = 287540129)

BertKleewein: Er, I'd like to revisit iothub_mqtt_topic.py

---
In reply to: [287568462](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287568462) [](ancestors = 287568462,287540129)

azure-iot-device/azure/iot/device/iothub/transport/pipeline_stages_iothub.py
BertKleewein: defensive programming.  I didn't consult the providers and just considered the idea that module_id isn't required.  

---
In reply to: [287536915](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287536915) [](ancestors = 287536915)

azure-iot-device/tests/common/pipeline_test.py
BertKleewein: I think you're talking about  not testing abstract classes.  I want to cover that under a different PR.  

---
In reply to: [287540616](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287540616) [](ancestors = 287540616)

azure-iot-device/tests/common/pipeline_test.py
BertKleewein: Will revisit in a different PR.  I agree that this is awkward.  

---
In reply to: [287540855](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287540855) [](ancestors = 287540855)

azure-iot-device/tests/common/pipeline_test_fixtures.py
BertKleewein: Agree that it's awkward.  Will revisit in a later PR.

---
In reply to: [287541060](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#discussion_r287541060) [](ancestors = 287541060)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -35,7 +35,7 @@ def _run_op(self, op):
             self.run_ops_serial(
                 pipeline_ops_iothub.SetAuthProviderArgs(
                     device_id=auth_provider.device_id,
-                    module_id=auth_provider.module_id,
+                    module_id=getattr(auth_provider, "module_id", None),
```
cartertinney: Making a new comment since the old one got squashed - I understand defaulting to None for defensive purposes, but this is only important if the auth_provider might not have a module_id attribute, which it always does. The "optional" part of the module_id attribute just means it may or may not be set to None. But the point is, the attribute is always there.

Using getattr is only needed if there's some kind of weird class hierarchy where some auth_providers may or may not have the attribute (but here, all auth providers have the attribute because it comes from the Abstract Class that serves as the super to all auth providers)

cartertinney:
Yeah I think I may have been wrong about what I said in the past. I think addressing this should be done all at once in a separate PR though. 

---
In reply to: [495931350](https://github.com/Azure/azure-iot-sdk-python-preview/pull/94#issuecomment-495931350) [](ancestors = 495931350)

PR #96, opened 2019-05-23
stubs for twin functionality along with some impl.

BertKleewein:
This has function stubs, object definitions, and some minor implementation for twin.  I'm looking for feedback on the client API surface.  I also have some open issues that I'd like help resolving, all marked with `# OPEN ISSUE` text.

Some background:
* twins have 3 operations:
1. get a twin.  This is a query that is sent from the device/module to iothub.  iothub returns a complete twin object with desired and, I think,  reported properties.  It uses the request/response pattern, where a request id  (rid) is included in the topic we send, and that same request id is included in the response.
2. patch reported properties.  This is a push from the device/module to iothub where the device/module updates a partial or full set of the reported properties.  This uses the same request/response pattern.
3. wait for desired property patch.  This is like waiting for c2d or method calls.  There is no request (except for subscribing).  The iothub service sends a desired property patch to the client whenever the desired properties are updated on the service.

To get the twin, the client publishes to `$iothub/twin/GET/?$rid=<guid>` with an empty request body.  The service responds by sending a message to `$iothub/twin/res/<status>/?$rid=<guid>` with the JSON for the twin in the body.

To patch the reported properties, the client publishes to `$iothub/twin/patch/properties/reported/?rid=<guid>` with the JSON for the desired properties in the body.  The service responds on `$iothub/twin/res/<status>/?$rid=<guid>` with an empty body.

To wait for desired properties patches, the client subscribes to `$iothub/twin/PATCH/properties/desired/#` and any messages on this topic contains the desired properties in the body of the message

Pipeline flow (for GET):
1. `PipelineAdapter/MQTTTransport` creates a `GetTwin` operation and runs it on the pipeline.
2. the `HandleTwinOperations` stage receives the `GetTwin` operation.  It converts it into a `IotRequestAndResponse` object and sends it down.
3. the `CoordinateRequestAndResponse` converts the `IotRequestAndResponse` operation into a `SendIotRequest` operation and sends it down.
4. the `IotHubMQTTConverter` stage converts the `SendIotRequest` op into a `Publish` op and sends it down for publishing.

5. later on, the response arrives back from the service as an MQTT message inside a `IncomingMessage` event.
6. The `IotHubMQTTConverter` stage converts the message into a `IotResponseEvent` event and passes it up
7. The `CoordinateRequestAndResponse` takes the `IotResponseEvent` event, matches the RID back to the `IotRequestAndResponse` operation from step 2 and completes that operation.
8. The callback from the `IotResponseEvent`, which is inside the `HandleTwinOperations stage`, completes the `GetTwin` operation and the result is returned back to the caller inside `PipelineAdapter/MQTTTransport`

The path was done this way to remove as much of the MQTT-ness as possible, and make it so the twin code is completely transport-agnostic, except for the code inside the IotHubMQTTConverter, which is the minimal set of code to convert generic operations into mqtt operations.


azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
```diff
@@ -162,3 +162,81 @@ def __init__(self, sas_token, callback=None):
...
+
+    Even though this is an base operation, it will most likely be generated and also handled by more specifics stages
+    (such as IotHub or Mqtt stages).
+
+    The caller is required to parse the request_body that is returned into a more appropraite form (such as running a
+    JSON parser to convert a string into an object.)
+
+    :ivar status_code: The status code returned by the response.  Any value under 300 is considered success.
+    :type status_code: int
+    :ivar response_body: The body of the response.
```
olivakar: >response_body [](start = 10, length = 13)

this is bytes ? at least for MQTT it is.
or is thus converted to string ?

azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
```diff
@@ -162,3 +162,81 @@ def __init__(self, sas_token, callback=None):
...
+        """
+        Initializer for IotRequestAndResponse objects
+
+        :param str request_type: The type of request.  This is a string which is used by transport-specific stages to
+          generate the actual request.  For example, if request_type is "twin", then the iothub_mqtt stage will convert
+          the request into an MQTT publish with topic that begins with $iothub/twin
+        :param str verb: The verb for the request, such as "POST", "GET", etc.
+        :param str resource: The resource that the verb is acting on.  For twin request with verb "GET", this is most
+          likely the string "/" which retrieves the entire twin resource
+        :param request_body: The body of the request.  This is a required field, and a single space can be used to denote
```
olivakar: >request_body [](start = 15, length = 12)

mismatch in param type above and below

azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
```diff
@@ -162,3 +162,81 @@ def __init__(self, sas_token, callback=None):
...
+class IotRequestAndResponse(PipelineOperation):
+    """
+    A PipelineOperation object which wraps the common operation of sending a request to iothub with a request_id ($rid)
+    value and waiting for a response with the same $rid value.  This convention is used by both Twin and Provisioning
+    features.
+
+    Even though this is an base operation, it will most likely be generated and also handled by more specifics stages
+    (such as IotHub or Mqtt stages).
+
+    The caller is required to parse the request_body that is returned into a more appropraite form (such as running a
```
olivakar: >returned [](start = 61, length = 8)

why is request returned ?

olivakar: >appropraite [](start = 82, length = 11)

spell

azure-iot-device/azure/iot/device/common/transport/pipeline_ops_base.py
```diff
@@ -162,3 +162,81 @@ def __init__(self, sas_token, callback=None):
         """
         super(SetSasToken, self).__init__(callback=callback)
         self.sas_token = sas_token
+
+
+class IotRequestAndResponse(PipelineOperation):
```
olivakar: >IotRequestAndResponse [](start = 6, length = 21)

the name of this does not feel like an operation. It feels more like a model or object name.
Not sure what to call it but it is lacking some verb..

azure-iot-device/azure/iot/device/common/transport/pipeline_stages_base.py
```diff
@@ -611,3 +613,62 @@ def on_connected(self):
...
+            # operations which have not received responses yet.  If the operation is found,
+            # complete it.
+
+            if event.request_id in self.pending_responses:
+                op = self.pending_responses[event.request_id]
+                del (self.pending_responses[event.request_id])
+                op.status_code = event.status_code
+                op.response_body = event.response_body
+            else:
+                # OPEN ISSUE: what to do if we don't have a matching $rid?  Drop it, error, i
```
olivakar: >OPEN ISSUE [](start = 18, length = 10)

is this response important ? like will anyone do anything with this response ? In that case it may be passed up.
This must be coming to an rid that was sent in the past. 
I would say that just log an error and log out the response , in a way drop it.  
Error may be a bit overboard.

azure-iot-device/azure/iot/device/iothub/transport/pipeline_stages_iothub.py
```diff
@@ -47,3 +48,42 @@ def _run_op(self, op):
...
             self.continue_op(op)
+
+
+class HandleTwinOperations(PipelineStage):
+    """
+    PipelineStage which handles twin operations. In particular, it converts twin GET and PATCH
+    operations into IotRequestAndResponse operations.  This is done at the IotHub level because
+    there is nothing transport-specific about this code.  The transport-specific implementation
+    for twin requests and responses is handled inside IotHubMQTTConverter, when it converts
+    the SendIotRequest toa transport-specific send operation and when it converts the
```
olivakar: >toa [](start = 23, length = 3)

to a

olivakar:
> class MethodRequest(PipelineEvent):

This is unrelated to this review. But i see every word ending with event except this one.

---
Refers to: azure-iot-device/azure/iot/device/iothub/transport/pipeline_events_iothub.py:43 in 623fc25. [](commit_id = 623fc252c2ca3951adf0fb6c2db8bf03ac430684, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_topic.py
```diff
@@ -252,3 +252,38 @@ def encode_properties(message_to_send, topic):
         topic += user_properties_encoded
 
     return topic
+
+
+def get_twin_response_topic_for_subscribe():
+    return "$iothub/twin/res/#"
```
olivakar: >$iothub/twin [](start = 12, length = 12)

can this be a topic base for twin like we have a topic base for iothub telemetry operations ?

azure-iot-device/azure/iot/device/iothub/transport/mqtt/pipeline_stages_iothub_mqtt.py
```diff
@@ -134,7 +151,20 @@ def _handle_pipeline_event(self, event):
                 method_received = MethodRequest(
                     request_id=rid, name=method_name, payload=json.loads(event.payload)
                 )
-                self._handle_pipeline_event(pipeline_events_iothub.MethodRequest(method_received))
+                self.handle_pipeline_event(pipeline_events_iothub.MethodRequest(method_received))
+
+            elif mqtt_topic.is_twin_response_topic(topic):
+                rid = mqtt_topic.get_twin_request_id_from_topic(topic)
+                status_code = mqtt_topic.get_twin_status_code_from_topic(topic)
```
olivakar: nice..I did the same change to provisioning this time.

azure-iot-device/azure/iot/device/iothub/models/twin.py
```diff
@@ -0,0 +1,28 @@
...
+    :ivar desired_properties: The desired properties for the Twin.  These are properties
+      which are sent _to_ the device or module to indicate the _desired_ state of the device
+      or module
+    :type desired_properties: dict, str, int, float, bool, or None (JSON compatible values)
+    :ivar reported_properties: The reported properties for the Twin.  These are properties
+      which are sent _from_ the device or module to indicate the _actual_ state of the device.
+    :type reported_properties: dict, str, int, float, bool, or None (JSON compatible values)
+    """
+
+    # OPEN QUESTION: can desired_properties and reported_properties only be dict, or can they be any JSON-compatible value?
```
olivakar: > OPEN QUESTION [](start = 5, length = 14)

is there a set of properties that remain constant for all devices and modules ?

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -183,6 +183,62 @@ def callback():
...
+    def get_twin(self):
+        """
+        Gets the device or module twin from the Azure IoT Hub or Azure IoT Edge Hub service.
+
+        This is a synchronous call, meaning that this function will not return until the twin
+        has been retrieved from the service.
+
+        :returns: Twin object which was retrieved from the hub
+        """
+        # OPEN QUESTON: does this have a timeout parameter and does it raise on timeout?
```
olivakar: >OPEN QUESTON [](start = 10, length = 12)

I think it should have a time out.

cartertinney: This definitely needs to follow the same pattern as the receive APIs - it needs both the `block` and `timeout` optional params.

---
In reply to: [287531765](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r287531765) [](ancestors = 287531765)

BertKleewein: For block,  that parameter is meaningless because this is a case where we're sending a request to the hub and waiting for a response.  Setting block=False would effectively say that the library needs to send a request but not wait for the response.

For timeout, I agree that we need a mechanism for this, but timeout needs to apply to almost all operations and the timeout for get_twin is different than the timeout for receive_method_request.  In receive_method-request, a timeout means "nothing is wrong, but we didn't receive anything".  In get_twin, a timeout means "I tried to get the twin, but it didn't come back.  Something is wrong."

Timeout also implies that we need a cancellation mechanism that we don't have.  If a get_twin operation times out, there should be a way to cancel that operation, and that doesn't exist yet.


---
In reply to: [288237256](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288237256) [](ancestors = 288237256,287531765)

azure-iot-device/azure/iot/device/iothub/transport/mqtt/mqtt_topic.py
BertKleewein: That's a good idea, but I think that would make the code a little harder to read and the same value isn't there (because the base is constant, unlike telemetry which is based on device_id and module_id).

---
In reply to: [287524645](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r287524645) [](ancestors = 287524645)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
cartertinney: Yeah I was working on editing the comment when you replied. Ignore the previous comment, per our discussion.

---
In reply to: [288271386](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288271386) [](ancestors = 288271386,288237256,287531765)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -18,16 +18,16 @@
```
cartertinney: As per discussion, we should consider using a twin object with some kind of reported_properties delta

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -18,16 +18,16 @@
 class AbstractIoTHubClient(object):
     """A superclass representing a generic client. This class needs to be extended for specific clients."""
 
-    def __init__(self, pipeline):
```
cartertinney: As per discussion, we should consider if we want this to also apply the patch to a twin object

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -72,7 +72,7 @@ def callback():
             connect_complete.set()
             logger.info("Successfully connected to Hub")
 
-        self._pipeline.connect(callback=callback)
+        self._transport.connect(callback=callback)
         connect_complete.wait()
```
cartertinney: > InboxEmpty needs to be replaced with something more appropriate [](start = 43, length = 64)

I think we should hold off on this until a separate PR that adds across the board client exception story

azure-iot-device/azure/iot/device/iothub/transport/constant.py
```diff
@@ -3,10 +3,12 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
-"""This module contains constants realted to the pipeline package.
+"""This module contains constants realted to the transport package.
```
cartertinney: These cannot be enabled/disabled separately from what I understand, due to limitations of IoTHub

azure-iot-device/azure/iot/device/iothub/pipeline/constant.py
```diff
@@ -10,3 +10,5 @@
 C2D_MSG = "c2d"
 INPUT_MSG = "input"
 METHODS = "methods"
+TWIN = "twin"
+TWIN_PATCHES = "twin_patches"
```
cartertinney: Not sure if we actually need these to be separate

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -162,3 +162,82 @@ def __init__(self, sas_token, callback=None):
...
+    """
+
+    def __init__(self, request_type, verb, resource, request_body, callback=None):
+        """
+        Initializer for SendIotRequestAndWaitForResponse objects
+
+        :param str request_type: The type of request.  This is a string which is used by transport-specific stages to
+          generate the actual request.  For example, if request_type is "twin", then the iothub_mqtt stage will convert
+          the request into an MQTT publish with topic that begins with $iothub/twin
+        :param str verb: The verb for the request, such as "POST", "GET", etc.
```
cartertinney: > :param str verb: The verb for the request, such as "POST", "GET", etc. [](start = 7, length = 71)

1) I think the correct term for these are HTTP methods, not verbs

2) What does `POST` or `GET` even mean in the context of MQTT?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -162,3 +162,82 @@ def __init__(self, sas_token, callback=None):
...
+        :param str resource: The resource that the verb is acting on.  For twin request with verb "GET", this is most
```
cartertinney: >resource [](start = 19, length = 8)

Like a URI?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -162,3 +162,82 @@ def __init__(self, sas_token, callback=None):
...
+          likely the string "/" which retrieves the entire twin resource
+        :param request_body: The body of the request.  This is a required field, and a single space can be used to denote
+          an empty body.
+        :type request_body: Undefined
```
cartertinney: >Undefined [](start = 28, length = 9)

I thought it was a string?

cartertinney:
>         }

Need to add twin in here

---
Refers to: azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_adapter.py:35 in ef5e278. [](commit_id = ef5e278533866478ab644c563eeb1ec3689b3d5e, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -45,3 +46,42 @@ def _run_op(self, op):
             )
         else:
             self.continue_op(op)
+
+
+class HandleTwinOperations(PipelineStage):
+    """
+    PipelineStage which handles twin operations. In particular, it converts twin GET and PATCH
+    operations into SendIotRequestAndWaitForResponse operations.  This is done at the IotHub level because
+    there is nothing transport-specific about this code.  The transport-specific implementation
```
cartertinney: > transport-specific [](start = 20, length = 19)

Even if we end up keeping "transport" around in the API for legacy purposes, I don't think we should keep referring to it internally. Makes no sense to anyone who doesn't know the history of the SDK. We're talking about protocols here.

azure-iot-device/azure/iot/device/iothub/models/twin.py
BertKleewein: There is not.  the desired and reported items at the root are the only guaranteed items.

---
In reply to: [287529657](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r287529657) [](ancestors = 287529657)

azure-iot-device/azure/iot/device/iothub/pipeline/constant.py
BertKleewein: This is somewhat internal and it needs to be separate because they represent two subsets of functionality.  TWIN is for the app-initiated "GetTwin" and 'SendTwinReportedPropertiesPatch".  TWIN_PATCHES Is for the service-initiated ReceiveTwinDesiredPropertiesPatch.  The app-initiated ops need the response topic and the service-initiated op needs a downstream topic.  

(This is also related to something that Tim told you -- by separating these, the user can use the twin for desired properties, or they can use it for reported properties, but they're not required to subscribe for both.)

---
In reply to: [288334435](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288334435) [](ancestors = 288334435)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
BertKleewein: I've seen the term "method" used in other codebases.  I'll change it.

These method names mean nothing in the concept of MQTT, but they do have meaning in the context of iothub, where the convention is to use MQTT and AMQP following a pattern that is like "REST-over-MQTT" or "REST-over-AMQP".  in this way, the iothub protocols are designed for the least-common-denominator HTTP.  They don't really take advantage of MQTT and AMQP as much as they could.  This is why provioning, for instance, needs to poll the service for status instead of relying on a service-push.

---
In reply to: [288341305](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288341305) [](ancestors = 288341305)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
BertKleewein: Yep, like the R in URI.  REST-over-MQTT FTW, baby!

---
In reply to: [288342591](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288342591) [](ancestors = 288342591)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
BertKleewein: Everything that comes in over MQTT arrives as a byte array.  At this level, the code doesn't care what it is.  It's only the producer and the consumer that care.  I don't want to define it if I don't have to because i don't want to force multiple conversions if I don't have to. 

---
In reply to: [288342762](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288342762) [](ancestors = 288342762)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
BertKleewein: I'm opening a thread.  protocol is a very generic word.  transport has a very specific definition that doesn't quite match either.  The last T in MQTT stands for Transport, but the last P in HTTP and AMQP stands for Protocol.  But, in the OSI 7-layer stack, HTTP is an Application Protocol and AMQP and MQTT would be Session Protocols.  In the case of MQTT, there are at least 4 protocols involved - MQTT, TCP, IP, and Ethernet (or whatever physical connection you're using).  Saying "the protocol" is meaningless unless we have agreement on what "a protocol" is.  

---
In reply to: [288349774](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288349774) [](ancestors = 288349774)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -137,7 +154,20 @@ def _handle_pipeline_event(self, event):
                 method_received = MethodRequest(
                     request_id=rid, name=method_name, payload=json.loads(event.payload)
                 )
-                self._handle_pipeline_event(pipeline_events_iothub.MethodRequest(method_received))
+                self.handle_pipeline_event(pipeline_events_iothub.MethodRequest(method_received))
+
+            elif mqtt_topic_iothub.is_twin_response_topic(topic):
```
cartertinney: It is kinda weird that this has to go from generic (some event) to specific (this is a twin response) to generic (make it a generic IoTResponseEvent) to specific again (resolve the GetTwin operation).

I don't necessarily think this is wrong, nor do I see a better way to do it, but it is weird IMO

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: What I mean is, is the resource always provided in URI form? "resource" doesn't tell us much about what the parameter *is*, just what it represents. If it's a URI, lets call it a URI.

---
In reply to: [288677292](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288677292) [](ancestors = 288677292,288342591)

BertKleewein: It's not in URI form.  It can be _part_ of a URI, and, if this was REST instead of our bogus REST-over-MQTT, it would be used to build a URI (along with hostname and query parameters).  iothub steals the concept of a resource from REST, just like it steals the concept of a method.  For example, in the REST call to GET [https://api.github.com/repos/Azure/azure-iot-sdk-node/pulls/346](https://api.github.com/repos/Azure/azure-iot-sdk-node/pulls/346), the resource for this request (the entity that you're acting upon) is "/repos/Azure/azure-iot-sdk-node/pulls/346".

I think I can phrase this better in the comments by saying that it is the resource "in the REST API sense of the word" and I should add this same clause to the doc for the method param.

Does that address your concern, or am I still missing the point of your question?  


---
In reply to: [288742923](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288742923) [](ancestors = 288742923,288677292,288342591)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
BertKleewein: yes, very weird.

---
In reply to: [288741545](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288741545) [](ancestors = 288741545)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: It's unfortunate that "method" also isn't a great name, since it can be confused for the method of an object... but I think the docs make it clear, so I'm okay with it

---
In reply to: [288663168](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288663168) [](ancestors = 288663168,288341305)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: No, that was the point I was getting at. How about resource_location?

---
In reply to: [288763007](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288763007) [](ancestors = 288763007,288742923,288677292,288342591)

azure-iot-device/azure/iot/device/iothub/pipeline/constant.py
cartertinney: My understanding of what Tim said was the opposite - that Twin always has two AMQP links going.

But even if I'm mistaken, and we do need separate ones, SEND_TWIN and RECEIVE_TWIN might be better, since neither TWIN nor TWIN_PATCHES clearly indicates a direction (patches could be sent or received for instance).

Furthermore, why do we need a TWIN/SEND_TWIN feature enable/disable? We don't need to enable sending telemetry, or sending a method response. Why do we need to enable the sending of a twin patch, or the request for a twin?

---
In reply to: [288657819](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288657819) [](ancestors = 288657819,288334435)

azure-iot-device/azure/iot/device/iothub/models/twin.py
```diff
@@ -0,0 +1,28 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module contains classes related to device twin and module twin functionality
+"""
+
+
+class Twin(object):
```
cartertinney: >Twin(object): [](start = 6, length = 13)

Do we need metadata fields on this object? Or are desired/reported properties enough?

azure-iot-device/azure/iot/device/iothub/pipeline/constant.py
BertKleewein: I think (if I heard correctly) that Tim way saying that the Java implementation had both links going, but it was theoretically possible to only use one.

This whole mess happens because MQTT and AMQP don't support the idea of replying to something.  Sending a request is one operation (upstream) and receiving the response means listening for a second operation (downstream).  We don't need to enable anything to _send_ the patch or to send the GET request.  We need the enable because we need to _listen_ for the response.  For GET, the request gets published on one topic,then we have to listen on a different topic for the response.   The "TWIN" feature name enables us to listen for the response., and it just happens that GET_TWIN and SEND_TWIN_PATCH both use the same downstream channel.  

I really struggled with this, because it's not really SEND_TWIN vs RECEIVE_TWIN, its GET_AND_SEND_TWIN vs RECEIVE_TWIN.  There's a difference, but it's hard to explain and hard to understand.  Maybe there should be one thing to enable, and it enables both scenarios under the covers.  Or maybe I make it 3 features and 2 of them happen to overlap.  That would be a little simpler (because there's a 1:1 mapping between feature and topic).



---
In reply to: [288770383](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288770383) [](ancestors = 288770383,288657819,288334435)

azure-iot-device/azure/iot/device/iothub/models/twin.py
BertKleewein: We get some $version metadata (I think maybe only the top level), and that's it.  I think maybe there's value in adding it, but I'll have to chew on that.  

---
In reply to: [288771330](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288771330) [](ancestors = 288771330)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
BertKleewein: I am unable to resolve this comment.  I will try again. 

---
In reply to: [288331321](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288331321) [](ancestors = 288331321)

BertKleewein: <bzzzzt> unable to resolve.  github bug?  codeflow bug?

---
In reply to: [288777099](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288777099) [](ancestors = 288777099,288331321)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -252,3 +252,38 @@ def encode_properties(message_to_send, topic):
         topic += user_properties_encoded
 
     return topic
+
+
+def get_twin_response_topic_for_subscribe():
```
cartertinney: Can we organize these in with the like functions related to other operations?

e.g. `get_<thing>_topic_for_subscribe()` functions are all grouped together, `is_<thing>_topic()` functions are all grouped together, etc.

azure-iot-device/azure/iot/device/iothub/sync_clients.py
cartertinney: It's because this comment only shows up on Revision 4 or less, starting with 5 it's "unavailable" because the line it was on got deleted.

---
In reply to: [288777644](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288777644) [](ancestors = 288777644,288777099,288331321)

azure-iot-device/azure/iot/device/iothub/pipeline/constant.py
cartertinney: I followed up with Tim (and also chatted with Zoltan) and it does seem like the operations are not independent. But I think we should check with service.

---
In reply to: [288776071](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288776071) [](ancestors = 288776071,288770383,288657819,288334435)

azure-iot-device/azure/iot/device/iothub/models/twin.py
cartertinney: Yeah, I'm not really sure either, I just know Twin can get complicated so I figured I'd raise it. We can always add it later if we need it.

---
In reply to: [288776802](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288776802) [](ancestors = 288776802,288771330)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -252,3 +252,38 @@ def encode_properties(message_to_send, topic):
...
+
+
+def is_twin_response_topic(topic):
+    return topic.startswith("$iothub/twin/res/")
+
+
+def get_twin_request_id_from_topic(topic):
+    parts = topic.split("/")
+    if is_twin_response_topic(topic) and len(parts) >= 4:
+        properties = _extract_properties(topic.split("?")[1])
```
cartertinney: >topic.split("?")[1] [](start = 41, length = 19)

I wonder if we should move the `topic.split("?")[1]` logic inside `_extract_properties`, since it's always kind of needed.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
BertKleewein: that's a great idea.  I want to revisit this file and put it under test in a separate PR.  Its reached what a friend of mine would call critical mess.  It's sufficiently complex that it needs its own tests.  Plus, this manual URI parsing needs to be replaced by library calls -- splitting strings is just too error-prone.  

---
In reply to: [288783848](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288783848) [](ancestors = 288783848)

azure-iot-device/azure/iot/device/iothub/pipeline/constant.py
BertKleewein: sure, but don't forget that this feature_name stuff is all internal.  We're not exposing any of it to users.  

---
In reply to: [288790187](https://github.com/Azure/azure-iot-sdk-python-preview/pull/96#discussion_r288790187) [](ancestors = 288790187,288776071,288770383,288657819,288334435)

PR #97, opened 2019-05-23
pipeline structures for provisioning

olivakar:
Made into a pipeline
Renamed and moved relevant files as per discussion
Updated tests

Some parts of code that dealt with extraction of key-values from topic has been removed from request response provider and shifted to the `_handle_pipeline_event` in the pipeline_stages_provisioning_mqtt` file. Hence the unit tests have been updated.

NO new tests have been added. Will work on those now.


azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
```diff
@@ -8,30 +8,27 @@
...
-def create_from_security_client(provisioning_host, security_client, transport_choice):
+def create_from_security_client(security_client, transport_choice):
     """
     Creates different types of registration clients which can enable devices to communicate with Device Provisioning
     Service based on parameters passed.
-    :param provisioning_host: Host running the Device Provisioning Service. Can be found in the Azure portal in the
-    Overview tab as the string Global device endpoint
     :param security_client: Instance of Security client object which can be either of SymmetricKeySecurityClient,  TPMSecurtiyClient or X509SecurityClient
-    :param transport_choice: A string representing the transport the user wants
+    :param transport_choice: A string representing the pipeline the user wants
```
cartertinney: I don't think this makes much semantic sense.

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
```diff
@@ -8,30 +8,27 @@
...
     :return: A specific registration client based on parameters and validations.
     """
     transport_choice = transport_choice.lower()
     if transport_choice == "mqtt":
         if isinstance(security_client, SymmetricKeySecurityClient):
-            mqtt_state_based_provider = StateBasedMQTTProvider(provisioning_host, security_client)
+            mqtt_state_based_provider = PipelineAdapter(security_client)
```
cartertinney: >mqtt_state_based_provider [](start = 12, length = 25)

Pretty sure this is a pipeline, and not a state_based_provider

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
```diff
@@ -8,30 +8,27 @@
...
             return SymmetricKeyProvisioningDeviceClient(mqtt_state_based_provider)
             # TODO : other instances of security provider can also be checked before creating mqtt and client
         else:
             raise ValueError("A symmetric key security provider must be provided for MQTT")
 
     else:
-        raise NotImplementedError("This transport has not yet been implemented")
+        raise NotImplementedError("This pipeline has not yet been implemented")
```
cartertinney: >raise NotImplementedError("This pipeline has not yet been implemented") [](start = 8, length = 71)

the word we're looking for here is probably protocol, but IDK what we'll end up using.

azure-iot-device/azure/iot/device/provisioning/internal/polling_machine.py
```diff
@@ -9,7 +9,7 @@
 import traceback
 from threading import Timer
 from transitions import Machine
-from ..transport import constant
+from ..pipeline import constant
```
cartertinney: >.. [](start = 5, length = 2)

relative imports that involve going up a directory make me queasy.

If you have to go up a level, you should probably use the full import namespace instead

azure-iot-device/azure/iot/device/provisioning/internal/request_response_provider.py
```diff
@@ -21,74 +21,73 @@ class RequestResponseProvider(object):
 
     def __init__(self, state_based_provider):
 
-        self._state_based_provider = state_based_provider
+        self._mqtt_transport = state_based_provider
```
cartertinney: What transport? There is no transport anymore.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_adapter.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class PipelineAdapter(object):
+    def __init__(self, security_client):
+        """
+        Constructor for instantiating a pipeline
+        :param security_client: The security client which stores credentials
+        """
+        # Event Handlers - Will be set by Client after instantiation of Transport
+        self.on_transport_connected = None
```
cartertinney: >on_transport_connected [](start = 13, length = 22)

Probably should change these names

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_events_provisioning.py
```diff
@@ -0,0 +1,26 @@
...
+        :param request_id : The id of the request to which the response arrived.
+        :param status_code: The status code received in the topic.
+        :param key_values: A dictionary containing key mapped to a list of values that were extarcted from the topic.
+        :param response: The response received from a registration process
+        """
+        super(RegistrationResponseEvent, self).__init__()
+        self.rid = request_id
+        self.status_code = status_code
+        self.key_values = key_values
+        self.response = response
```
cartertinney: Consider maybe using a model to represent a Registration Response that has these fields?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
```diff
@@ -0,0 +1,154 @@
...
+        :param id_scope: The ID scope is used to uniquely identify the specific provisioning service the device will
+        register through. The ID scope is assigned to a Device Provisioning Service when it is created by the user and
+        is generated by the service and is immutable, guaranteeing uniqueness.
+        """
+        super(SetSymmetricKeySecurityClientArgs, self).__init__(callback=callback)
+        self.provisioning_host = provisioning_host
+        self.registration_id = registration_id
+        self.id_scope = id_scope
+        # self.symmetric_key = symmetric_key
+
```
cartertinney: Remove this if not used.

azure-iot-device/azure/iot/device/provisioning/internal/request_response_provider.py
```diff
@@ -21,74 +21,73 @@ class RequestResponseProvider(object):
...
-        logger.info("Received response:{} on topic:{}".format(response, topic))
-
-        if topic.startswith("$dps/registrations/res/"):
-            topic_parts = topic.split("$")
-            key_value_dict = urllib.parse.parse_qs(topic_parts[POS_QUERY_PARAM_PORTION])
-            rid = key_value_dict["rid"][0]
-        # TODO Not DPS may have other ways to retrieve rid
+        logger.info("Received response {}:".format(response))
 
         if rid in self._pending_requests:
```
cartertinney: >if rid in self._pending_requests: [](start = 8, length = 33)

Is the management of pending requests threadsafe?

This design of how pending requests are managed resembles the model used down in the MQTTProvider, and I just had to fix a threading bug there. You may find that something similar will arise here - I'd check out the latest changes to the MQTTProvider in master branch and see if what's going on there applies here.

azure-iot-device/azure/iot/device/provisioning/internal/request_response_provider.py
```diff
@@ -21,74 +21,73 @@ class RequestResponseProvider(object):
...
-        self._state_based_provider.connect(callback=callback)
+        self._mqtt_transport.connect(callback=callback)
 
     def disconnect(self, callback=None):
         if callback is None:
             callback = self._on_connection_state_change
-        self._state_based_provider.disconnect(callback=callback)
+        self._mqtt_transport.disconnect(callback=callback)
 
     def publish(self, topic, request, callback=None):
```
cartertinney: >def publish(self, topic, request, callback=None): [](start = 4, length = 49)

Method is unused, and should be deleted

azure-iot-device/azure/iot/device/provisioning/internal/polling_machine.py
```diff
@@ -213,23 +207,21 @@ def _query_operation_status(self, event_data):
         self._operations[rid] = constant.PUBLISH_TOPIC_QUERYING.format(rid, operation_id)
         self._request_response_provider.send_request(
             rid=rid,
-            topic=constant.PUBLISH_TOPIC_QUERYING.format(rid, operation_id),
             request=" ",
```
cartertinney: >request=" " [](start = 12, length = 11)

Is this the payload? Why is it always a " "?

Is there some way we could make Request an object that combines this payload with rid to be less confusing as it moves down the stack? I had to do a lot of tracing to realize that the "request" being used in the pipeline is actually this payload.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
```diff
@@ -0,0 +1,154 @@
...
+        :param rid : The id of the request being sent
+        :param request: The request that we are sending to the service
+        :param Function callback: The function that gets called when this operation is complete or has failed.
+         The callback function must accept A PipelineOperation object which indicates the specific operation which
+         has completed or failed.
+        """
+        super(SendRegistrationRequest, self).__init__(callback=callback)
+        self.rid = rid
+        self.request = request
```
cartertinney: As mentioned in another comment, these fields are kind of confusing due to the lack of encapsulation. `request` actually appears to be some kind of payload?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning_mqtt.py
```diff
@@ -0,0 +1,121 @@
...
+            topic = mqtt_topic.get_topic_for_subscribe()
+            self.continue_with_different_op(
+                original_op=op, new_op=pipeline_ops_mqtt.Unsubscribe(topic=topic)
+            )
+
+        else:
+            # All other operations get passed down
+            self.continue_op(op)
+
+    def _handle_pipeline_event(self, event):
```
cartertinney: This method is the best place to start data encapsulation (if possible)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
```diff
@@ -0,0 +1,154 @@
...
+        :param rid
+        :param request: The request that we are sending to the service
+        :param Function callback: The function that gets called when this operation is complete or has failed.
+         The callback function must accept A PipelineOperation object which indicates the specific operation which
+         has completed or failed.
+        """
+        super(SendQueryRequest, self).__init__(callback=callback)
+        self.rid = rid
+        self.operation_id = operation_id
+        self.request = request
```
cartertinney: Same as above

azure-iot-device/azure/iot/device/provisioning/pipeline/mqtt_topic.py
```diff
@@ -0,0 +1,92 @@
...
+    Topics for responses from DPS are of the following format:
+    $dps/registrations/res/<statuscode>/?$<key1>=<value1>&<key2>=<value2>&<key3>=<value3>
+    :param topic: The topic string
+    """
+    if get_topic_for_response() in topic:
+        return True
+    return False
+
+
+def extract_properties_from_topic(topic):
```
cartertinney: Not sure how relevant this is here, but in the iothub client there's an issue where this function has a generic name, but is specifically tied to a certain type of topic string (messages), and needs to be refactored as a result.

If there are different types of possible topic strings for DPS, this may present some kind of problem for you? Or not. I'm not quite sure the scenarios here. If this is working fine, leave it alone I guess.

When we refactor how this stuff all happens in the IoTHubClient, depending on the approach, we may end up bringing some of the changes over here to reduce duplicate code.

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
olivakar: iothub calls it a transport. are you referring to call it transport and go back to transport rather than pipeline ? the below line is from iothub
`:param transport: The transport that the client will use.`

---
In reply to: [287542349](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287542349) [](ancestors = 287542349)

cartertinney:
I didn't look too much at the tests since getting those all up to snuff will probably be a separate PR, and we're still developing a testing strategy.

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
olivakar: changed to transport protocol as I like that name as well.

---
In reply to: [287542482](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287542482) [](ancestors = 287542482)

azure-iot-device/azure/iot/device/provisioning/internal/request_response_provider.py
olivakar: I don't think there is any issue is similar to the one mqtt provider has. The pending requests have the request or does not have it. After seeing Bert's twin PR today I am thinking what to do if the rid is not there. But that is a different discussion point.

---
In reply to: [287547224](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287547224) [](ancestors = 287547224)

azure-iot-device/azure/iot/device/provisioning/pipeline/mqtt_topic.py
olivakar: as far as I know this would cover all topic strings for DPS

---
In reply to: [287548205](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287548205) [](ancestors = 287548205)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
olivakar: yes. changed to request_body

---
In reply to: [287547711](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287547711) [](ancestors = 287547711)

azure-iot-device/azure/iot/device/provisioning/internal/polling_machine.py
olivakar: changed the name as of now. Lot of these things will see change eventually.  Things may become operations or events. I am not going to encapsulate as of now.

---
In reply to: [287547668](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287547668) [](ancestors = 287547668)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning_mqtt.py
olivakar: its actually not possible here. It is not advisable to create a DPS specific response and send. The request response provider may become a generic pipeline much like what the Twin has now. In that model a very generic response needs to be sent upwards.

---
In reply to: [287547898](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287547898) [](ancestors = 287547898)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_events_provisioning.py
olivakar: renamed for now.

---
In reply to: [287546829](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287546829) [](ancestors = 287546829)

BertKleewein:
Why did you add azure-iot-device/__init__.py?  I don't think we want this file there.

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
BertKleewein: agreed that we should be consistent here, and I also think it's important to be consistent between languages.  I don't think any other SDKs use the term "protocol".  "transport_name" is what I put into iothub.  

---
In reply to: [287548505](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287548505) [](ancestors = 287548505,287542349)

BertKleewein:
> class PollingMachine(object):

I'm curious to hear what your long term plan is for this class.  So far, you've put some state machines on top of a pipeline, but the goal should be to get rid of state machines and make everything into pipeline stages.  

---
Refers to: azure-iot-device/azure/iot/device/provisioning/internal/polling_machine.py:25 in 44a0835. [](commit_id = 44a0835116459a28da09c9a1eb1097ae1d37cc53, deletion_comment = False)

BertKleewein:
> class RequestResponseProvider(object):

I have the same question about long-term plans for this object.  Is this eventually going to become a pipeline stage?  I designed the CoordinateRequestAndResponse stage in my Twin PR so it _could_ also be used for provisioning.

---
Refers to: azure-iot-device/azure/iot/device/provisioning/internal/request_response_provider.py:17 in 44a0835. [](commit_id = 44a0835116459a28da09c9a1eb1097ae1d37cc53, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
cartertinney: Yeah as long as we get it consistent between the two now, we can figure out what to do after. Let's go with `transport_name` for now.

---
In reply to: [287575351](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r287575351) [](ancestors = 287575351,287548505,287542349)

olivakar:
> class PollingMachine(object):

I am not sure yet. I thought we discussed with Pierre that this would be left for now and we can come back to it for pipeline at a later stage

---
In reply to: [495949513](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#issuecomment-495949513) [](ancestors = 495949513)

---
Refers to: azure-iot-device/azure/iot/device/provisioning/internal/polling_machine.py:25 in 44a0835. [](commit_id = 44a0835116459a28da09c9a1eb1097ae1d37cc53, deletion_comment = False)

olivakar:
> class RequestResponseProvider(object):

I checked your PR. yes that stage can be used for provisioning.  based on discussion with we are supposed to do this later

---
In reply to: [495949878](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#issuecomment-495949878) [](ancestors = 495949878)

---
Refers to: azure-iot-device/azure/iot/device/provisioning/internal/request_response_provider.py:17 in 44a0835. [](commit_id = 44a0835116459a28da09c9a1eb1097ae1d37cc53, deletion_comment = False)

azure-iot-device/tests/common/pipeline_test.py
```diff
@@ -0,0 +1,118 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
```
BertKleewein: >information [](start = 10, length = 11)

I'm not sure why the files in this directory are showing up in your commit.  They already exist in master.

azure-iot-device/tests/provisioning/internal/test_polling_machine.py
```diff
@@ -34,8 +35,10 @@
 
 
 class TestRequestResponseProvider(RequestResponseProvider):
```
BertKleewein: >TestRequestResponseProvider [](start = 6, length = 27)

suggest a better name.  This looks like a test class.  probably harmless, but still confusing.  
PytestWarning: cannot collect test class 'TestRequestResponseProvider' because it has a __init__ constructor

azure-iot-device/tests/common/pipeline_test.py
BertKleewein: shoot.  I messed up.  This is not in master yet.  I'm not sure how I did that, but I'll get it in there asap.

---
In reply to: [288537437](https://github.com/Azure/azure-iot-sdk-python-preview/pull/97#discussion_r288537437) [](ancestors = 288537437)

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
```diff
@@ -8,30 +8,27 @@
...
     :param security_client: Instance of Security client object which can be either of SymmetricKeySecurityClient,  TPMSecurtiyClient or X509SecurityClient
     :param transport_choice: A string representing the transport the user wants
     :return: A specific registration client based on parameters and validations.
     """
     transport_choice = transport_choice.lower()
     if transport_choice == "mqtt":
         if isinstance(security_client, SymmetricKeySecurityClient):
-            mqtt_state_based_provider = StateBasedMQTTProvider(provisioning_host, security_client)
-            return SymmetricKeyProvisioningDeviceClient(mqtt_state_based_provider)
+            mqtt_pipeline_adapter = PipelineAdapter(security_client)
```
cartertinney: >mqtt_pipeline_adapter [](start = 12, length = 21)

I don't love this naming either, but for now, while the object is called PipelineAdapter this should be fine (I suspect we should maybe change it at some point to be `mqtt_pipeline = ProvisioningPipeline(security_client)`)

azure-iot-device/azure/iot/device/provisioning/sk_provisioning_device_client.py
```diff
@@ -21,13 +20,13 @@ class SymmetricKeyProvisioningDeviceClient(ProvisioningDeviceClient):
     using Symmetric Key authentication.
     """
 
-    def __init__(self, mqtt_state_based_provider):
+    def __init__(self, mqtt_provisioning_pipeline):
```
cartertinney: >mqtt_provisioning_pipeline [](start = 23, length = 26)

Probably shouldn't have mqtt

PR #98, opened 2019-05-31
Added async client

olivakar:
- `ProvisioningDeviceClient` renamed to `AbstractProvisioningDeviceClient`
- The `create_from_security_client` has been moved to this abstract class


azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
```diff
@@ -0,0 +1,79 @@
...
+def log_on_register_complete(result=None, error=None):
+    # This could be a failed/successful registration result from the HUB
+    # or a error from polling machine. Response should be given appropriately
+    if result is not None:
+        if result.status == "assigned":
+            logger.info("Successfully registered with Hub")
+        else:  # There be other statuses
+            logger.error("Failed registering with Hub")
+    if error is not None:  # This can only happen when the polling machine runs into error
+        logger.info(error)
```
cartertinney: Importing this helper elsewhere is kind of weird. I see why you do it, but by exposing it between modules, it's part of the module API, and thus kind of an independent unit rather than a helper (and thus subject to testing as a unit).

That said, there isn't an elegant solution I can think of. Duplicating it in both sync and async clients is probably what I would do, but I don't think that's necessarily better.

I'm fine with it as is, but it's just weird and I wanted to comment on it.

azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(self, provisioning_pipeline):
+        """
+        Initializes the provisioning client.
+        :param provisioning_pipeline: Instance of the provisioning pipeline object.
+        """
+        self._provisioning_pipeline = provisioning_pipeline
+
+    @classmethod
+    def create_from_security_client(cls, security_client, protocol_name):
```
cartertinney: > def create_from_security_client(cls, security_client, protocol_name): [](start = 3, length = 70)

For now I think this works, but I suspect we'll have to revisit the client architecture entirely when X509 comes along.

azure-iot-device/samples/advanced-hub-scenarios/register_symmetric_key.py
```diff
@@ -0,0 +1,53 @@
...
+
+    await provisioning_device_client.register()
+
+
+async def main():
+    task1 = asyncio.create_task(register_device())
+    await task1
+
+    # If using Python 3.6 or below, use the following code instead of asyncio.run(main()):
+    # await register_device()
```
cartertinney: I don't think this is true

azure-iot-device/tests/provisioning/aio/test_async_sk_provisioning_device_client.py
```diff
@@ -0,0 +1,141 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
cartertinney: I have some quibbles about some of the test structure and naming here, but we can revisit them after - once iothub is consistent, we can bring dps up to par.

azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
olivakar: yes that is the next step

---
In reply to: [289559936](https://github.com/Azure/azure-iot-sdk-python-preview/pull/98#discussion_r289559936) [](ancestors = 289559936)

azure-iot-device/tests/provisioning/aio/test_async_sk_provisioning_device_client.py
olivakar: okay

---
In reply to: [289561458](https://github.com/Azure/azure-iot-sdk-python-preview/pull/98#discussion_r289561458) [](ancestors = 289561458)

azure-iot-device/samples/advanced-hub-scenarios/register_symmetric_key.py
olivakar: yes it will. The upper one with the task is better convention. Specially when there are multiple tasks there is some difference versus the lower one. So preferred way is upper whenever possible.

---
In reply to: [289560351](https://github.com/Azure/azure-iot-sdk-python-preview/pull/98#discussion_r289560351) [](ancestors = 289560351)

azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
olivakar: Yes it a little weird. it does not do much but still its multiple lines of code. I can put this in a separate module but I felt weird doing it. 

---
In reply to: [289559526](https://github.com/Azure/azure-iot-sdk-python-preview/pull/98#discussion_r289559526) [](ancestors = 289559526)

PR #101, opened 2019-06-03
Test Revisions for Requirements output

cartertinney:
* Adjusted device-iothub client tests (except for auth - that's a future PR)
* Adjusted device-common tests
* Added additional tests
* Made minor adjustments to implementation


azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -200,37 +236,201 @@ def test_receive_method_request_default_mode(self, mocker, client, method_name):
...
-class TestIoTHubModuleClient(ClientSharedTests):
+################
+# DEVICE TESTS #
+################
+class IoTHubDeviceClientTestsConfig(object):
+    client_class = IoTHubDeviceClient
+
+    @pytest.fixture
+    def client(self, pipeline):
+        return IoTHubDeviceClient(pipeline)
```
cartertinney: This class exists to prevent having to add fixtures to all the Device test classes. I'm looking for a better way to do it, since I'd rather avoid multiple inheritance in the test classes, but this works for now at least.

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -200,37 +236,201 @@ def test_receive_method_request_default_mode(self, mocker, client, method_name):
...
+
+
+@pytest.mark.describe("IoTHubDeviceClient (Synchronous) - Instantiation")
+class TestIoTHubDeviceClientInstantiation(
```
cartertinney: This class inherits from the SharedClientInstantiationTests that define most of the tests for inheritance, as well as the DeviceClientTestsConfig which provides a fixture. Then it is extended with a test specific to instantiation of a DeviceClient.

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -200,37 +236,201 @@ def test_receive_method_request_default_mode(self, mocker, client, method_name):
...
+@pytest.mark.describe("IoTHubDeviceClient (Synchronous) - .from_authentication_provider()")
+class TestIoTHubDeviceClientFromAuthenticationProvider(
+    SharedClientFromAuthenticationProviderTests, IoTHubDeviceClientTestsConfig
+):
+    client_class = IoTHubDeviceClient  # TODO: find a way to not need this
+
+
+@pytest.mark.describe("IoTHubDeviceClient (Synchronous) - .connect()")
+class TestIoTHubDeviceClientConnect(SharedClientConnectTests, IoTHubDeviceClientTestsConfig):
+    pass
```
cartertinney: >pass [](start = 4, length = 4)

There are no additional DeviceClient specific tests for .connect(), so we just pass here, and it will only run the tests from SharedClientConnectTests

BertKleewein:
This isn't bad, but the one-class-per-method pattern makes the amount of boilerplate grow exponentially.  I'm thinking about pipeline stages here.  Imagine you have an abstract base class with 10 methods, and you have 5 derived classes.  That means you have 50 instances of this 3 line copypasta:
@pytest.mark.describe("IoTHubDeviceClient (Synchronous) - .send_event()")
class TestIoTHubDeviceClientSendEvent(SharedClientSendEventTests, IoTHubDeviceClientTestsConfig):
    pass

I have an idea about how to make this shorter, but it involves monkeypatching the test files.  This is, more or less, how I've done this in node before.  I'm still forming this in my head, but the rough idea is:
In this PR, you have a class called SharedClientInstantiationTests and if you want to run these tests on a sync device client, you make a derived class that inherits from this class and a config class IoTHubDeviceClientTestsConfig.

Instead, what if we do something like this which basically creates derived classes on-the-fly when the test files get loaded:

```
# in base_test.py

# function which creates a test class for a given class-under-test and adds it to the module namespace.
def add_instantiation_tests(class_under_test, module):

    # create a test class with local scope.  The name of the class here doesn't matter because the 
    # local-scoped class isn't where pytest is going to find the test.  
    # Instead, we'll add this this local class to the module namespace, and pytest will be able to find it.
    
    # Add the @describe decorator though, because (I assume) that adds metadata to the class that pytest will be able to find.
    
    @pytest.mark.describe(class_under_test.__name__)
    class SharedClientInstantiationTests(object):
        # I think we're OK adding a fixture here if fixture loading happens the way I imagine, but I might be wrong. 
        @pytest.fixture
        def client(self, pipeline):
            # notice how we create an object based on the class that was passed into the outer-scope function.  This should work just fine.
            return class_under_test(pipeline)
            
        @pytest.mark.it("Sets on_connected handler in pipeline")
        def test_sets_on_connected_handler_in_pipeline(self, client):
            assert client._pipeline.on_transport_connected is not None
            assert client._pipeline.on_transport_connected == client._on_state_change

        @pytest.mark.it("Sets on_disconnected handler in pipeline")
        def test_sets_on_disconnected_handler_in_pipeline(self, client):
            assert client._pipeline.on_transport_disconnected is not None
            assert client._pipeline.on_transport_disconnected == client._on_state_change

        @pytest.mark.it("Sets on_method_request_received handler in pipeline")
        def test_sets_on_method_request_received_handler_in_pipleline(self, client):
            assert client._pipeline.on_transport_method_request_received is not None
            assert (
                client._pipeline.on_transport_method_request_received
                == client._inbox_manager.route_method_request
            )
            
    # Now, take that class object put it into the module namespace using a name that pytest will find.
    setattr(module, "Test{}Instantiation".format(class_under_test.__name__), SharedClientInstantiationTests)
    
    
# in test_sync_client.py:

from base_test import add_instantiation_tests
from somewhere import IoTHubModuleClient

this_module = sys.modules[__name__]
# this function adds a class called TestIoTHubModuleClientInstantiation to the test_sync_client.py namespace.  
add_instantiation_tests(IoTHubModuleClient, this_module)
```

After the python interpreter loads test_sync_client.py, it will see a class called TestIoTHubModuleClientInstantiation.  It doesn't matter to the interpreter that the class was added to the namespace at load-time -- it just looks like any other class,

BertKleewein:
Oops.  The code grows geometrically, not exponentially...

---
In reply to: [498453061](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#issuecomment-498453061) [](ancestors = 498453061)

cartertinney:
Some thoughts on this:

1) I agree that the boilerplate isn't ideal, but I think it's only an issue in extreme cases - how often would there be five classes inheriting 10 methods in the same file?
2) This solution is quite nice, when there is no extension being done during the inheritance, but when there is (i.e. there are additional Module specific inheritance tests), it breaks down. You could have separate logic that then adds additional methods to the monkeypatched class but then you've added additional (and notably, separate) boilerplate, and, in my experience, monkeypatching methods is kind of strange, and has subtle differences between Python 2 and Python 3 that I have not yet fully explored.
3) I don't think we necessarily need to choose - In the case of these clients, I don't really have a problem with the boilerplate since there's not THAT much of it, but elsewhere, like in your example about five classes inheriting 10 methods, the boilerplate IS much more severe. Perhaps different situations could have different approaches, as there seem to be pros/cons to both (see point 2 for instance)
4) I have also been thinking of some similar stuff to this suggestion for down the road... I spent a good deal of time last night thinking about a potential pytest extension that structures a requirements doc output..... It would be very interesting to do..... but alas, there's no time to take on such a project.... at the moment.... hmmmm.... it would be a good way to "give back" to the open source community.... hmmm....

---
In reply to: [498453400](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#issuecomment-498453400) [](ancestors = 498453400,498453061)

olivakar:
The per method per class pattern is good in general and I understand that if we follow that for 1 method we should follow that for all. But I am only saying this because I feel the number of classes have increased. Do you think if its okay to combine some of them logically? Like connect and disconnect may be together. Or the methods can be together. In cases where the method has too many tests it may still be separate.

azure-iot-device/tests/iothub/test_sync_clients.py
olivakar: Also the config classes are only used for fixtures right ? its only being done to pass client around right ? If the fixtures are defined at top with one being device_client and the other being module client and then used around then probably the Test classes wont need to inherit from the config classes

---
In reply to: [290048211](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290048211) [](ancestors = 290048211)

cartertinney:
Not really no, the test classes each represent a logical unit under test for the requirements document. The only time you want to combine things under the same class, is if it's all part of the same unit - sometimes that makes sense, sometimes it doesn't. 

The AwaitableCallback class in common/async_adapter is a good example of something where the class itself is the unit under test, and thus all functionality for all methods is tested within one testing class. But on the client, each API is clearly it's own distinct unit, and thus I think it would be wrong to combine units, even if it would be handy.

As per our initial testing discussions, the priority here is the test output, not the organization of the test code itself.

---
In reply to: [498473186](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#issuecomment-498473186) [](ancestors = 498473186)

azure-iot-device/tests/iothub/test_sync_clients.py
cartertinney: The problem is that the tests are shared, and thus have to import a fixture with the same name.

For instance, both device and module have "send_event", so there is a SharedClientSendEventTests class. But for the tests to be generic, they need to use a singular client fixture. I can't say "device_client" or "module_client", because the test itself is agnostic.

I could potentially instead have three fixtures at module level, device_client, module_client, and client. client would be a parametrized fixture that returns both device and module client. Then shared tests could rely on client, and module/device specific tests could rely on the specific fixture.

I think that would change how the output is formatted though - I'll need to double check.

---
In reply to: [290088275](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290088275) [](ancestors = 290088275,290048211)

BertKleewein:
should __init__.py in azure-iot-device/ be there?

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
```diff
@@ -0,0 +1,34 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
BertKleewein: >-------------------------------------------------------------------------- [](start = 2, length = 74)

I assume all the files added under azure/iot/device/provisioning are merge mistakes as well as the files added under test/provisoning

azure-iot-device/tests/common/test_async_adapter.py
```diff
@@ -53,27 +80,31 @@ class TestAwaitableCallback(object):
...
-    async def test_can_be_called_using_kwargs(self, mocker, mock_function):
+    @pytest.mark.it("Can be called using explicit keyword arguments")
+    async def test_can_be_called_using_explicit_kwargs(self, mocker, mock_function):
         callback = async_adapter.AwaitableCallback(mock_function)
         result = callback(a=1, b=2, c=3)
         assert mock_function.call_count == 1
         assert mock_function.call_args == mocker.call(a=1, b=2, c=3)
         assert result == mock_function.return_value
 
+    @pytest.mark.it("Can have its callback completion awaited upon")
```
BertKleewein: >("Can have its callback completion awaited upon") [](start = 19, length = 49)

awkward wording.  I have no idea how to word this better without creating a new test class for the completion method, but I'm not sure if that even makes sense.  maybe "Has a method named completion which returns an awaitable that wraps the associated callback"

azure-iot-device/tests/common/test_connection_string.py
```diff
@@ -8,61 +8,68 @@
...
+                id="Duplicate key",
             ),
         ],
     )
     def test_raises_value_error_on_bad_input(self, input_string):
         with pytest.raises(ValueError):
             ConnectionString(input_string)
 
+    @pytest.mark.it(
+        "Provides the input connection string as a string representation for an instance"
```
BertKleewein: >Provides the input connection string as a string representation for an instance" [](start = 9, length = 80)

Awkward.   Maybe "Returns the input connection string as its string representation"?

azure-iot-device/tests/iothub/client_fixtures.py
```diff
@@ -71,4 +67,15 @@ def send_method_response(self, method_response, callback=None):
...
+    """
     return mocker.MagicMock(wraps=FakePipelineAdapter())
+
+
+@pytest.fixture
+def pipeline_manual_cb(mocker):
+    """This fixture is for use in tests where manual triggering of a
+    callback is required
+    """
+    return mocker.MagicMock()
```
BertKleewein: >() [](start = 27, length = 2)

spec=FakePipelineAdapter()?

azure-iot-device/tests/iothub/conftest.py
```diff
@@ -8,7 +8,7 @@
 import pytest
 
 # These fixtures are shared between sync and async clients
-from .client_fixtures import auth_provider, pipeline
+from .client_fixtures import auth_provider, pipeline, pipeline_manual_cb
```
BertKleewein: in conftest.py, you can use pytest_plugins to import fixtures into your test: pytest_plugins = ["tests.iothub.client_fixtures", "tests.common.pipeline_test_fixtures"].  This way you don't have to list every fixture in an import statement.

azure-iot-device/tests/iothub/test_inbox_manager.py
```diff
@@ -12,52 +12,86 @@
...
+    return MethodRequest(request_id="1", name="some_method", payload="{'key': 'value'}")
+
+
+@pytest.mark.describe("InboxManager")
 class TestInboxManager(object):
-    @pytest.fixture(params=inbox_type_list)
-    def inbox_type(self, request):
-        return request.param
+    @pytest.mark.it("Instantiates with an empty C2D inbox")
+    def test_instantiates_with_empty_c2d_inbox(self, manager):
```
BertKleewein: >manager [](start = 53, length = 7)

I know that you know this, but I'm pointing out that you're using a parameterized fixture for the object that you're testing and that's a no-no.  Too bad there's not a syntax to use a parameterized fixture (or any fixture) as a class modifier

azure-iot-device/azure/iot/device/iothub/models/message.py
```diff
@@ -52,3 +53,6 @@ def __init__(
         self.content_encoding = content_encoding
         self.content_type = content_type
         self.output_name = output_name
+
+    def __str__(self):
```
olivakar: >__str__ [](start = 8, length = 7)

if there is a string representation why should it not return all the attributes ?

azure-iot-device/azure/iot/device/iothub/models/message.py
```diff
@@ -52,3 +53,6 @@ def __init__(
...
+        return str(self.data)
```
olivakar: >         [](start = 0, length = 8)

For MQTT data can be string , number or bytes. But in other cases if it can be a complex class this will only work if that class also has a dunder string method defined right ? I see in python it is not a practice to define dunder string , will this pose a problem ?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_adapter.py
```diff
@@ -0,0 +1,172 @@
...
+    pipeline_stages_provisioning,
+    pipeline_stages_provisioning_mqtt,
+)
+from azure.iot.device.provisioning.pipeline import pipeline_events_provisioning
+from azure.iot.device.provisioning.pipeline import pipeline_ops_provisioning
+
+logger = logging.getLogger(__name__)
+
+
+class PipelineAdapter(object):
```
olivakar: >PipelineAdapter [](start = 6, length = 15)

may be this file is not intended to be a part of PR. 
I am pretty sure I changed the name of this file and class to be provisioning pipeline.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py~HEAD
```diff
@@ -0,0 +1,47 @@
...
+# license information.
+# --------------------------------------------------------------------------
+
+import logging
+from azure.iot.device.common.transport import pipeline_ops_base
+from azure.iot.device.common.transport.pipeline_stages_base import PipelineStage
+from . import pipeline_ops_provisioning
+
+
+class UseSymmetricKeySecurityClient(PipelineStage):
```
olivakar: >UseSymmetricKeySecurityClient [](start = 6, length = 29)

same goes for this file. I don't think there are changes here

azure-iot-device/tests/common/test_connection_string.py
olivakar: Another suggestion combining both :

Returns the input connection string as the string representation of an instance

---
In reply to: [290982473](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290982473) [](ancestors = 290982473)

azure-iot-device/tests/provisioning/test_provisioning_device_client_factory.py
```diff
@@ -0,0 +1,63 @@
+# -------------------------------------------------------------------------
```
olivakar: >------------------------------------------------------------------------- [](start = 2, length = 73)

this file should not be there right ?

azure-iot-device/tests/iothub/test_inbox_manager.py
BertKleewein: OK, we talked about this one yesterday.  This is the weird one because the parameter is the type of inbox, not the manager itself.  

---
In reply to: [290984609](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290984609) [](ancestors = 290984609)

azure-iot-device/tests/iothub/aio/test_async_inbox.py
```diff
@@ -26,17 +29,52 @@ def test_check_item_is_in_inbox(self, mocker):
         inbox._put(item)
         assert item in inbox
 
-    # this test also raises runtime warning if it's asynchronous
-    def test__put_adds_item_to_inbox(self, mocker):
+    @pytest.mark.it("Can be checked for whether or not it is empty")
+    @pytest.mark.asyncio
```
BertKleewein: I'm thinking out loud.  Do not change this now, but this phrase is a smell to me that says that the test class needs to be refactored.  "Can be checked" is not an action verb.

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -40,67 +62,154 @@ def test_from_authentication_provider_instantiates_client(
...
         client.disconnect()
         assert pipeline.disconnect.call_count == 1
 
-    def test_send_event_calls_transport(self, client, pipeline):
+    @pytest.mark.it(
+        "Waits for the completion of the 'disconnect' pipeline operation before returning"
+    )
+    def test_waits_for_pipeline_op_completion(self, mocker, client_manual_cb, pipeline_manual_cb):
+        client = client_manual_cb
+        pipeline = pipeline_manual_cb
```
BertKleewein: There's a trick that I learned for this that I really like.  Instead of have a "client automatic" and "client manual" fixture with 2 different names, you could have a single "client" fixture that defaults to automatic callback.  Then, you can have a second fixture that changes your client fixture into manual.  This works because the fixtures have method scope, and your "convert to manual fixture" can only change the client, it can't replace it, but it's still cool.

```
@pytest.fixture
def client()
    return MockAutomaticClient()
    
@pytest.fixture
def make_client_manual(client, mocker):
    client.method_foo = mocker.MagicMock()
    client.method_bar = mocker.MagicMock()
    
def test_blah(client, make_client_manual):
    # client object is from MockAutomaticClient() in the client fixture
    # client.method_foo and method_bar were replaced by the make_client_manual fixture
```

azure-iot-device/tests/provisioning/test_provisioning_device_client_factory.py
cartertinney: I have no idea where this came from. If you say it shouldn't, I'll delete it.

---
In reply to: [290989784](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290989784) [](ancestors = 290989784)

azure-iot-device/tests/iothub/test_inbox_manager.py
cartertinney: Yeah, it's parametrizing different configurations of the same unit under test. The antipattern to avoid is using parametrization to cover multiple units, but since this one is the same unit, it's fine (and in fact, desirable)

---
In reply to: [291182134](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r291182134) [](ancestors = 291182134,290984609)

azure-iot-device/tests/iothub/client_fixtures.py
cartertinney: Nope. I mean, I could, that would also work, but there's no real reason to that I can see (unless we want to be extremely rigid about specing Mocks whenever possible - And even if we did, the more correct thing to do would be to spec on the actual PipelineAdapter)

---
In reply to: [290983218](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290983218) [](ancestors = 290983218)

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -40,67 +62,154 @@ def test_from_authentication_provider_instantiates_client(
...
+        assert pipeline.connect.call_count == 1
 
-    def test_instantiation_sets_on_method_request_received_handler_in_transport(self, client):
-        assert client._pipeline.on_transport_method_request_received is not None
-        assert (
-            client._pipeline.on_transport_method_request_received
-            == client._inbox_manager.route_method_request
-        )
+    @pytest.mark.it("Waits for the completion of the 'connect' pipeline operation before returning")
+    def test_waits_for_pipeline_op_completion(self, mocker, client_manual_cb, pipeline_manual_cb):
```
BertKleewein: >client_manual_cb [](start = 60, length = 16)

how does this even work?  client_manual_cb isn't even defined in this scope as far as I can tell.

cartertinney:
No. It's an artifact of a bad merge.

---
In reply to: [499299203](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#issuecomment-499299203) [](ancestors = 499299203)

azure-iot-device/tests/iothub/test_sync_clients.py
BertKleewein: never mind.  figured it out.

---
In reply to: [291322441](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r291322441) [](ancestors = 291322441)

azure-iot-device/azure/iot/device/iothub/models/message.py
cartertinney: In the absence of `__str__` it defaults to `__repr__`, which comes automatically defined for every type of object.

---
In reply to: [290987260](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290987260) [](ancestors = 290987260)

azure-iot-device/azure/iot/device/iothub/models/message.py
cartertinney: It could. That's a discussion we could have. The main point of adding this is just to have SOMETHING now, because in the user study people kept trying to get the data from Message by printing or doing str convert.

---
In reply to: [290987167](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290987167) [](ancestors = 290987167)

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -40,67 +62,154 @@ def test_from_authentication_provider_instantiates_client(
...
+            cb = pipeline.connect.call_args[1]["callback"]
+            cb()
+
+            # Assert Event is now completed
+            assert event_mock.set.call_count == 1
+
+        event_mock.wait.side_effect = check_callback_completes_event
 
-    def test_connect_calls_transport(self, client, pipeline):
         client.connect()
```
BertKleewein: >       client.connect() [](start = 1, length = 23)

This function and others with the same pattern seem overly complex.  Reading this code, I can't say if it's correct or not.  All I can say from reading this is: 
1. The client.connect function completes
2. If check_callback_completes_event gets called, then a bunch of things get asserted.  

I understand the way this test is _supposed_ to execute.  I just can't say that it does execute that way without running it through a debugger.  For example, if the implementation of client.connect() was replaced with pass, this test would still pass.  

One easy way to work around this would be to assert that check_callback_completes_event is actually called after client.connect returns.  That sort-of works around the issue.  


Another way, which avoids all the Event patching gymnastics, would be like this:
```
call_complete = False
def t():
    client.connect()
    call_complete=True
threading.Thread(t).start()

assert not call_complete
cb = pipeline.connect.call_args[1]["callback"]
cb()
assert call_complete
```

Plus, I think your event_init_mock is mocking _all_ Event() objects and not just the one you want to mock.

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -52,69 +73,126 @@ class ClientSharedTests(object):
...
+class SharedClientDisconnectTests(object):
+    @pytest.mark.it("Begins a 'disconnect' pipeline operation")
+    async def test_calls_pipeline_disconnect(self, client, pipeline):
+        await client.disconnect()
+        assert pipeline.disconnect.call_count == 1
+
+    @pytest.mark.it(
+        "Waits for the completion of the 'disconnect' pipeline operation before returning"
+    )
+    async def test_waits_for_pipeline_op_completion(self, mocker, client, pipeline):
```
BertKleewein: >test_waits_for_pipeline_op_completion [](start = 14, length = 37)

This pattern is less convoluted than the sync version, and it looks like it asserts everything correctly, but it's still a very roundabout way to do this.  You could do a similar pattern as the sync code and avoid the patching.

```
future = client.disconnect()
assert not future.done()
manually_trigger_callback()
assert future.done()
```

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client_factory.py
cartertinney: Yup. Thanks.

---
In reply to: [290980777](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290980777) [](ancestors = 290980777)

BertKleewein:
> # Copyright (c) Microsoft Corporation. All rights reserved.

My gut  feel on this file and test_async_clients.py is that they have too many different things in them.  It feels to me like there should be 3 .py files: 1 for shared classes, one for device tests, and one for module tests.  I'm saying this because I got confused at least once when I thought I was reading device tests but I was actually reading shared tests.   That's just my 2 cents, but that's in US currency, so it's closer to 2.67 cents for you right now.

---
Refers to: azure-iot-device/tests/iothub/test_sync_clients.py:2 in 42839ff. [](commit_id = 42839ff42145a4df81ee8c61c515c6b5a66afa65, deletion_comment = False)

cartertinney:
> # Copyright (c) Microsoft Corporation. All rights reserved.

The problem is that the shared classes can't be shared between sync and async, because the async ones use async syntax, and furthermore, any async code needs to be quarantined off in an aio directory to prevent it from being collected when pytest runs in Python 2.7/3.4

I agree that the situation isn't ideal (and I'm thinking about adding some enforcement somehow to make sure that sync and async tests have parity), but I don't see a better solution.

---
In reply to: [499629578](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#issuecomment-499629578) [](ancestors = 499629578)

---
Refers to: azure-iot-device/tests/iothub/test_sync_clients.py:2 in 42839ff. [](commit_id = 42839ff42145a4df81ee8c61c515c6b5a66afa65, deletion_comment = False)

BertKleewein:
Final comment: signing off because the pattern is good and the new testdox content is good.  I really don't like the tests that mock Event and AwaitableCallback, but I would block commit for that.

azure-iot-device/tests/common/test_connection_string.py
cartertinney: Since the word "return" has certain implications, I decided to go with

"Uses the input connection string as a string representation"

I don't need to specify "for an instance", I think that's implied

---
In reply to: [290988868](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290988868) [](ancestors = 290988868,290982473)

azure-iot-device/tests/iothub/conftest.py
cartertinney: I think this is a good idea, but it triggers a warning when I attempt to do it right now, so let's punt this down the road.

---
In reply to: [290983779](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290983779) [](ancestors = 290983779)

azure-iot-device/tests/iothub/aio/test_async_inbox.py
cartertinney: Well, actually, it is a verb - "be" - the most forgotten verb. But I do agree the phrasing is SUPER awkward.

My frustration here (and in similar tests) is that this doesn't really feel like it should be a separate unit - it's validating a feature of the object itself, that relates to the properties of the object, and thus doesn't feel like an separate operation/unit. The fact that facts about the unit can be exposed is a requirement of the object IMO.

I think I'm missing a good word or two that might help. I'll revisit this in the morning with a refreshed mind.


---
In reply to: [291185176](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r291185176) [](ancestors = 291185176)

azure-iot-device/tests/common/test_async_adapter.py
cartertinney: Like similarly other awkwardly worded stuff, I'm going to re-examine these with a refreshed mind.

---
In reply to: [290982038](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r290982038) [](ancestors = 290982038)

azure-iot-device/tests/iothub/test_sync_clients.py
cartertinney: The key thing that this test does is show that the callback completes the Event (because Event.set has NOT been called until the callback is triggered). My test runs a side effect when the .wait() is triggered that A) verifies that .wait() has been called (kind of redundant, but makes it more readable) but .set() has NOT been called, B) manually triggers the callback, and C) then shows that after the callback .set() HAS been called. 

This proves the functionality because we know for a fact how Event works, and what .wait() and .set() do. If .wait() has been triggered, we know that it blocks until .set() is also triggered. So we know .wait() triggers (thus blocks by implication), and we know .set() is only triggered as a result of the callback.

Admittedly, as you point out, I'm mocking all Event objects, because there's no way to mock this specific Event that is used in connect since it's created and used only within the function. This would normally be okay, but the limitation is that, yeah, there's no way to prove that the .set() call is happening on the same object that called .wait(). I could add check to the number of calls to the Event constructor though - that would prove that only one Event() object exists, and thus .wait() and .set() are both called on the same instance.

Your test doesn't show that it is ONLY by completion of the callback that the call to connect is able to complete. It's close, but because of the Threading model used, there's no way to be 100% certain that `assert not call_complete` doesn't evaluate to `True` due to timing of the threads - `client.connect()` may just not have finished running yet. I could add a time.sleep(1) in there, but that still wouldn't guarantee it since from a black box perspective, I don't know how long .connect() takes to complete. I don't know that the .connect() call is explicitly blocking on the completion of the callback.

---
In reply to: [291325038](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r291325038) [](ancestors = 291325038)

azure-iot-device/tests/iothub/aio/test_async_clients.py
cartertinney: No, that won't work. As mentioned in the other comment on the sync implementation, this doesn't prove that the completion happens as a direct result of the callback as opposed to just timing issues.

---
In reply to: [291329756](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r291329756) [](ancestors = 291329756)

azure-iot-device/tests/iothub/test_sync_clients.py
cartertinney: This is actually a very nice pattern that I will keep in mind, but I think I'll pass on using it here for the following reasons:

1) I prefer to avoid accessing convention-private attributes when possible. It's not a hard and fast rule, because it does have to happen (fairly frequently too...), but I like to avoid it where I can.
2) I like the symmetry between `pipeline_manual_cb` and `client_manual_cb`. If we were to patch the `client._transport` attribute to be `pipeline_manual_cb`, within `make_client_manual`, then the relationship between `client` and `pipeline_manual_cb` is obfuscated within the test.
3) "Well Carter, why even have `pipeline_manual_cb` when all it is, is a plain mocker.MagicMock()?" Okay, fair, I could get rid of that fixture altogether, but then how do we access the pipeline? "Through the client of course, `client._pipeline`" Okay, fair, but that's another access to a convention-private attribute. "Come on, really? You could remove TWO WHOLE FIXTURES BY DOING IT" Okay, fair, that's a... really good point... BUT... now the tests lose symmetry with all the other tests that import client and pipeline, and it becomes harder to parse why and how this test is different from the others.

Building off that last point/imagined rubber duck discussion with you, I think it would have been beneficial to consider this pattern earlier on in the overall construction of these tests, but without overhauling the fixture model used here, I think the way it currently is slightly gets the edge. I'll definitely keep it mind for future test patterns though, it's very nice.

---
In reply to: [291192784](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#discussion_r291192784) [](ancestors = 291192784)

cartertinney:
I hope that means "not block commit"... that's a pretty big difference in meaning. I'm going to take it as such since you did sign off, and move ahead with the commit.

---
In reply to: [499630778](https://github.com/Azure/azure-iot-sdk-python-preview/pull/101#issuecomment-499630778) [](ancestors = 499630778)

PR #102, opened 2019-06-04
Change all blanket catch statements to only catch Exception

BertKleewein:
Code change was easy.  Test changes were much deeper.


azure-iot-device/tests/common/test_mqtt_provider.py
```diff
@@ -206,6 +210,20 @@ def test_event_handler_callback_raises_exception(self, mocker, mock_mqtt_client,
...
+        event_cb = mocker.MagicMock(side_effect=UnhandledException)
+        provider.on_mqtt_connected = event_cb
+
+        provider.connect(fake_password)
+        with pytest.raises(UnhandledException):
+            mock_mqtt_client.on_connect(
+                client=mock_mqtt_client, userdata=None, flags=None, rc=fake_rc
+            )
+
+        # Callback was called, but exception did not propagate
```
cartertinney: Remove this comment, it's copypasta that is not correct in the new test.

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning_mqtt.py
```diff
@@ -162,14 +163,24 @@ def test_sets_username(self, mock_stage, set_security_client_args):
...
+        self, mock_stage, mocker, fake_exception, set_security_client_args
     ):
-        mock_stage.next._run_op = mocker.Mock(side_effect=fake_error)
+        mock_stage.next._run_op = mocker.Mock(side_effect=fake_exception)
         mock_stage.run_op(set_security_client_args)
-        assert_callback_failed(op=set_security_client_args, error=fake_error)
+        assert_callback_failed(op=set_security_client_args, error=fake_exception)
+
+    @pytest.mark.it(
+        "does not handle any BaseExceptions raised inside the SetConnectionArgs operation"
```
cartertinney: >"does not handle  [](start = 8, length = 17)

Here and everywhere, I might phrase this as "Allows BaseException to propagate" instead of "does not handle BaseException"

PR #104, opened 2019-06-06
Update pipeline op & event tests to follow new pattern

BertKleewein:
This has a few related changes:
1. Moved pipeline_test_fixtures.py into the pipeline folder and renamed.  Changed how these fixtures got included in tests.
2. Moved and renamed pipeline_test.py
3. Wrote helpers to make pipeline_ops_* and pipeline_events_* tests consistent.  Found a few bugs along the way.  Now, all tests for these data objects look like this:
```
DisableFeature - Instantiation
 [x] Accepts 1 positional arguments that get assigned to attributes of the same name: feature_name
 [x] Accepts the following keyword arguments that get assigned to attributes of the same name: feature_name, callback
 [x] Has the following default attributes: needs_connection=True, error=None, name='DisableFeature', callback=None
```
4. While doing ##3, I discovered that we were inconsistently mixing the names "rid" and "request_id", and this was causing confusion, especially when we had self.rid = request_id or self.request_id = rid.  I changed all variable names to be `request_id`.  The only time `rid` is used now is inside topic names (e.g. topics with `$rid={request_id}` in them)


azure-iot-device/tests/common/pipeline/fixtures.py
```diff
@@ -20,7 +20,7 @@ def fake_exception():
 
 @pytest.fixture
 def fake_base_exception():
-    return pipeline_test.UnhandledException()
+    return helpers.UnhandledException()
```
cartertinney: Good change

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
```diff
@@ -0,0 +1,134 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import pytest
+
+fake_count = 0
```
cartertinney: Does this ever get reset to 0?

I see it gets incremented in `get_next_fake_value()`, but I don't see the reset to 0

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
```diff
@@ -0,0 +1,134 @@
...
+    kwargs = {}
+    for key in positional_arguments:
+        kwargs[key] = get_next_fake_value()
+    for key in keyword_arguments:
+        kwargs[key] = get_next_fake_value()
+
+    # LocalTestObject is a local class which tests the object that was passed in.  pytest doesn't test
+    # against this local object, but it does test against it when we put it into the module namespace
+    # for the module that was passed in.
+    @pytest.mark.describe("{} - Instantiation".format(obj.__name__))
```
cartertinney: >@pytest.mark.describe("{} - Instantiation".format(obj.__name__)) [](start = 4, length = 64)

This is clever. I could probably have done this in the client tests.

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
```diff
@@ -0,0 +1,134 @@
...
+    class LocalTestObject(object):
+        @pytest.mark.it(
+            "Accepts {} positional arguments that get assigned to attributes of the same name: {}".format(
```
cartertinney: Do ALL positional arguments get assigned to attributes? Will this ALWAYS be the case?

azure-iot-device/tests/common/pipeline/conftest.py
```diff
@@ -0,0 +1,7 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+pytest_plugins = ["tests.common.pipeline.fixtures"]
```
cartertinney: This is throwing warnings for me.

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
cartertinney: Also, this renders weird for objects with no positional arguments:

`Accepts 0 positional arguments that get assigned to attributes of the same name:`

azure-iot-device/tests/common/pipeline/test_pipeline_events_base.py
```diff
@@ -3,13 +3,21 @@
...
+this_module = sys.modules[__name__]
 
-@pytest.mark.describe("PipelineEvent object")
-class TestPipelineEvent(object):
-    @pytest.mark.it("Sets name attribute on instantiation")
-    def test_default_arguments(self):
-        obj = pipeline_events_base.PipelineEvent()
-        assert obj.name is obj.__class__.__name__
+pipeline_data_object_test.add_event_test(
+    obj=pipeline_events_base.PipelineEvent,
```
cartertinney: Shouldn't this be an AbstractBaseClass, and thus, not directly under test?

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
cartertinney: I would argue that since these are the only tests for the Event/Ops, that you don't even need to specify "Instantiation" here.

cartertinney:
>     [pipeline_ops_base.Connect, []],

Unclear what the second element of these lists represents/is used for. Perusing the way these lists are used, I can't seem to find a use of the second element.

---
Refers to: azure-iot-device/tests/common/pipeline/helpers.py:17 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

cartertinney:
> @pytest.mark.describe("PipelineStage run_ops_serial function")

1) This is not a function, it's a method
2) Given that it isn't anywhere in this PR, I'm assuming you aren't yet converting to the `<classname> - .<methodname>()` syntax yet?
3) Given that you have to structure tests a bit differently (different configs, lead to different behaviors), we should find some way to bring that into the standardized syntax.

I say leave the syntax alone (for now) until the standardized syntax is A) documented B) extended to support your needs

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:92 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

cartertinney:
> class TestPipelineStageRunOpsSerial(object):

Again, I thought this was an AbstractBaseClass though?

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:93 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

cartertinney:
Some of these test directories had `__init__.py` files added - why?

azure-iot-device/tests/iothub/pipeline/conftest.py
```diff
@@ -0,0 +1,7 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+pytest_plugins = ["tests.common.pipeline.fixtures"]
```
cartertinney: As mentioned before, this throws warnings

cartertinney:
>     [pipeline_ops_iothub.SetAuthProvider, [None]],

Again, unclear on what this second element of the lists is, and why it is needed

---
Refers to: azure-iot-device/tests/iothub/pipeline/helpers.py:9 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

BertKleewein:
When I was running tests under 2.7, it was failing to import modules from those directories and the empty __init__.py fixed that.  

---
In reply to: [499969634](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499969634) [](ancestors = 499969634)

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
BertKleewein: Doesn't reset.  Doesn't need to reset.  I just wanted a unique auto-generated string that was human-readable.  

---
In reply to: [291658380](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291658380) [](ancestors = 291658380)

cartertinney:
>     @pytest.mark.it("Sets pipeline_root attribute to None on instantiation")

Still has stacked .its here

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:70 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

cartertinney:
>     @pytest.mark.it("calls _run_op")

Is this a meaningful requirement?

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:78 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
BertKleewein: Yes, for events and ops, this is always the case, and it should be so by definition since they're data objects and nothing more.

---
In reply to: [291664240](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291664240) [](ancestors = 291664240)

BertKleewein:
> class TestPipelineStageRunOpsSerial(object):

Yes, that's still on my queue of things to change.  Hold your horses, I'm getting to it eventaully.

---
In reply to: [499969222](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499969222) [](ancestors = 499969222)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:93 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

BertKleewein:
> @pytest.mark.describe("PipelineStage run_ops_serial function")

Yes, all true.  I'm batching my UT changes and I haven't gotten here yet.

---
In reply to: [499969089](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499969089) [](ancestors = 499969089)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:92 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

BertKleewein:
>     @pytest.mark.it("Sets pipeline_root attribute to None on instantiation")

yes, I'm not fixing everything in this one commit.

---
In reply to: [499974622](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499974622) [](ancestors = 499974622)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:70 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

BertKleewein:
>     @pytest.mark.it("calls _run_op")

Good point.  Still haven't had a chance to refactor these tests yet.

---
In reply to: [499974709](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499974709) [](ancestors = 499974709)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:78 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

BertKleewein:
>     [pipeline_ops_iothub.SetAuthProvider, [None]],

This is the args array for instantiation.  I've change my @pytest.param patterns since I wrote this and it will be more explicit once I get to revisit the rest of the pipeline tests.  

---
In reply to: [499970266](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499970266) [](ancestors = 499970266)

---
Refers to: azure-iot-device/tests/iothub/pipeline_test.py:9 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

BertKleewein:
>     [pipeline_ops_base.Connect, []],

It's an arg list for instantiation.  I will be revisiting this once I get to the stage tests.  We're not there yet. 

---
In reply to: [499966551](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499966551) [](ancestors = 499966551)

---
Refers to: azure-iot-device/tests/common/pipeline_test.py:17 in d50542c. [](commit_id = d50542c33800051ecfa8ba781a04f7dcba30867f, deletion_comment = False)

azure-iot-device/tests/iothub/pipeline/conftest.py
BertKleewein: OK.  Will change.

---
In reply to: [291683903](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291683903) [](ancestors = 291683903)

azure-iot-device/tests/common/pipeline/test_pipeline_events_base.py
BertKleewein: Good point.  Will change.

---
In reply to: [291674656](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291674656) [](ancestors = 291674656)

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
BertKleewein: It happens when testing other objects.  The "Connect" operation, for example, accepts no arguments, as does Disconnect, Reconnect, etc.  I will fix the text for those cases and make the base classes abstract.

---
In reply to: [291673510](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291673510) [](ancestors = 291673510)

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
BertKleewein: Fair enough.  Does that change the phrasing of the @it clauses?  Instantiation accepts positional arguments, but is it correct to say the object accepts positional arguments?  I'm having trouble with the phrasing here.  

---
In reply to: [291679669](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291679669) [](ancestors = 291679669)

azure-iot-device/tests/common/pipeline/conftest.py
BertKleewein: I think you're right.  I will change this.  

---
In reply to: [291668287](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291668287) [](ancestors = 291668287)

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
BertKleewein: Actually, never mind.  I'm leaving this because I have some code-reuse in mind that works better if we leave this.

---
In reply to: [291699321](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291699321) [](ancestors = 291699321,291679669)

azure-iot-device/tests/common/pipeline/test_pipeline_ops_base.py
```diff
@@ -3,124 +3,47 @@
...
-        obj = pipeline_ops_base.DisableFeature(
-            feature_name=fake_feature_name, callback=fake_callback
-        )
-        assert obj.callback is fake_callback
+from tests.common.pipeline import pipeline_data_object_test
+
+this_module = sys.modules[__name__]
+
+pipeline_data_object_test.add_operation_test(
+    obj=pipeline_ops_base.PipelineOperation,
```
olivakar: >PipelineOperation [](start = 26, length = 17)

This one should be abstract right so directly not under test?

BertKleewein: Yes.  Will fix.

---
In reply to: [291707448](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291707448) [](ancestors = 291707448)

azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py
cartertinney: Okay cool. 

---
In reply to: [291705442](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#discussion_r291705442) [](ancestors = 291705442,291699321,291679669)

cartertinney:
Hm, okay.

---
In reply to: [499973088](https://github.com/Azure/azure-iot-sdk-python-preview/pull/104#issuecomment-499973088) [](ancestors = 499973088,499969634)

PR #106, opened 2019-06-07
more test refactoring

BertKleewein:
This is my final round of pipeline test refactoring to conform to standards.  There are still many manu improvements I could make, but I have to finish sometime.
* Change helper methods in PipelineStage into helper functions and moved them into operation_flow.py
* Make PipelineEvent and PipelineOperation into ABCs
* Change code to only test concrete classes and never test ABCs.
* Added helper functions to test base functionality for stages
* Change @describe and @it to conform to current standards
* When I have multiple test classes that test the same function or method, I introduced this convention with a double-dash separating the method/function name from the condition which defines the separate test class: `@pytest.mark.describe("Provider - .run_op() -- called with SetConnectionArgs")`
* Removed stacked `@it`s


azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -0,0 +1,243 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
cartertinney: Curious why this became a separate module, given that the first param of ever function contained within is `stage`. Is it just for ease of testing? I'd be a bit concerned if we had to restructure code for test purposes.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_events_base.py
```diff
@@ -22,6 +25,7 @@ class PipelineEvent(object):
     :type name: str
     """
 
+    @abc.abstractmethod
```
cartertinney: >abstractmethod [](start = 9, length = 14)

Is this abstract? It has an implementation...

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -30,6 +33,7 @@ class PipelineOperation(object):
     :type error: Error
     """
 
+    @abc.abstractmethod
     def __init__(self, callback=None):
```
cartertinney: Again, I don't see how this is abstract.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -103,8 +106,10 @@ def _run_op(self, op):
                     request_id=op.request_id,
                 )
                 payload = json.dumps(op.request_body)
-                self.continue_with_different_op(
-                    original_op=op, new_op=pipeline_ops_mqtt.Publish(topic=topic, payload=payload)
+                operation_flow.delegate_to_different_op(
+                    self,
```
cartertinney: here and everywhere - more than 3 args -> use kwargs. Yes, even for self. `stage=self`

azure-iot-device/tests/common/pipeline/fixtures.py
```diff
@@ -23,8 +23,51 @@ def fake_base_exception():
     return helpers.UnhandledException()
 
 
+class MockEvent(pipeline_events_base.PipelineEvent):
```
cartertinney: Is this really a `mock`? It seems like it's just a subclass of an abstract class, for test purposes, with none of the expected functionality of a Mock() or MagicMock()

azure-iot-device/tests/common/pipeline/fixtures.py
```diff
@@ -23,8 +23,51 @@ def fake_base_exception():
...
+
 @pytest.fixture
 def event():
-    ev = pipeline_events_base.PipelineEvent()
-    ev.name = "test event"
-    return ev
+    return MockEvent()
+
+
+class MockOperation(pipeline_ops_base.PipelineOperation):
```
cartertinney: Ditto here

cartertinney:
> # license information.

Should the module name also be changed to `pipeline_object_instantiation_tests` or something, since the function name changed?

I don't have a strong opinion about this, just wanted to make sure it wasn't forgotten in the refactor

---
Refers to: azure-iot-device/tests/common/pipeline/pipeline_data_object_test.py:4 in 2fb62fb. [](commit_id = 2fb62fbd532644db20b1b275bcafc122828f6e5a, deletion_comment = False)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+    add_unknown_ops_tests(cls=cls, module=module, all_ops=all_ops, handled_ops=handled_ops)
+    add_unknown_events_tests(
+        cls=cls, module=module, all_events=all_events, handled_events=handled_events
+    )
+
+
+def add_unknown_ops_tests(cls, module, all_ops, handled_ops):
+    """
+    Add tests for properly handling of "unkonwn operations," which are operations that aren't
```
cartertinney: >unkonwn [](start = 40, length = 7)

typo

cartertinney:
> class UseSkAuthProvider(PipelineStage):

Not related to this PR, but inspired by looking at the test output - I wonder if these Stage-derived classnames should themselves have "Stage" in them, i.e. `UseSkAuthProviderStage`. Not super important right now, and in fact, I'd probably do it in a separate PR, but I **was** caught off guard by tests with names like `Provider - Instantiation`

---
Refers to: azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py:12 in 2fb62fb. [](commit_id = 2fb62fbd532644db20b1b275bcafc122828f6e5a, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -0,0 +1,243 @@
...
+      operation executed will be passed as the parameter to this function.
+    """
+    worklist = queue.Queue()
+    for arg in args:
+        worklist.put_nowait(arg)
+    finally_op = kwargs.get("finally_op", None)
+    callback = kwargs.get("callback", None)
+
+    if len([x for x in kwargs.keys() if x not in ["finally_op", "callback"]]):
+        raise TypeError("finally_op and callback are the only allowed keyword args")
```
cartertinney: >TypeError [](start = 14, length = 9)

Is this kind of kwarg-checking standard practice? I'm not honestly sure.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -0,0 +1,167 @@
...
+            op = make_mock_op_or_event(op_cls)
+            op.callback = callback
+            op.action = "pend"
+            return op
+
+        @pytest.fixture
+        def stage(self, mocker):
+            return make_mock_stage(mocker=mocker, stage_to_make=cls)
+
+        @pytest.mark.it("Passes unknown operation to next stage")
```
cartertinney: I'm not quite clear on the parametrization here. Output looks something like

```CoordinateRequestAndResponse - .run_op() -- unknown and unhandled operations
[x] Passes unknown operation to next stage[Connect]
...```

In this example, `Connect` would be the next stage right?

Why do we need to test this differently for each possible type of next stage? Shouldn't the next stage be somewhat of a black box from the perspective of the unit under test, in this case, the `CoordinateRequestAndResponse` stage?

I know we briefly talked about this, and it made sense to me at the time, but I'm a little confused looking at it now.

File: azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py [hunk 10]
```diff
@@ -181,14 +144,14 @@ def op_set_sas_token(callback):
     return pipeline_ops_base.SetSasToken(sas_token=fake_sas_token, callback=callback)
 
 
-@pytest.mark.describe("MQTT Provider pipeline stage _runOp function with SetSasToken operations")
+@pytest.mark.describe("Provider - .run_op() -- called with SetSasToken")
```
cartertinney: In a previous comment I mentioned that maybe stages/operations/etc. should have what they are, in their classname so it's clearer to understand what they are when deprived of context.

But even if we don't go down that route, it might be nice to modify some of these test names to be clearer - It took me a while to figure out what this test was for.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -337,16 +304,20 @@ def test_provider_function_throws_base_exception(
 
 
 @pytest.mark.parametrize("params", connection_ops)
-@pytest.mark.describe("MQTT Provider pipeline stage _runOp function with Connect operations")
+@pytest.mark.describe(
+    "Provider - .run_op() -- called with op that connects, disconnects, or reconnects"
```
cartertinney: >called with op that connects, disconnects, or reconnects [](start = 29, length = 56)

Not a problem with this test - but I do wonder if the shared handler for connect/disconnect is something we should try to get rid of at some point. I've been less than thrilled with it from the client layer. I'm guessing it's also kind of weird to work with farther down the stack, given the way this TestClass has to be awkwardly grouped around the shared functionality. Is there any good reason we're doing it this way still?

This comment inspired by this PR, but obviously not relevant to it.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -298,36 +261,40 @@ def provider_function_throws_base_exception(params, stage, mocker, fake_base_exc
 
 @pytest.mark.parametrize("params", connection_ops + pubsub_ops)
 @pytest.mark.describe(
-    "MQTT Provider pipeline stage _run_op with ops that map directly to provider libraray calls"
+    "Provider - .run_op() -- called with op that maps directly to protocol client library calls"
 )
 class TestMQTTProviderBasicFunctionality(object):
-    @pytest.mark.it("calls the appropriate function on the provider library")
+    @pytest.mark.it("Calls the appropriate function on the protocol client library")
```
cartertinney: >Calls the appropriate function on the protocol client library [](start = 21, length = 61)

Not sure about a lot of the phrasing in this test class.

- Not really calling a function, but calling a method
- It's not really calling the protocol client library. That would be Paho, right?. This is calling into the MQTTProvider class. I don't know what to call that though tbh.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -298,36 +261,40 @@ def provider_function_throws_base_exception(params, stage, mocker, fake_base_exc
...
     def test_calls_provider_function(self, stage, create_provider, params, op):
         stage.run_op(op)
         assert getattr(stage.provider, params["provider_function"]).call_count == 1
 
-    @pytest.mark.it("passes the correct args to the provider function")
+    @pytest.mark.it("Passes the correct args to the protocol client library function")
     def test_passes_correct_args_to_provider_function(self, stage, create_provider, params, op):
```
cartertinney: Not sure I think this should be a separate test. I think passing the correct args is implied by testing the calling, right? I'd personally combine this test with the one above.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -388,33 +361,39 @@ def test_provider_function_throws_exception_2(
         assert handler_before == handler_after
 
 
-@pytest.mark.describe("MQTT Provider pipeline stage events from transport library")
+@pytest.mark.describe("Provider - .on_mqtt_message-received()")
```
cartertinney: >- [](start = 50, length = 1)

Typo, should be an underscore

cartertinney: So, again, I'm confused by what this is.

It SEEMS to me, after looking into the test, that you're testing what happens when the `MQTTProvider` triggers it's `.on_mqtt_message_received` handler, thus triggering the `Provider` method `_on_message_received`.

- This `.describe` refers to the wrong method name
- This should probably use the `EVENT:` syntax, instead of the method syntax, right? After all, we're testing what happens in response to an external event using handlers, and the handler isn't the unit under test because it's convention-private. The unit under test is the external event, and what happens in response (the convention private `_on_message_received` is implicitly tested by testing the event)
- I keep repeating this, but `Provider` really should be made into `ProviderStage` to reduce confusion
-

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -388,33 +361,39 @@ def test_provider_function_throws_exception_2(
...
-    @pytest.mark.it("fires an IncomingMessage event for each mqtt message received")
+    @pytest.mark.it("Fires an IncomingMessage event for each mqtt message received")
     def test_incoming_message_handler(self, stage, create_provider, mocker):
         stage.provider.on_mqtt_message_received(topic=fake_topic, payload=fake_payload)
         assert stage.previous.handle_pipeline_event.call_count == 1
         call_arg = stage.previous.handle_pipeline_event.call_args[0][0]
         assert isinstance(call_arg, pipeline_events_mqtt.IncomingMessage)
 
-    @pytest.mark.it("passes topic and payload as part of the IncomingMessage event")
+    @pytest.mark.it("Passes topic and payload as part of the IncomingMessage event")
```
cartertinney: >topic [](start = 28, length = 5)

So here and everywhere I guess, see that other comment about checking args passed should probably be implicit in checking the call to the function/method, rather than it's own separate test (unless you think there's a really good reason for it to be separate)

cartertinney:
>             logger.info("{}({}): disconneting".format(self.name, op.name))

typo

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py:114 in 2fb62fb. [](commit_id = 2fb62fbd532644db20b1b275bcafc122828f6e5a, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -276,102 +144,6 @@ def _handle_pipeline_event(self, event):
...
-            )
-            original_op.error = new_op.error
-            self.complete_op(original_op)
-
-        new_op.callback = new_op_complete
-        self.continue_op(new_op)
-
     def on_connected(self):
         """
         Called by lower layers when the transport connects
```
cartertinney: >transport [](start = 40, length = 9)

Outdated

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -276,102 +144,6 @@ def _handle_pipeline_event(self, event):
...
-                    self.name, original_op.name, new_op.name
-                )
-            )
-            original_op.error = new_op.error
-            self.complete_op(original_op)
-
-        new_op.callback = new_op_complete
-        self.continue_op(new_op)
-
     def on_connected(self):
```
cartertinney: >on_connected [](start = 8, length = 12)

Also, why is this on the base Pipeline class anyway?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -388,33 +361,39 @@ def test_provider_function_throws_exception_2(
...
         stage.provider.on_mqtt_message_received(topic=fake_topic, payload=fake_payload)
         call_arg = stage.previous.handle_pipeline_event.call_args[0][0]
         assert call_arg.payload == fake_payload
         assert call_arg.topic == fake_topic
 
+
+@pytest.mark.describe("Provider - .on_mqtt_connected()")
+class TestMQTTProviderOnConnected(object):
     @pytest.mark.it(
-        "calls self.on_connected and passes it up when the transport connected event fires"
```
cartertinney: >self.on_connected [](start = 15, length = 17)

on_connected handler

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -388,33 +361,39 @@ def test_provider_function_throws_exception_2(
...
-    @pytest.mark.it("passes topic and payload as part of the IncomingMessage event")
+    @pytest.mark.it("Passes topic and payload as part of the IncomingMessage event")
     def test_verify_incoming_message_attributes(self, stage, create_provider, mocker):
         stage.provider.on_mqtt_message_received(topic=fake_topic, payload=fake_payload)
         call_arg = stage.previous.handle_pipeline_event.call_args[0][0]
         assert call_arg.payload == fake_payload
         assert call_arg.topic == fake_topic
 
+
+@pytest.mark.describe("Provider - .on_mqtt_connected()")
```
cartertinney: >"Provider - .on_mqtt_connected()" [](start = 22, length = 33)

Similarly odd naming as above

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -388,33 +361,39 @@ def test_provider_function_throws_exception_2(
...
         assert stage.previous.on_connected.call_count == 1
 
+
+@pytest.mark.describe("Provider - .on_mqtt_disconnected()")
+class TestMQTTProviderOnDisconnected(object):
     @pytest.mark.it(
-        "calls self.on_disconnected and passes it up when the transport disconnected event fires"
+        "Calls self.on_disconnected and passes it up when the transport disconnected event fires"
     )
     def test_disconnected_hanlder(self, stage, create_provider, mocker):
```
cartertinney: >hanlder [](start = 26, length = 7)

typo

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub.py
```diff
@@ -196,23 +176,21 @@ def next_run_op(self, op):
...
     def test_calls_get_current_sas_token(self, stage, set_auth_provider):
         stage.run_op(set_auth_provider)
         assert set_auth_provider.auth_provider.get_current_sas_token.call_count == 1
         set_sas_token_op = stage.next._run_op.call_args_list[1][0][0]
         assert set_sas_token_op.sas_token == fake_sas_token
 
-    @pytest.mark.it(
-        "if the SetSasToken operation succeeds, _run_op calls the SetAuthProvider callback with no error"
-    )
+    @pytest.mark.it("Calls the callback with no error if the SetSasToken op succeeds")
```
cartertinney: >no error [](start = 45, length = 8)

What does "with no error" mean?

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
BertKleewein: I pulled these out because they're helper functions that can be used by a stage, but they're not necessarily part of the identity of a stage.  If I kept them inside the PipelineStage object, they would be more appropriately treated as private functions inside the stage.  But, these helpers provide more functionality than any one stage might need.  In that way, keeping these as privates inside PipelineStage would make each stage much more powerful than it needs to be.

Another way to think about it is to ask if every method on PipelineStage (or any stage) has to do with the core functionality of that stage.  A stage can run operations and handle events.  If we take the first helper (run_ops_in_serial), I would argue that running operations in serial is not part of the identity of a stage.  Rather, running ops in serial is more of an external helper function.

---
In reply to: [293418167](https://github.com/Azure/azure-iot-sdk-python-preview/pull/106#discussion_r293418167) [](ancestors = 293418167)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_events_base.py
BertKleewein: discussing in teams

---
In reply to: [293419085](https://github.com/Azure/azure-iot-sdk-python-preview/pull/106#discussion_r293419085) [](ancestors = 293419085)

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
BertKleewein: If you specify kwargs in your method prototype, this check happens automatically.  In this function, we can't specify it in the prototype, so we have to check manually.  Ideally, we would define this function as def run_ops_in_serial(stage, *args, callback=None, finally_op=None), and this works in newer python versions but not in older ones.  If we did prototype this way, we would get an error if the user tried to use an invalid keyword.  I'm just replicating this behavior.

---
In reply to: [293450525](https://github.com/Azure/azure-iot-sdk-python-preview/pull/106#discussion_r293450525) [](ancestors = 293450525)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
BertKleewein: I like your idea of having the word "Stage" in stage types.  I already have Event in event types (except for one).  Adding "Op" also makes sense.  I'll do that in a separate commit.

---
In reply to: [293455914](https://github.com/Azure/azure-iot-sdk-python-preview/pull/106#discussion_r293455914) [](ancestors = 293455914)

olivakar:
>     def __init__(self, callback=None):

You mentioned in the chat that `I did this by marking the initializer as abstract, but calling it anyway.  This works, but it's strange.`. Why did it feel strange ? Did it not work correctly ?

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py:33 in 3b44611. [](commit_id = 3b4461152b26d89c9deaad5dadb94a1f61624ce2, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -38,6 +38,10 @@ def __init__(self, callback=None):
           failed.  The callback function must accept A PipelineOperation object which indicates
           the specific operation which has completed or failed.
         """
+        if self.__class__ == PipelineOperation:
+            raise TypeError(
```
olivakar: >TypeError [](start = 18, length = 9)

I think I am more in favor for the abstract init

PR #110, opened 2019-06-16
Removal of 2-Step IoTHubClient Instantiation

cartertinney:
* Created factory classmethods on IoTHubClient classes to instantiate
* Removed the old 'create_from_authentication_provider' method
* Refactored IoTEdge-related authentication code
* Added error handling logic to IoTEdge authentication
* Enhanced testing for IoTEdgeAuthenticationProvider and IoTEdgeHsm classes
* Added tests for new IoTHubClient instantiation methods
* Fixed a minor bug in unrelated sync_client tests
* Added beginnings of a test structure to span AuthenticationProvider class family

NOTE: Auth provider package still needs a broad refactor to both code and tests. I tried to keep this PR to as narrow a scope as possible, however as a result, some test coverage for inherited AuthenticationProvider-related methods is lacking.


cartertinney:
>         self.ca_cert = None

What's the deal with this field?

I see it gets used in E2E tests by being set manually (???) when a connection string has a `GatewayHostName`. Can someone explain this to me?

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/sk_authentication_provider.py:67 in 1937d24. [](commit_id = 1937d24a2dc8878fcdb183077ff7a020ae4122a3, deletion_comment = False)

BertKleewein:
>         self.ca_cert = None

It's not a backdoor for testing, but it could be implemented better.  My reasoning in putting this here is that it's an optional attribute required for successful authentication, but it's not part of the connection string.  It should also be set by the edge HSM auth provider.  It does feel awkward, but I'm not sure how to do it better.  Maybe it would work as an optional parameter on the initializer.  (for the record, it should be "trusted_certificate_chain")

---
In reply to: [502478604](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#issuecomment-502478604) [](ancestors = 502478604)

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/sk_authentication_provider.py:67 in 1937d24. [](commit_id = 1937d24a2dc8878fcdb183077ff7a020ae4122a3, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -23,39 +26,40 @@ def __init__(self, pipeline):
 
         :param pipeline: The pipeline that the client will use.
         """
+        # We pass the pipeline through instead of the auth provider to allow
+        # for future support of different protocols - the pipeline will come in
+        # already configured.
         self._pipeline = pipeline
```
BertKleewein: >self._pipeline = pipeline [](start = 8, length = 25)

OK, this answers a question I had on Oliva's review.  Still feels awkward, but I get it.

BertKleewein:
>     def get_current_sas_token(self):

Since you're removing things that don't belong in the abstract base, this doesn't belong here either.  x509 doesn't use sas tokens.

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/authentication_provider.py:33 in fb43701. [](commit_id = fb43701c95ef54435d65badf2ebda8c2f244fa27, deletion_comment = False)

azure-iot-device/tests/iothub/auth/test_iotedge_authentication_provider.py
```diff
@@ -4,61 +4,318 @@
...
+
+
+@pytest.fixture
+def gateway_hostname():
+    return "__FAKE_GATEWAY_HOSTNAME__"
+
+
+@pytest.fixture
+def module_generation_id():
+    return "__FAKE_MODULE_GENERATION_ID__"
```
BertKleewein: Interesting.  Offline, I'd like to hear your opinion of this (using fixtures for simple strings that hold fake values).  Do you think this is a reasonable practice and worth the overhead (if it is indeed overhead)?

azure-iot-device/tests/iothub/auth/test_iotedge_authentication_provider.py
```diff
@@ -4,61 +4,318 @@
...
+    @pytest.mark.it(
+        "Sets a certificate acquired from the IoTEdgeHsm as the ca_cert instance attribute"
+    )
+    def test_ca_cert_from_edge_hsm(self, auth_provider, mock_hsm):
+        assert auth_provider.ca_cert is mock_hsm.get_trust_bundle.return_value
+        assert mock_hsm.get_trust_bundle.call_count == 1
+
+
+# TODO: Potentially get rid of this test class depending on how the parent class is tested/refactored
+@pytest.mark.describe("IoTEdgeAuthenticationProvider - _sign()")
```
BertKleewein: >- _sign()") [](start = 53, length = 11)

missing dot (._sign)?   Or is this a static function?

azure-iot-device/tests/iothub/auth/conftest.py
```diff
@@ -0,0 +1,9 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+from .shared_auth_fixtures import *
+
+# from .shared_auth_tests import *
```
olivakar: >shared_auth_tests [](start = 8, length = 17)

Do we need this ?

azure-iot-device/tests/iothub/auth/test_iotedge_authentication_provider.py
olivakar: For the simple cases like this I feel literal values of string or integer would do.

---
In reply to: [294473417](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#discussion_r294473417) [](ancestors = 294473417)

azure-iot-device/tests/iothub/auth/test_iotedge_authentication_provider.py
cartertinney: Yup, it's missing it. Good catch. Even if it were static, it would still need a dot.

---
In reply to: [294474131](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#discussion_r294474131) [](ancestors = 294474131)

cartertinney:
>         self.ca_cert = None

Since auth providers are no longer directly instantiated by the user, this will mean I need to make it a optional argument on `create_from_connection_string` which is... not ideal, but I also can't think of anything better. With the right custom errors I think we could reduce the confusion - if not in this PR, in the eventual second part of auth revision.

---
In reply to: [502805197](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#issuecomment-502805197) [](ancestors = 502805197,502478604)

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/sk_authentication_provider.py:67 in 1937d24. [](commit_id = 1937d24a2dc8878fcdb183077ff7a020ae4122a3, deletion_comment = False)

cartertinney:
>     def get_current_sas_token(self):

Okay I'll remove this ahead of x509 being added. This whole hierarchy is in need of revision though. I want to revisit this package as soon as possible.

---
In reply to: [502808803](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#issuecomment-502808803) [](ancestors = 502808803)

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/authentication_provider.py:33 in fb43701. [](commit_id = fb43701c95ef54435d65badf2ebda8c2f244fa27, deletion_comment = False)

azure-iot-device/tests/iothub/auth/conftest.py
cartertinney: Nope

---
In reply to: [294488050](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#discussion_r294488050) [](ancestors = 294488050)

azure-iot-device/tests/iothub/auth/test_iotedge_authentication_provider.py
cartertinney: Because these values have to be shared between multiple test modules, not just for various things that all use these values, but also for extensions of inherited methods, I believe fixtures are the best way to do it. Not only are they easily shared between tests modules without imports, but they can fit easily be used across an inheritance hierarchy while keeping the signatures the same. If that makes sense. 

I'll likely have to explain this in a bit more detail with additional clarity later, but this technique is crucial for testing complex object hierarchies.

---
In reply to: [294500681](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#discussion_r294500681) [](ancestors = 294500681,294473417)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: Actually no. Next PR will move away from this due to the need to support multiple pipelines for multiple destinations.

---
In reply to: [294460475](https://github.com/Azure/azure-iot-sdk-python-preview/pull/110#discussion_r294460475) [](ancestors = 294460475)

PR #116, opened 2019-06-26
script

olivakar:
Usage :- 

```
python create_x509_chain.py "leviosa" --ca-password "hogwarts" --intermediate-password "lumos" --device-password "alohomora"
```

### DIFF
--- a/azure-iot-device/tools/create_x509_chain.py
+++ b/azure-iot-device/tools/create_x509_chain.py
@@ -1,0 +1,190 @@
+import os
+import re
+import argparse
+import getpass
+
+
+def create_custom_config():
+    # This is very specific to installations in various system
+    openssl_path = os.getenv("OpenSSLDir")
+    config_path = openssl_path + "/bin/openssl.cfg"
+    with open(config_path, "r") as openssl_config:
+        config = openssl_config.read()
+    lines = config.splitlines()
+    policy_loose_found = False
+    policy_any_found = False
+
+    policy_any_regex = re.compile(r"\s*\[\s*policy_anything\s*\]\s*")
+
+    # First, try to find policy_anything in the openssl config file
+    for number, line in enumerate(lines):
+        if re.search(policy_any_regex, line):
+            policy_any_found = True
+            break
+    # Of not found the try the search with policy_loose
+    if not policy_any_found:
+        policy_loose_regex = re.compile(r"\s*\[\s*policy_loose\s*\]\s*")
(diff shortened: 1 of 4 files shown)


azure-iot-device/tools/create_x509_chain.py
```diff
@@ -0,0 +1,190 @@
+import os
+import re
+import argparse
```
cartertinney: Put in the scripts folder

azure-iot-device/tools/create_x509_chain.py
```diff
@@ -0,0 +1,190 @@
...
+        for number, line in enumerate(lines):
+            if re.search(policy_loose_regex, line):
+                policy_loose_found = True
+                break
+
+    list_of_lines = list()
+    ca_default_regex = re.compile(r"\s*\[\s*CA_default\s*\]\s*")
+    ca_default_section_found = False
+    policy_regex = re.compile(r"\s*\s*(policy\s*=\s*policy_strict|policy\s*=\s*policy_match)")
+    with open("C:/OpenSSL-Win64/bin/openssl.cfg", "r") as change_openssl_config:
```
cartertinney: >C:/OpenSSL-Win64/bin/openssl.cfg [](start = 15, length = 32)

Only valid on Windows, and with 64 bit at that.

cartertinney:
If you can confirm that it works in both Python 2.7 and 3.5, and ensure that it works on multiple platforms (Windows, OSX, Linux), I'll sign off.

azure-iot-device/tools/create_x509_chain.py
olivakar: changed to environment variable like the previous one. 

---
In reply to: [297798367](https://github.com/Azure/azure-iot-sdk-python-preview/pull/116#discussion_r297798367) [](ancestors = 297798367)

olivakar:
Confirmed that works on Python 2..
For mac and linux I have to do a little bit more time. 


---
In reply to: [505985870](https://github.com/Azure/azure-iot-sdk-python-preview/pull/116#issuecomment-505985870) [](ancestors = 505985870)

pierreca:
Ha I see why the config file was a problem: you went with invoking the openssl binary directly.
There might be a way to greatly simplify the code (especially all that stuff about dealing with the config file) by using an existing library
have you looked at this for example: [https://cryptography.io/en/latest/x509/](https://cryptography.io/en/latest/x509/)

PR #118, opened 2019-06-28
Add yaml file to build public sdk release artifacts

fratster:


build-release.yml
```diff
@@ -0,0 +1,53 @@
...
+  clean: 'resources'
+  
+- task: UsePythonVersion@0
+  displayName: 'Use Python 3.x'
+
+- task: PowerShell@2
+  inputs:
+    targetType: 'inline'
+    script: |
+      . $(System.ArtifactsDirectory)/scripts/python/azure-iot-sdk-bump-version.ps1
```
pierreca: >/scripts/python/azure-iot-sdk-bump-version.ps1 [](start = 36, length = 46)

missing this file?

cartertinney:
Can we put the file in the existing "vsts" folder rather than a new one?

build-release.yml
fratster: This file lives in Microsoft internal repo but is called from this public yaml file.

PR #128, opened 2019-07-05
module id for x509

olivakar:


azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -67,6 +67,24 @@ def create_from_shared_access_signature(cls, sas_token):
         pipeline = IoTHubPipeline(authentication_provider)
         return cls(pipeline)
 
+    @classmethod
+    def create_from_x509_certificate(cls, hostname, x509, device_id, module_id=None):
```
cartertinney: >x509 [](start = 52, length = 4)

I think this should probably be the first argument

azure-iot-device/azure/iot/device/iothub/auth/x509_authentication_provider.py
```diff
@@ -17,14 +17,16 @@ class X509AuthenticationProvider(AuthenticationProvider):
...
         Constructor for X509 Authentication Provider
         :param hostname: The hostname of the Azure IoT hub.
-        :param device_id: The device unique identifier as it exists in the Azure IoT Hub device registry.
         :param x509: The X509 object containing certificate, key and passphrase
+        :param device_id: The device unique identifier as it exists in the Azure IoT Hub device registry.
+        :param module_id: The module unique identifier of the device. Ir is optional when dealing with only devices.
         """
-        logger.info("Using X509 authentication for {%s, %s}", hostname, device_id)
+        # logger.info("Using X509 authentication for {%s, %s}", hostname, device_id)
+        logger.info("Using X509 authentication for {hostname},{device_id},{module_id}".format(hostname=hostname, device_id=device_id, module_id=module_id))
```
cartertinney: Wait, why are you passing in the module id if the only change is that now we log it?

Nothing else is even happening with the module id...

cartertinney:
> @pytest.mark.describe("X509AuthenticationProvider")

No tests with module id?

---
Refers to: azure-iot-device/tests/iothub/auth/test_x509_authentication_provider.py:22 in e93f2be. [](commit_id = e93f2be8477827f9dc1c31b05ab65fe0f2e0c9a6, deletion_comment = False)

PR #129, opened 2019-07-08
module id for x509

olivakar:
The samples are not working.

### DIFF
--- a/azure-iot-device/azure/iot/device/iothub/abstract_clients.py
+++ b/azure-iot-device/azure/iot/device/iothub/abstract_clients.py
@@ -67,24 +67,6 @@ class AbstractIoTHubClient(object):
         pipeline = IoTHubPipeline(authentication_provider)
         return cls(pipeline)
 
-    @classmethod
-    def create_from_x509_certificate(cls, hostname, x509, device_id, module_id=None):
-        """
-        Instantiate a client which using X509 certificate authentication.
-        :param hostname: Host running the IotHub. Can be found in the Azure portal in the Overview tab as the string hostname.
-        :param x509: The complete x509 certificate object, To use the certificate the enrollment object needs to contain cert (either the root certificate or one of the intermediate CA certificates).
-        If the cert comes from a CER file, it needs to be base64 encoded.
-        :type x509: X509
-        :param device_id: The ID is used to uniquely identify a device in the IoTHub
-        :param module_id : The ID of the module to uniquely identify a module on a device on the OpTHub
-        :return: A IoTHubClient which can use X509 authentication.
-        """
-        authentication_provider = auth.X509AuthenticationProvider(
-            hostname=hostname, x509=x509, device_id=device_id, module_id=module_id
-        )
-        pipeline = IoTHubPipeline(authentication_provider)
-        return cls(pipeline)
-
     @abc.abstractmethod
     def connect(self):
         pass
@@ -120,6 +102,22 @@ class AbstractIoTHubClient(object):
 
 @six.add_metaclass(abc.ABCMeta)
 class AbstractIoTHubDeviceClient(AbstractIoTHubClient):
+    @classmethod
+    def create_from_x509_certificate(cls, x509, hostname, device_id):
+        """
+        Instantiate a client which using X509 certificate authentication.
+        :param hostname: Host running the IotHub. Can be found in the Azure portal in the Overview tab as the string hostname.
+        :param x509: The complete x509 certificate object, To use the certificate the enrollment object needs to contain cert (either the root certificate or one of the intermediate CA certificates).
+        If the cert comes from a CER file, it needs to be base64 encoded.
+        :type x509: X509
+        :param device_id: The ID is used to uniquely identify a device in the IoTHub
+        :return: A IoTHubClient which can use X509 authentication.
+        """
+        authentication_provider = auth.X509AuthenticationProvider(
+            x509=x509, hostname=hostname, device_id=device_id
+        )
+        pipeline = IoTHubPipeline(authentication_provider)
+        return cls(pipeline)
 
     @abc.abstractmethod
     def receive_c2d_message(self):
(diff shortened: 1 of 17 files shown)


azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -67,6 +67,24 @@ def create_from_shared_access_signature(cls, sas_token):
         pipeline = IoTHubPipeline(authentication_provider)
         return cls(pipeline)
 
+    @classmethod
+    def create_from_x509_certificate(cls, hostname, x509, device_id, module_id=None):
```
cartertinney: Put this argument first? It is "create_from_x509_certificate" after all.

azure-iot-device/azure/iot/device/iothub/auth/x509_authentication_provider.py
```diff
@@ -17,15 +17,16 @@ class X509AuthenticationProvider(AuthenticationProvider):
     for device identities connecting directly to an Azure IoT hub.
     """
 
-    def __init__(self, hostname, device_id, x509):
+    def __init__(self, hostname, x509, device_id, module_id=None):
```
cartertinney: >x509 [](start = 33, length = 4)

Also put the argument first here

azure-iot-device/samples/advanced-hub-scenarios/send_to_output_x509.py
```diff
@@ -0,0 +1,64 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+
+messages_to_send = 10
+
+
+async def main():
+    hostname = os.getenv("HOSTNAME")
+
+    # The device having a certain module that has been created on the portal
```
cartertinney: There's a conversation to be had about what samples we want to actually have, and how we want to structure them - we don't want to drown customers in tons of redundant samples that are subtly different. But for now this is fine.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: This should also be two separate APIs, one for device, and one for module, given that the signatures are different.

azure-iot-device/tests/iothub/auth/test_x509_authentication_provider.py
```diff
@@ -25,22 +26,38 @@ class TestX509AuthenticationProvider(object):
...
 
+    @pytest.mark.it("Instantiates correctly with module_id")
+    def test_instantiates_correctly_with_module_id(self):
+        x509_cert_object = x509()
+        x509_auth_provider = X509AuthenticationProvider(
+            hostname=hostname, x509=x509_cert_object, device_id=device_id, module_id=module_id
+        )
+        assert x509_auth_provider.module_id == module_id
+
+    @pytest.mark.it("Instantiates correctly with optional module_id")
```
cartertinney: >Instantiates correctly with optional module_id [](start = 21, length = 46)

"Module id defaults to None"

cartertinney:
>     @pytest.mark.it("Instantiates correctly with hostname")

Here and everywhere can drop the "correctly".

---
Refers to: azure-iot-device/tests/iothub/auth/test_x509_authentication_provider.py:25 in 1ca2812. [](commit_id = 1ca281254730c4db5e197decc665795aad0964fd, deletion_comment = False)

scripts/create_x509_chain.py
```diff
@@ -12,8 +12,7 @@ def create_custom_config():
     # Best options is to have the location of openssl config file in an env variable
     # The openssl config file extension could be "cfg" or "cnf"
 
-    # config_path = os.getenv("OPENSSLCONFIG")
-    config_path = "C:/OpenSSL-Win64/bin/openssl.cfg"
+    config_path = os.getenv("OPENSSLCONFIG")
```
cartertinney: good idea

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub.py
```diff
@@ -113,9 +113,9 @@ def set_auth_provider(self, callback, params_auth_provider_ops):
     @pytest.fixture
     def set_auth_provider_all_args(self, callback, params_auth_provider_ops):
```
cartertinney: As per our discussion, is this test suite providing correct coverage?

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
BertKleewein: Yes, but not only because the signatures are different.  They are also intended to create different objects -- they should be on AbstractIoTHubModuleClient and AbstractIoTHubDeviceClient.  Otherwise, we're giving implicit permission to add a module ID when creating a device client or skip the module ID when creating a module client.

---
In reply to: [301326329](https://github.com/Azure/azure-iot-sdk-python-preview/pull/129#discussion_r301326329) [](ancestors = 301326329)

azure-iot-device/samples/advanced-hub-scenarios/send_to_output_x509.py
```diff
@@ -0,0 +1,64 @@
...
+    # Connect the client.
+    await module_client.connect()
+
+    async def send_test_message(i):
+        print("sending message #" + str(i))
+        msg = Message("test wind speed " + str(i))
+        msg.message_id = uuid.uuid4()
+        msg.correlation_id = "correlation-1234"
+        msg.custom_properties["tornado-warning"] = "yes"
+        await module_client.send_to_output(msg, "twister")
```
BertKleewein: send_to_output is only "officially supported" for iot edge modules, but iot edge doesn't (as far as I know) support x509 auth, so the premise of this sample may be invalid.

I say not "officially supported" because you _can_ use send_to_output with iot hub, but it arrives at iothub as telemetry with some extra attributes.  (At least you could do this at one point).  This is not documented or supported behavior AFAIK.

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -949,6 +943,36 @@ def mock_open(self, mocker):
...
+    async def test_instantiates_client(self, client_class, x509):
+        client = client_class.create_from_x509_certificate(
+            hostname="durmstranginstitute.farend",
+            x509=x509,
+            device_id="MySnitch",
+            module_id="Charms",
+        )
+        assert isinstance(client, client_class)
+
+    @pytest.mark.it("Uses a X509AuthenticationProvider to create the client's IoTHub pipeline")
```
BertKleewein: >Uses a X5 [](start = 21, length = 9)

grammar nit: Even though X is a consonant, this is pronounced as "ex five oh nine", so it's more correct to say "Uses an X509AuthenticationProvider".

Issue #127, opened 2019-07-03
Unable to run the basic sample: 'azure.iot.device.iothub.auth' has no attribute 'from_connection_string'


ericmitt:
I installed the preview sdk: pip install azure-iot-device //OK
I have created test.py file with the sample code from the readme

When I try to run it, I got this message:

Traceback (most recent call last):
  File "test.py", line 30, in <module>
    asyncio.run(main())
  File "C:\Users\ericmitt\AppData\Local\Programs\Python\Python37\lib\asyncio\runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "C:\Users\ericmitt\AppData\Local\Programs\Python\Python37\lib\asyncio\base_events.py", line 584, in run_until_complete
    return future.result()
  File "test.py", line 12, in main
    auth_provider = auth.from_connection_string(conn_str)
AttributeError: module 'azure.iot.device.iothub.auth' has no attribute 'from_connection_string'

oerpli:
Afaik there has been a change from rc6 to rc7 (the latest release) which is not reflected in the readme

I assume you use the following:
```
    # Create an authentication provider using the connection string
    auth_provider = auth.from_connection_string(conn_str)

    # Create instance of the device client using the authentication provider
    device_client = IoTHubDeviceClient.from_authentication_provider(auth_provider, "mqtt")
```

You should use this instead:

```
    # Create instance of the device client using the connection string
    device_client = IoTHubDeviceClient.create_from_connection_string(conn_str)
```
See [azure-iot-device/samples/simple_send_d2c_message.py](https://github.com/Azure/azure-iot-sdk-python-preview/blob/master/azure-iot-device/samples/simple_send_d2c_message.py#L18)

cartertinney:
Yes, the README is outdated, we'll be fixing that immediately. Thank you for bringing this to our attention. As @oerpli pointed out, the correct API call is now to use class methods of the desired client:

`device_client = IoTHubDeviceClient.create_from_connection_string(conn_str)`

PR #130, opened 2019-07-08
 Updated example in readme (2.0.0rc7 compatibility)

oerpli:
Addresses #127


cartertinney:
We can't currently support external PRs, so I've pulled in these changes (and added more on top of them) in #131

PR #133, opened 2019-07-08
marshal all calls into a single pipeline thread

BertKleewein:
Still need to add unit tests.  Please review as though this is final.


azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+
+1. All pipeline functions execute in a single thread, known as the "pipeline
+  thread".  The `invoke_on_pipeline_thread` and `invoke_on_pipeline_thread_nowait`
+  decorators cause the decorated function to run on the pipeline thread.
+
+2. If the pipeline thread is busy running a different function, the invoke
+  decorators will wait until that function is complete before invoking another
+  function on that thread.
+
+3. There is a different thread which is used for callsbacks into user code, known
```
olivakar: >callsbacks [](start = 49, length = 10)

spelling

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+  the function which entered the pipeline thread.
+
+5. Calls into the pipeline thread can either block or not block.  Blocking is used
+  for cases where the caller needs a return value from the pipeline or is
+  expecting to handle any errors raised from the pipeline thread.  Blocking is
+  not used when the code calling into the pipeline is not waiting for a response
+  and is not expecting to handle any exceptions, such as protocol library
+  handlers which call into the pipeline to deliver protocol messages.
+
+6. Calls into the callback thread could theretically block, but we currently
```
olivakar: >theretically [](start = 40, length = 12)

spell

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+    Get an object which contains "thread local storage" using the threading.local()
+    function.  If no such object exists, one is created.  The object returned from
+    `threading.local()` appears to be a global object, but has a different value for
+    each individual thread.  We use this so we can check which thread we're running
+    inside at runtime.
+    """
+    if not getattr(_get_thread_local_storage, "local", None):
+        logger.info("Creating thread local storage")
+        _get_thread_local_storage.local = threading.local()
+    return _get_thread_local_storage.local
```
cartertinney: >local [](start = 37, length = 5)

Is this pattern of dynamically adding a attribute to the function following any kind of desired convention or pattern? This seems like a hack to me.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _invoke_on_executor_thread(thread_name, block=True, _func=None):
+    """
+    Decorator which runs the wrapped function on a given thread.  If block==False,
+    the call returns immediately without waiting for the decorated function to complete.
+    If block==True, the call waits for the decorated function to complete before returning.
+    """
+
+    def decorator(func):
```
cartertinney: This needs the `@functools.wraps` decorator put on it to carry over docstring, etc.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+        function_name = getattr(func, "__name__", str(func))
+
+        def wrapper(*args, **kwargs):
```
cartertinney: >def wrapper(*args, **kwargs) [](start = 8, length = 28)

This needs the @functools.wraps decorator put on it to carry over function metadata

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+        else:
+            return wrapper
+
+    if _func is None:
+        return decorator
+    else:
+        return decorator(_func)
+
+
+def invoke_on_pipeline_thread(_func=None):
```
olivakar: >_func=None [](start = 30, length = 10)

is the function truly optional ?

olivakar:
>     def _set_topic_names(self, device_id, module_id):

what about this function ?

---
Refers to: azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py:132 in d4c9b74. [](commit_id = d4c9b74d8eb7635ce6fb1499774ed83cc2929243, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+    Decorator which runs the wrapped function on a given thread.  If block==False,
+    the call returns immediately without waiting for the decorated function to complete.
+    If block==True, the call waits for the decorated function to complete before returning.
+    """
+
+    def decorator(func):
+        function_name = getattr(func, "__name__", str(func))
+
+        def wrapper(*args, **kwargs):
+            if not getattr(_get_thread_local_storage(), "in_{}_thread".format(thread_name), False):
```
cartertinney: All these `getattr`s make me nervous, especially when they start mapping to strings to check statuses. `getattr` is one of those language constructs that really isn't supposed to be used except in exceptional circumstances. Could we maybe use some kind of dict for cached values instead?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+def _assert_executor_thread(thread_name, _func=None):
+    """
+    Decorator which asserts that the given function only gets called inside the given
+    thread.
+    """
+
+    def decorator(func):
+        function_name = getattr(func, "__name__", str(func))
+
+        @functools.wraps(func)
```
cartertinney: >@functools.wraps(func) [](start = 8, length = 22)

Is this not going to have the same issue as above?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+        def wrapper(*args, **kwargs):
+            if not getattr(_get_thread_local_storage(), "in_{}_thread".format(thread_name), False):
+                assert False, """Function {function_name} is not running inside {thread_name} thread.
+                    It should be. You should use invoke_on_{function_name}_thread(_nowait) to enter the
+                    {thread_name} thread before calling this function.  If you're hitting this from
+                    inside a test function, you may need to add the fake_pipeline_thread fixture to
+                    your test.  (grep for apply_fake_pipeline_thread).""".format(
+                    function_name=function_name, thread_name=thread_name
+                )
```
cartertinney: I need to research a bit more to see if we can compile out the asserts when we build the packages. If not, its probably preferable to use regular Exceptions/Errors.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+        executor = ThreadPoolExecutor(max_workers=1)
+        setattr(_get_named_executor, thread_name, ThreadPoolExecutor(max_workers=1))
+    return executor
+
+
+def _get_thread_local_storage():
+    """
+    Get an object which contains "thread local storage" using the threading.local()
+    function.  If no such object exists, one is created.  The object returned from
+    `threading.local()` appears to be a global object, but has a different value for
```
olivakar: >appears to be a global object [](start = 24, length = 29)

when you say it appears to be a global object , u mean its globally available ? has it anything to do with globals ?  `local()`  is just a class that represents thread-local data right ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    executor = getattr(_get_named_executor, thread_name, None)
+    if not executor:
+        logger.info("Creating {} executor".format(thread_name))
+        executor = ThreadPoolExecutor(max_workers=1)
+        setattr(_get_named_executor, thread_name, ThreadPoolExecutor(max_workers=1))
+    return executor
+
+
+def _get_thread_local_storage():
```
olivakar: >_get_thread_local_storage [](start = 4, length = 25)

if this just returns thread specific local data is there a reason why it is named `local_storage` ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
olivakar: yes same question i had

---
In reply to: [301796343](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301796343) [](ancestors = 301796343)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+
+def _invoke_on_executor_thread(thread_name, block=True, _func=None):
+    """
+    Decorator which runs the wrapped function on a given thread.  If block==False,
+    the call returns immediately without waiting for the decorated function to complete.
+    If block==True, the call waits for the decorated function to complete before returning.
+    """
+
+    def decorator(func):
+        function_name = getattr(func, "__name__", str(func))
```
olivakar: >function_name [](start = 8, length = 13)

if this function does not have a function name what will this do ?

olivakar:
> resources:

is there any change to this file ?

---
Refers to: vsts/build.yaml:1 in d4c9b74. [](commit_id = d4c9b74d8eb7635ce6fb1499774ed83cc2929243, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,199 @@
...
+
+        return wrapper
+
+    if _func is None:
+        return decorator
+    else:
+        return decorator(_func)
+
+
+def runs_on_pipeline_thread(_func=None):
```
olivakar: >runs_on_pipeline_thread [](start = 4, length = 23)

so this in only for assertion ? in that case could we add an `ensure` or `assert`  before ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: This was unclear. I rewrote it.  it is a single object, but each thread has different contents

---
In reply to: [301801795](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301801795) [](ancestors = 301801795)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: Sort of.  I did this to make a lazy-init singleton and I put it as an attribute to hide the details.   In a different language, I would have used a static.  I see your point -- I'm changing it to a global.

---
In reply to: [301727424](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301727424) [](ancestors = 301727424)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: it defaults to str(func)

---
In reply to: [301843881](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301843881) [](ancestors = 301843881)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: Yes, if you're using this as a decorator, it gets called with _func = None.  If you use it to wrap a function (like foo = invoke_on_pipeline_thread(bar).(()), then _func = bar

---
In reply to: [301759558](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301759558) [](ancestors = 301759558)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: We only need that hack above because we use this decorator to wrap MicroMock objects

---
In reply to: [301843801](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301843801) [](ancestors = 301843801,301796343)

BertKleewein:
> resources:

I think there's a bug in codeflow.  I'm unindenting line 62 by 2 spaces, but it's not showing up here.  (it does show up on [https://github.com/Azure/azure-iot-sdk-python-preview/pull/133/files](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133/files))

---
In reply to: [509858667](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#issuecomment-509858667) [](ancestors = 509858667)

---
Refers to: vsts/build.yaml:1 in d4c9b74. [](commit_id = d4c9b74d8eb7635ce6fb1499774ed83cc2929243, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: "thread local storage" is an industry term.  Regardless, I've removed this in my next iteration.

---
In reply to: [301802099](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301802099) [](ancestors = 301802099)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: `ensure` wouldn't be appropriate because that implies that it might change something.  `ensure_running_on_pipeline_thread` would imply a function that checks the thread and change it if necessary.
Either way, I'm going to leave it like this for now.

---
In reply to: [301844700](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r301844700) [](ancestors = 301844700)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -56,6 +56,7 @@ def _create_mqtt_client(self):
         self._mqtt_client = mqtt.Client(
             client_id=self._client_id, clean_session=False, protocol=mqtt.MQTTv311
         )
+        self._mqtt_client.enable_logger(logging.getLogger("paho"))
```
cartertinney: good touch

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
cartertinney: I thought decorators always took the function they are decorating as the first argument? Am I wrong about this?

---
In reply to: [302245692](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r302245692) [](ancestors = 302245692,301759558)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -26,6 +26,17 @@
...
 this_module = sys.modules[__name__]
 
+
+# This fixture makes it look like all test in this file  tests are running
+# inside the pipeline thread.  Because this is an autouse fixture, we
+# manually add it to the individual test.py files that need it.  If,
+# instead, we had added it to some conftest.py, it would be applied to
+# every tests in every file and we don't want that.
+@pytest.fixture(autouse=True)
+def apply_fake_pipeline_thread(fake_pipeline_thread):
```
cartertinney: I would also want to see tests that show the various methods can ONLY be run in the pipeline thread as well though

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: I was wrong about this.  I was (incorrectly) copying a pattern that I didn't need here.  
(The pattern that I didn't need is here: [https://realpython.com/primer-on-python-decorators/#both-please-but-never-mind-the-bread](https://realpython.com/primer-on-python-decorators/#both-please-but-never-mind-the-bread))

---
In reply to: [302770609](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r302770609) [](ancestors = 302770609,302245692,301759558)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def _invoke_on_executor_thread(func, thread_name, block=True):
+    """
+    Return wrapper to run the function on a given thread.  If block==False,
+    the call returns immediately without waiting for the decorated function to complete.
+    If block==True, the call waits for the decorated function to complete before returning.
+    """
+
+    # Mocks on py27 don't have a __name__ attribute.  Use str() if you can't use __name__
```
cartertinney: >Mocks on py27 don't have a __name__ attribute.  Use str() if you can't use __name__ [](start = 6, length = 83)

I'm not sure if philosophically, test-related compat issues should be handled in the code under test. Couldn't the testing module manually attach a `__name__` attribute in 2.7?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -62,6 +64,7 @@ def _run_op(self, op):
         elif isinstance(op, pipeline_ops_base.ConnectOperation):
             logger.info("{}({}): conneting".format(self.name, op.name))
 
+            @pipeline_thread.invoke_on_pipeline_thread_nowait
```
cartertinney: I'm unclear as to why this has to specifically be invoked on the pipeline thread, when the outer function is clearly running on the pipeline thread (due to the assertion decorator)

cartertinney:
>             self.transport.on_mqtt_connected = on_connected

I'm still not thrilled about the dynamic switching of handlers, but it's well documented, and without digging deep into this, I don't have any better suggestions. Let's just keep an eye on this space.

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py:100 in 66b1602. [](commit_id = 66b16026e42196fc8c5b1854c4995e887fa7f3bd, deletion_comment = False)

azure-iot-device/azure/iot/device/common/unhandled_exceptions.py
```diff
@@ -0,0 +1,27 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
cartertinney: Shouldn't this file be in the pipeline subdirectory?

cartertinney: Also, I don't love that this is it's own module, but I see why it has to be - it's used in multiple places (`operation_flow` and `pipeline_stages_base`). This is an consequence of `operation_flow` being made into it's own module rather than having it's contents implemented as part of` pipelinge_stages_base`, the purpose of which I was never exactly clear on. 

I'm not suggesting any changes to that approach right now, but if we ever revisit and unify, we should probably fold this into the unification as well.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -167,6 +187,81 @@ def test_passes_event_to_previous_stage_which_throws_base_exception(
...
+
+        @pytest.mark.parametrize("method_name", methods_that_assert_pipeline_thread)
+        @pytest.mark.it("Has a method that must be in a pipeline thread")
+        def test_asserts_in_pipeline(self, stage, method_name, fake_non_pipeline_thread):
+            func = getattr(stage, method_name)
+            args = [None for i in (range(get_arg_count(func) - 1))]
+            with pytest.raises(AssertionError):
+                func(*args)
+
+        if methods_that_enter_pipeline_thread:
```
cartertinney: This is wild, I didn't know you could put conditional statements inside a class. Very interesting.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -167,6 +187,81 @@ def test_passes_event_to_previous_stage_which_throws_base_exception(
...
+    ]
+
+    @pytest.mark.describe("{} - Pipeline threading".format(cls.__name__))
+    class LocalTestObject(object):
+        @pytest.fixture
+        def stage(self):
+            return cls()
+
+        @pytest.mark.parametrize("method_name", methods_that_assert_pipeline_thread)
+        @pytest.mark.it("Has a method that must be in a pipeline thread")
```
cartertinney: This is not a great requirement name, but I do see your dilemma - you test pipeline objects as objects, but the thread-running-ness of its methods is kind of within the domain of a method unit. So instead you have to parametrize the pipeline object methods and show that each is called in the correct thread.

Since obviously we don't want to revisit the entire testing approach, maybe a slightly better name? I might suggest `Enforces use of the pipeline thread when calling method`?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
BertKleewein: I completely agree.  I tried this and other parts of the test broke.  I don't remember the details, but it appeared as if someone was using the __name__ attribute to distinguish a mock from a real object.  My ability to experiment was somewhat hampered  by the fact that __name__ is special.  For example, I could set it, but I could not clear it -- the interpreter gave me an error if I tried to del it.  This told me that I was perhaps playing with fire and shouldn't.  

---
In reply to: [305460661](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r305460661) [](ancestors = 305460661)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
BertKleewein: Good question.  This is invoked by paho after the outer function returns.  (side note: I'm revisiting this usage as part of my paho errors PR, so this will change).

---
In reply to: [305463891](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r305463891) [](ancestors = 305463891)

azure-iot-device/azure/iot/device/common/unhandled_exceptions.py
BertKleewein: I'm working on the assumption that exceptions in background threads will be thrown from other places that are not part of the pipeline.  For example, SAS token renewal takes place based on a timer interval, so this will be on a background thread.  If that fails, we need a place to send that exception.  

---
In reply to: [305478394](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r305478394) [](ancestors = 305478394)

BertKleewein: I'm working off the assumption that this functionality will get at least slightly more complex and will get reworked.  This may be a temporary holding point.  I mostly wanted something simple that didn't make any assumptions.  For example, if we're returning these errors to the user somehow, this maybe needs to be exposed by the client object(s), but I'm not ready to take on the work to put it there until we're sure that's the right place  

(as far as operation_flow.py goes, we can discuss offline -- it's really more of a philosophical debate.)

---
In reply to: [305482114](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r305482114) [](ancestors = 305482114)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
BertKleewein: I thought you'd appreciate this.  This is another way that javascript has warped my fragile little mind -- instead of parametrize, the javascript test framework lets you just add the same test function multiple times inside a for loop, and that bends your brain until it doesn't.  

---
In reply to: [305485131](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#discussion_r305485131) [](ancestors = 305485131)

BertKleewein:
>             self.transport.on_mqtt_connected = on_connected

I'm hitting this in my other PR.

---
In reply to: [513331095](https://github.com/Azure/azure-iot-sdk-python-preview/pull/133#issuecomment-513331095) [](ancestors = 513331095)

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py:100 in 66b1602. [](commit_id = 66b16026e42196fc8c5b1854c4995e887fa7f3bd, deletion_comment = False)

PR #134, opened 2019-07-09
Add pip bumpversion to setup.cfg and build-release.yaml

fratster:
+ add pip bumpversion and pipeline variables to allow engineer to bump major, minor, patch and num (pre-release) values respectively from Azure DevOps release UI
+ update setup.cfg with current version in setup.py to boostrap bumpversion tracking
+ update yaml file to call CreateReleaseBranch after version bump


azure-iot-device/setup.cfg
```diff
@@ -1,12 +1,29 @@
...
+	13
+	14
+	15
+	16
 
 [bdist_wheel]
-# This flag says to generate wheels that support both Python 2 and Python
-# 3. If your code will not run unchanged on both Python 2 and 3, you will
-# need to generate separate wheels for each Python version that you
-# support.
```
cartertinney: Please restore this comment

azure-iot-device/setup.cfg
```diff
@@ -1,12 +1,29 @@
-#-------------------------------------------------------------------------
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License. See License.txt in the project root for
-# license information.
-#--------------------------------------------------------------------------
+[bumpversion]
+current_version = 2.0.0-preview.7
```
cartertinney: This is now maintained in two different spots as I understand it, which is concerning.

fratster: ok let me clarify - in using pip bumpversion utility we declare this value in this file to be the version authority and it is henceforth updated by the bumpversion tool.  So changes to setup.py for a release will include a corresponding change in this file.  The tool uses regex for search and replace.  This flexible design accommodates incrementation of the arbitrarily labeled preview version as required.  But the trade-off is the tool must first know what to replace.  So here we are bootstrapping that value manually as a first time setup step.

azure-iot-device/setup.cfg
```diff
@@ -5,8 +5,36 @@
...
+	7
+	8
+	9
+	10
+	11
+	12
+	13
+	14
+	15
+	16
```
olivakar: >16 [](start = 1, length = 2)

are these values for major minor patch and num ? are  we okay to bump till 16 for the patch field specially ?

fratster: I'll update the documentation on this one and repost

fratster: The values are for num.  I updated the internal documentation about this section so please check that out (https://github.com/Azure/iot-sdks-internals/pull/105)

Based on this (https://www.python.org/dev/peps/pep-0386/) which says, "In Python there are no real restrictions yet on how a project should manage its versions, and how they should be incremented.", we are ok to bump until version 16 (and beyond if you choose).  But I'm no Python dev - if you have other information please share.

olivakar: I have no specific information except that I have seen the python versions itself be bumped to 10 or 16. For example python 2.7 started from 2.7.0  and ended at 2.7.16 , 
python 3.4 started at 3.4.0 and ended at 3.4.10. 
These are the examples where the incremental part though. So 16 seems fine to be a choice.

If it is for num (so the part after preview) I don't think it matters that much as it will no longer be there when we stop having preview.



---
In reply to: [302317534](https://github.com/Azure/azure-iot-sdk-python-preview/pull/134#discussion_r302317534) [](ancestors = 302317534)

azure-iot-device/setup.cfg
fratster: Also I updated the internal release doc with details about use of the bumpversion utility and linked to its reference page for more info.

cartertinney: Does this mean that we can get rid of it after first time setup then?

---
In reply to: [301743275](https://github.com/Azure/azure-iot-sdk-python-preview/pull/134#discussion_r301743275) [](ancestors = 301743275)

azure-iot-device/setup.cfg
cartertinney: I am also confused by this list of numbers. This num value can go arbitrarily high, we could theoretically get up to 2.0.0-preview.35342

Why don't we just increment num rather than picking it from a list of defined values?

---
In reply to: [302766600](https://github.com/Azure/azure-iot-sdk-python-preview/pull/134#discussion_r302766600) [](ancestors = 302766600,302317534)

azure-iot-device/setup.cfg
fratster: No it means the current_version key-value-pair will continue to live in setup.cfg and be updated by the bumpversion utility with each subsequent release.  So I'm trying to say this is the only time we have to deal with it - the tool will update the value thereafter as we direct it to.

If you have a specific reason not to include the [bumpversion] configuration setting in the .cfg file please lmk and I'll investigate if we can use the tool another way.  If you have any more questions about how bumpversion utility works please stop by my desk for a demo.

azure-iot-device/setup.cfg
fratster: I agree this approach seems awakward but it is by design.

The author give examples of enumerated strings for this configruation: alpha, beta, gamma, which are arbitrarily sorted as the user prefers.  In their documentation the author's stated intent is maximum flexibility.  Semver complainace is not mentioned and I suspect the tool predates semver but I'm not sure.

As Oliva mentions, this section will be removed when the -preview label is deprecated and the major, minor, patch values will increment as expected.

Source code: https://github.com/peritus/bumpversion

cartertinney: I took a look at the docs, and it seems the default values of the `values` attribute is already numeric incrementation, and that setting specified values is for using strings to indicate a named version. Can you double check that this is even necessary? My understanding from reading them is that you could remove these defined values and it should still work.

---
In reply to: [303091690](https://github.com/Azure/azure-iot-sdk-python-preview/pull/134#discussion_r303091690) [](ancestors = 303091690)

fratster: You are right - I tested and it works; apologies for my earlier confusion.  But in this testing I discovered a new problem - the tool strips the comments from setup.cfg as you noted so I'll need to make a separate file, test this one, and update the pr.  This file will be reverted.

PR #136, opened 2019-07-10
return errors from paho

BertKleewein:
This is not done, but I wanted to get feedback on the general organization.  Specific questions are added as comments.


azure-iot-device/azure/iot/device/common/errors.py
```diff
@@ -0,0 +1,151 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
BertKleewein: >Microsoft [](start = 16, length = 9)

open issue:  This contains a set of "base" errors.  Some are truly common (like ConnectionFailed), and others are IoTHub specific (like BadDeviceResponseError),  and even more are not truly common, but they are common between iothub and provisioning (like ThrottlingError).  From what I've seen in other SDKs, nobody has really taken the time to get a good look at what errors can be returned by what services, and that will eventually have to happen, but I'm not sure if that's a windmill that I want to tilt at.

azure-iot-device/azure/iot/device/common/error_map_paho.py
```diff
@@ -0,0 +1,110 @@
...
+    errors_paho.PahoPayloadSizeError: errors.TransportError,
+    errors_paho.PahoNotSupportedError: errors.TransportError,
+    errors_paho.PahoAuthError: errors.UnauthorizedError,
+    errors_paho.PahoACLDeniedError: errors.UnauthorizedError,
+    errors_paho.PahoUnknownError: errors.TransportError,
+    errors_paho.PahoQueueSizeError: errors.TransportError,
+}
+
+
+def _wrap_paho_error_in_generic_error(paho_error):
```
BertKleewein: >_wrap_paho_error_in_generic_error [](start = 4, length = 33)

open issue: I'd like feedback on this idea.  I think it makes sense.  For Paho errors, I convert the paho error code into a paho-specific error object (from errors_paho.py), and then I wrap that paho error inside a more generic error (from errors.py).  This way, the user gets a generic error, but they have the paho-specific details if they want them.

azure-iot-device/azure/iot/device/common/errors.py
```diff
@@ -0,0 +1,151 @@
...
+
+class MessageTooLargeError(Exception):
+    """
+    Service returned 413
+    """
+
+    pass
+
+
+class ThrottlineError(Exception):
```
BertKleewein: >ThrottlineError [](start = 6, length = 15)

spelling

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -95,37 +108,58 @@ def on_connected():
...
 
+            def on_connection_failure(cause):
+                logger.info(
+                    "{}({}): on_connection_failure.  error={}.  completing op.".format(
+                        self.name, op.name, cause
+                    )
+                )
+                op.error = cause
+                self.transport.on_mqtt_connected = self._on_unexpected_connection
+                self.transport.on_mqtt_connection_failure = self._on_unexpected_connection_failure
```
BertKleewein: These lines get repeated many times.  Should be a helper function.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -41,8 +41,9 @@ def _run_op(self, op):
                 username=self.username,
                 ca_cert=self.ca_cert,
             )
-            self.transport.on_mqtt_connected = self.on_connected
-            self.transport.on_mqtt_disconnected = self.on_disconnected
+            self.transport.on_mqtt_connected = self._on_unexpected_connection
+            self.transport.on_mqtt_connection_failure = self._on_unexpected_connection_failed
+            self.transport.on_mqtt_disconnected = self._on_unexpected_disconnection
```
BertKleewein: Another question where I think this makes sense.  these handlers (_on_unexpected_connection, etc) are only for cases where Paho calls the handler but we're not trying to connect, so I've added the "unexpected" name to indicate this.  Other cases, like we get connected and we were trying to connect, are handled by the temporary handlers below.  This seems to be the best way to handle this scenario, but it's all ugly so I'm wondering if there's a better way..

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -176,3 +210,32 @@ def _on_message_received(self, topic, payload):
             stage=self,
             event=pipeline_events_mqtt.IncomingMQTTMessageEvent(topic=topic, payload=payload),
         )
+
+    def _on_unexpected_connection(self):
```
BertKleewein: >_on_unexpected_connection [](start = 8, length = 25)

this should only get called when we get a successful CONACK but we weren't even trying to connect.  In this case, I think it makes sense to accept it as-is and call the on_connected handler,  it _shouldn't_ get called, but these 3 functions are all trying to make the best of an awkward situation.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -176,3 +210,32 @@ def _on_message_received(self, topic, payload):
...
+        """
+        Handler that gets called by the transport when it connections.  This handler only
+        gets called when it connects outside of the normal connect/reconnect logic.
+        """
+        logger.warn("an unexpected connection occured")
+        self.on_connected()
+
+    def _on_unexpected_connection_failure(self, cause):
```
BertKleewein: >_on_unexpected_connection_failure [](start = 8, length = 33)

This one is a little stranger.  In this case, Paho is saying, "that connection that you asked for failed", when we didn't ask for a connection (or we gave up).  It seems like passing the error up to the unhandled_error_handler is the best case.  Improvement suggestion: wrap this in an "unexpected connection failure" error, so the user can see an object that specifically says "there was an unexpected connection failure caused by an auth failure".  This seems more useful than just passing the auth failure up without the wrapper that indicates that it just might be useless because things are a little wonky anyway.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -176,3 +210,32 @@ def _on_message_received(self, topic, payload):
...
+        """
+        Handler that gets called by the transport when a connection fails.  This handler only
+        gets called when a connection fails outside of the normal connect/reconnect logic,
+        which should not happen, so the only real route of action is to pass the failure up
+        to the unhandled error handler
+        """
+        logger.error("an unexpected connection failure occured: {}".format(cause))
+        self.pipeline_root.unhandled_error_handler(cause)
+
+    def _on_unexpected_disconnection(self, cause):
```
BertKleewein: >_on_unexpected_disconnection [](start = 8, length = 28)

Another issue here.  An unexpected disconnection is easy, we use "on_disconnected" to pass the event up, but do we do anything special with the error that indicates why we disconnected?  I don't think we need to handle unexpected disconnections differently than expected disconnections.  Or do we?  Also, my code passes the cause up as a "unhandled error", but this could cause confusion.  Maybe I pass up an "unexpected disconnection" error wrapping the cause of the unexpected disconnection?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -176,3 +210,32 @@ def _on_message_received(self, topic, payload):
...
+        """
+        Handler that gets called by the transport when the transport disconnects.  This handler only
+        gets called when a disconnection fails outside of the normal connect/reconnect logic,
+        which should not happen, so the only real route of action is to pass the failure up
+        to the unhandled error handler
+        """
+        logger.error("an unexpected disconnection occured: {}".format(cause))
+        self.pipeline_root.unhandled_error_handler(cause)
```
BertKleewein: >unhandled_error_handler [](start = 27, length = 23)

side question: I'm looking for suggestions on a better name for "unhandled_error_handler" because this should eventually bubble up to the user.  I said "unhandled" because these are errors that wouldn't be caught by any user-level try/catch logic, usually because they happen in a background thread.  What do we call these errors?

azure-iot-device/azure/iot/device/common/errors.py
```diff
@@ -0,0 +1,151 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
olivakar: >-------------------------------------------------------------------------- [](start = 2, length = 74)

The errors with status codes like 4xx and 5xx seems to be more in line with http response status codes. Even if we may not use HTTP as a a protocol for some of these features , I am just trying to understand if it will be possible to use the HttpStatus class from python.

Also for the 4xx and 5xx errors I somehow feel it should be an enum or a dictionary mapping numeric status code to the error

azure-iot-device/azure/iot/device/common/errors_paho.py
```diff
@@ -0,0 +1,165 @@
+# --------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+
+class OperationCancelledError(Exception):
```
olivakar: >OperationCancelledError [](start = 6, length = 23)

there is an `OperationCancelledError`  in the other file. May be we should rename this to` PahoOperationCancelledError`

azure-iot-device/azure/iot/device/common/error_map.py
```diff
@@ -0,0 +1,35 @@
+# --------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+from . import errors
+
+status_code_to_error = {
```
olivakar: >status_code_to_error [](start = 0, length = 20)

oh never mind the second part of my comment on the errors.py file. You already have this dictionary here.

azure-iot-device/azure/iot/device/common/errors_paho.py
```diff
@@ -0,0 +1,165 @@
...
+
+class PahoConnectionRefusedProtocolVersionError(Exception):
+    """
+    Connection Refused: unacceptable protocol version.
+    """
+
+    pass
+
+
+class PahoConnectionRefusedIdentifierRejectedError(Exception):
```
olivakar: >PahoConnectionRefusedIdentifierRejectedError [](start = 6, length = 44)

Do you think it will be a good idea to suggest resolution for these type of error on the docstrings ? Like for this specific one it is not very clear from the name of the `CONNACK_REFUSED_IDENTIFIER_REJECTED` as well. We can suggest something like this can happen due to many reasons like 
1. `client_id` not present for paho. Make sure device_id and/or module_id is present
2. `client_id` wrong for paho

azure-iot-device/azure/iot/device/common/error_map_paho.py
```diff
@@ -0,0 +1,110 @@
...
+    mqtt.MQTT_ERR_ACL_DENIED: errors_paho.PahoACLDeniedError,
+    mqtt.MQTT_ERR_UNKNOWN: errors_paho.PahoUnknownError,
+    mqtt.MQTT_ERR_ERRNO: errors_paho.PahoUnknownError,
+    mqtt.MQTT_ERR_QUEUE_SIZE: errors_paho.PahoQueueSizeError,
+}
+
+# mapping of Paho-specific errors to more generic errors
+paho_error_to_generic_error = {
+    errors_paho.PahoConnectionRefusedProtocolVersionError: errors.TransportError,
+    errors_paho.PahoConnectionRefusedIdentifierRejectedError: errors.TransportError,
```
olivakar: I am thinking more in lines of `ConnectionFailedError` for these 2 as well

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -158,8 +181,12 @@ def connect(self, password=None, client_certificate=None):
         self._mqtt_client.tls_insecure_set(False)
         self._mqtt_client.username_pw_set(username=self._username, password=password)
 
-        self._mqtt_client.connect(host=self._hostname, port=8883)
         self._mqtt_client.loop_start()
+        rc = self._mqtt_client.connect(host=self._hostname, port=8884)
```
olivakar: >8884 [](start = 65, length = 4)

This number is different from the one before.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -158,8 +181,12 @@ def connect(self, password=None, client_certificate=None):
         self._mqtt_client.tls_insecure_set(False)
         self._mqtt_client.username_pw_set(username=self._username, password=password)
 
-        self._mqtt_client.connect(host=self._hostname, port=8883)
         self._mqtt_client.loop_start()
```
olivakar: >self._mqtt_client.loop_start() [](start = 8, length = 30)

Also I am not entirely sure but does it matter that now loop start is before the connect ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -176,3 +210,32 @@ def _on_message_received(self, topic, payload):
             stage=self,
             event=pipeline_events_mqtt.IncomingMQTTMessageEvent(topic=topic, payload=payload),
         )
+
+    def _on_unexpected_connection(self):
+        """
+        Handler that gets called by the transport when it connections.  This handler only
```
olivakar: >connections [](start = 58, length = 11)

connects or completes a connection

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
olivakar: can we call `on_disconnected()` ?

---
In reply to: [302278864](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302278864) [](ancestors = 302278864)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -176,3 +210,32 @@ def _on_message_received(self, topic, payload):
...
+        gets called when it connects outside of the normal connect/reconnect logic.
+        """
+        logger.warn("an unexpected connection occured")
```
olivakar: >occured [](start = 46, length = 7)

spell

azure-iot-device/azure/iot/device/common/error_map_paho.py
olivakar: i like the idea in general. But is it possible to see a stack trace with all the wrappings ? From the code here it seems it is mapping and returning the generic error.


---
In reply to: [302268054](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302268054) [](ancestors = 302268054)

azure-iot-device/azure/iot/device/common/error_map_paho.py
```diff
@@ -0,0 +1,110 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+from . import errors
+from . import errors_paho
+import paho.mqtt.client as mqtt
+
+# mapping of Paho conack rc codes to Error object classes
+paho_conack_rc_to_error = {
+    mqtt.CONNACK_REFUSED_PROTOCOL_VERSION: errors_paho.PahoConnectionRefusedProtocolVersionError,
```
cartertinney: I'm not sure I agree with the premise of what you're trying to do here.

Why would we 1:1 translate paho errors into erros we define in our SDK? Why turn all Paho errors into a generic (or a small number of more generic) SDK related errors, while preserving the error history to it's initial cause? (i.e. error chaining)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -60,8 +64,20 @@ def _create_mqtt_client(self):
         def on_connect(client, userdata, flags, rc):
             logger.info("connected with result code: {}".format(rc))
 
-            # TODO: how to do failed connection?
-            if self.on_mqtt_connected:
+            if rc:
```
cartertinney: >if rc [](start = 12, length = 5)

RC always indicates error? There's no success RC?

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: I believe loop_start is supposed to be after connect, at least, that's how all the samples in the docs have it.

---
In reply to: [302767253](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302767253) [](ancestors = 302767253)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: Was there a reason this was changed?

---
In reply to: [302767087](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302767087) [](ancestors = 302767087)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: Are externally driven connects possible? Reconnects? If so, wouldn't this spuriously trigger? (my understanding is that this is intended to log unexpected connections as a result of threading issues in the pipeline)

---
In reply to: [302270895](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302270895) [](ancestors = 302270895)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: I see the value in this, but I think the code is somewhat confusing. Perhaps we could find a better way to contain this logic.

Also does the dynamic switching of handlers present any potential threading issues / race conditions?

---
In reply to: [302270134](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302270134) [](ancestors = 302270134)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: brain fart

---
In reply to: [303132683](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r303132683) [](ancestors = 303132683,302767253)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: good catch.  not a clue how I changed this.  

---
In reply to: [303132850](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r303132850) [](ancestors = 303132850,302767087)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: Yes, rc==0 is success.

---
In reply to: [303131872](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r303131872) [](ancestors = 303131872)

azure-iot-device/azure/iot/device/common/error_map_paho.py
BertKleewein: They are connection failures, but it makes no sense to retry, so I separated them out.  I think I'll make something like "ConnectionFailedNoRetry" to be explicit about it.

---
In reply to: [302699678](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302699678) [](ancestors = 302699678)

azure-iot-device/azure/iot/device/common/error_map.py
```diff
@@ -0,0 +1,35 @@
+# --------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
cartertinney: Feels like this file, and errors.py could be combined into one module

azure-iot-device/azure/iot/device/common/errors.py
cartertinney: I agree, for now, I'm just happy to have errors, we can reorganize a bit later.

---
In reply to: [302267292](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302267292) [](ancestors = 302267292)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -61,8 +115,18 @@ def _create_mqtt_client(self):
...
```
cartertinney: >if rc [](start = 12, length = 5)

I feel like all this branching logic here could use a comment to clarify

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: I think a brief comment to that effect could be helpful, i.e. ```if rc:    #rc 0 == success```

---
In reply to: [305849146](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r305849146) [](ancestors = 305849146,303131872)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: >if rc: [](start = 12, length = 6)

I'm unclear on how (and why) this flow is different from the error handling based on the rc in the .`connect` method. Could you clarify?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -25,6 +26,28 @@ class MQTTTransportStage(PipelineStage):
...
 
+    @pipeline_thread.runs_on_pipeline_thread
+    def _cancel_active_connect_disconnect_ops(self):
+        """
+        Cancel any running connect or disconnect op.  Since our ability to "cancel" is fairly limited,
+        all this does (for now) is to fail the operation
+        """
+
+        ops_to_cancel = []
+        if self._active_connect_op:
```
cartertinney: >self._active_connect_op [](start = 11, length = 23)

Can there be more than one?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: I agree: put that in a `TODO:`

---
In reply to: [302272141](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302272141) [](ancestors = 302272141)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: I agree - our error story should be as clear as possible. Unexpected disconnection would be preferable.

---
In reply to: [302278864](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302278864) [](ancestors = 302278864)

cartertinney:
>     @pytest.mark.it("Triggers on_mqtt_connected event handler callback upon connect completion")

Change this to be "upon successful connect completion" now that there is also a fail case

---
Refers to: azure-iot-device/tests/common/test_mqtt_transport.py:282 in f54d013. [](commit_id = f54d013a0ff1934c2924b8debcf6befb7db9f1d3, deletion_comment = False)

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -94,7 +180,7 @@ def test_operation_infrastructure_set_up(self, mocker):
         assert transport._op_manager._unknown_operation_completions == {}
 
 
-@pytest.mark.describe("MQTTTransport - .connect()")
+@pytest.mark.describe("MQTTTransport - .connect() / Event: Connected, ConnectionFailure")
```
cartertinney: >Event: [](start = 52, length = 6)

All caps

cartertinney: >Connected [](start = 59, length = 9)

Is "Connected" really an externally driven event? How can connect be externally driven? Isn't "ConnectionFailure" the only externally driven event?

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -247,6 +359,19 @@ def test_event_handler_callback_raises_base_exception(
                 client=mock_mqtt_client, userdata=None, flags=None, rc=fake_rc
             )
 
+    @pytest.mark.parametrize(
+        "error_params",
+        operation_return_codes,
+        ids=["{}->{}".format(x["name"], x["error"].__name__) for x in operation_return_codes],
+    )
+    @pytest.mark.it("Raises an Exception if connect returns a failing rc code")
```
cartertinney: >Exception [](start = 31, length = 9)

custom Exception. Otherwise this implies you're literally raising `Exception`.

Here and everywhere.

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -208,6 +294,32 @@ def test_calls_event_handler_callback(self, mocker, mock_mqtt_client, transport)
         assert callback.call_count == 1
         assert callback.call_args == mocker.call()
 
+    @pytest.mark.parametrize(
+        "error_params",
+        conack_return_codes,
+        ids=["{}->{}".format(x["name"], x["error"].__name__) for x in conack_return_codes],
+    )
+    @pytest.mark.it(
+        "Triggers on_mqtt_connection_failure event handler callback upon connect completion with failed RC"
```
cartertinney: 1) "event handler callback" is redundant. Just "event handler" should suffice.
2) "Triggers on_mqtt_connection_failure event handler with custom Exception...." perhaps? This makes the Exception (which will display as a result of parametrization) make more sense in the requirement.

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -299,6 +424,29 @@ def test_calls_event_handler_callback_externally_driven(
         assert callback.call_count == 1
         assert callback.call_args == mocker.call()
 
+    @pytest.mark.parametrize(
+        "error_params",
+        conack_return_codes,
+        ids=["{}->{}".format(x["name"], x["error"].__name__) for x in conack_return_codes],
+    )
+    @pytest.mark.it(
+        "Triggers on_mqtt_connection_failure event handler callback upon completion of externally-driven reconnect returns failed RC"
```
cartertinney: >reconnect [](start = 105, length = 9)

If this is for reconnect, shouldn't it be in the reconnect tests?

Also, I'm not clear what an externally driven reconnect is

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -133,9 +142,15 @@ def test_passes_right_params(self, stage, transport, mocker, op_set_connection_a
     )
```
cartertinney: Needs to be updated to include `on_mqtt_connection_failure`

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -412,22 +374,145 @@ def test_verify_incoming_message_attributes(self, stage, create_transport, mocke
 @pytest.mark.describe("MQTTTransportStage - EVENT: MQTT connected")
 class TestMQTTProviderOnConnected(object):
     @pytest.mark.it(
-        "Calls self.on_connected and passes it up when the client library connected event fires"
+        "Calls self.on_connected and passes it up when the client library connected event fires, if there is no active connect op"
```
cartertinney: >it [](start = 44, length = 2)

it? What is "it"?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
BertKleewein: fixed in my new iteration.

---
In reply to: [303138886](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r303138886) [](ancestors = 303138886,302270134)

azure-iot-device/azure/iot/device/common/error_map.py
BertKleewein: Not used anywhere yet.  This is write-ahead.  I'll need it eventually before the review was scoped down to Paho-only.

---
In reply to: [307876441](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r307876441) [](ancestors = 307876441)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
BertKleewein: This should not be possible, but I believe strongly in handling unexpected conditions as gracefully as possible.  Regardless, this code has been refactored.

---
In reply to: [303137777](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r303137777) [](ancestors = 303137777,302270895)

azure-iot-device/azure/iot/device/common/errors.py
BertKleewein: Great idea.  I'll do this when I start actually consuming these status codes.

---
In reply to: [302681618](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302681618) [](ancestors = 302681618)

azure-iot-device/azure/iot/device/common/errors_paho.py
BertKleewein: these errors have been removed in favor of the more generic counterparts.


---
In reply to: [302682009](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302682009) [](ancestors = 302682009)

azure-iot-device/azure/iot/device/common/errors_paho.py
BertKleewein: I like that idea, but I think that's better added after we get an idea of what can cause the errors we have.  Right now we're still narrowing down the specific errors.

---
In reply to: [302699577](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r302699577) [](ancestors = 302699577)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: This has all been refactored, but a connection or reconnection can fail in two different ways:
1. If Paho fails before it can even send the CONNECT packet to the service, then the .connect method fails.
2. If Paho gets an error back from the server in the CONACK packet, then that error gets returned via the on_connected handler .

---
In reply to: [307882272](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r307882272) [](ancestors = 307882272)

azure-iot-device/tests/common/test_mqtt_transport.py
BertKleewein: brain fart.

---
In reply to: [307912760](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r307912760) [](ancestors = 307912760)

azure-iot-device/tests/common/test_mqtt_transport.py
BertKleewein: Neither _should_ be externally driven.  

---
In reply to: [307910811](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r307910811) [](ancestors = 307910811)

azure-iot-device/azure/iot/device/common/errors.py
```diff
@@ -0,0 +1,189 @@
...
+
+class FailedStatusCodeError(Exception):
+    """
+    Service returned unknown status code
+    """
+
+    pass
+
+
+class TransportError(Exception):
```
cartertinney: This should probably be changed to `ClientError` or `ProtocolClientError` since "Transport" refers to our class that wraps the protocol client.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -45,89 +69,47 @@ def _run_op(self, op):
                 ca_cert=self.ca_cert,
                 x509_cert=self.client_cert,
             )
-            self.transport.on_mqtt_connected = self.on_connected
-            self.transport.on_mqtt_disconnected = self.on_disconnected
-            self.transport.on_mqtt_message_received = self._on_message_received
+            self.transport.on_mqtt_connected = self._handle_mqtt_connected
+            self.transport.on_mqtt_connection_failure = self._handle_mqtt_connection_failure
+            self.transport.on_mqtt_disconnected = self._handle_mqtt_disconnected
```
cartertinney: Was there a particular reason we moved away from the `on_x` convention to a `handle_x` convention?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -45,89 +69,47 @@ def _run_op(self, op):
...
-                logger.info("{}({}): on_connected.  completing op.".format(self.name, op.name))
-                self.transport.on_mqtt_connected = self.on_connected
-                self.on_connected()
-                operation_flow.complete_op(self, op)
-
-            # See "A note on exception handling" above
-            self.transport.on_mqtt_connected = on_connected
+            # We set _active_connect_op here because a reconnect is the same as a connect for "active operation" tracking purposes.
+            self._cancel_active_connect_disconnect_ops()
+            self._active_connect_op = op
```
cartertinney: So, I have a slight concern about this.

We ALWAYS must call `_cancel_active_connect_disconnect_ops()` before setting the `_active_connect_op` (or the `_active_disconnect_op` for that matter). But it's not enforced. If someone were to modify this file, and forget that prerequisite call to `_cancel_active_connect_disconnect_ops()`, we would enter a land of undefined and unexpected behaviour.

In order to protect ourselves, I would ideally like to see rigorous tests for connect/disconnect/reconnect that show that we always have the expected behaviour (setting a new active op, clears all previously active ops). Another thing that could help is enforcing the cancel before the set.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -258,30 +278,28 @@ def transport_function_throws_base_exception(params, stage, mocker, fake_base_ex
 
 @pytest.mark.parametrize("params", connection_ops + pubsub_ops)
 @pytest.mark.describe(
-    "MQTTTransportStage - .run_op() -- called with op that maps directly to protocol client library calls"
+    "MQTTTransportStage - .run_op() -- called with op that maps directly to transport calls"
```
cartertinney: >"MQTTTransportStage - .run_op() -- called with op that maps directly to transport calls" [](start = 4, length = 88)

Taking this approach of parametrizing the ops limits us to having the same tests for all ops - which prevents us from being able to test the connect/reconnect/disconnect specific "only one op active at a time, doing another one cancels the pre-existing one" requirement.

cartertinney: Taking this approach of parametrizing the ops limits us to having the same tests for all ops - which prevents us from being able to test the connect/reconnect/disconnect specific "only one op active at a time, doing another one cancels the pre-existing one" requirement.

This flexibility is generally why I have favored the inheritance model for shared tests in the past.

cartertinney: I also don't think there's any validation of the callbacks `on_published()`, `on_subscribed()`, `on_unsubscribed()` due to the messaging family of operations having to share complete testing infrastructure with the connection family of operations.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
BertKleewein: Because we went through this whole thing a few months ago where went back and forth about what to call things and my memory wasn't in sync with the wiki.  I'll sync with the wiki.

---
In reply to: [309365971](https://github.com/Azure/azure-iot-sdk-python-preview/pull/136#discussion_r309365971) [](ancestors = 309365971)

PR #141, opened 2019-07-19
Client-level preparation for EdgePipeline

cartertinney:
* Added a stub for an `EdgePipeline` pipeline class
* Altered abstract client signatures to allow for multiple pipelines
* Added logic to create multiple pipelines where appropriate
* Added tests, and refactored existing tests to better ensure pipeline configuration validity
* Refactored all references of `trusted_certificate_chain` to `ca_cert` instead to be consistent.
* Cleaned up artifacts of `MQTTClientOperator` -> `MQTTTransport` rename


azure-iot-device/azure/iot/device/iothub/auth/sk_authentication_provider.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         self.shared_access_key = shared_access_key
         self.shared_access_key_name = shared_access_key_name
         self.gateway_hostname = gateway_hostname
-        self.ca_cert = None  # TODO: rename to trusted_certificate_chain?
+        self.ca_cert = None  # TODO: rename to ca_cert?
```
BertKleewein: >TODO: rename to ca_cert? [](start = 31, length = 24)

Um, no longer a TODO?

azure-iot-device/azure/iot/device/iothub/pipeline/edge_pipeline.py
```diff
@@ -0,0 +1,18 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class EdgePipeline(object):
```
BertKleewein: >EdgePipeline [](start = 6, length = 12)

I still think this name is confusing, but I'm not sure what's better.  

EdgeHTTPPipeline?  (nah, don't need HTTP, but there must be _some_ way to communicate why this pipeline is different.
EdgeSpecificFeaturePipeline?
EdgeMethodInvokePipeline?

Luckily this is internal, so we have a little more freedom to rename in the future.
FWIW, the methods in this pipeline are handled by the EdgeHub module, so there is some symmetry if we call this EdgeHubPipeline to mirror IoTHubPipeline
Or, if we're being pedantic, maybe this is EdgeHub pipeline because it's only useful with EdgeHub, but the other one is HubPipeline since it can be used with EdgeHub or IoTHub?

azure-iot-device/tests/iothub/client_fixtures.py
```diff
@@ -8,6 +8,12 @@
 from azure.iot.device.iothub.pipeline import constant
 from azure.iot.device.iothub.models import Message, MethodResponse, MethodRequest
 from azure.iot.device.common.models.x509 import X509
+from azure.iot.device.iothub.auth import (
+    SymmetricKeyAuthenticationProvider,
+    SharedAccessSignatureAuthenticationProvider,
+    IoTEdgeAuthenticationProvider,
+    X509AuthenticationProvider,
+)
```
BertKleewein: Not obviously used.  For future use or is there other pytest magic going on?

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -186,9 +206,9 @@ def create_from_edge_environment(cls):
                 workload_uri=workload_uri,
                 api_version=api_version,
             )
-
-        pipeline = IoTHubPipeline(authentication_provider)
-        return cls(pipeline)
+        iothub_pipeline = pipeline.IoTHubPipeline(authentication_provider)
+        edge_pipeline = pipeline.EdgePipeline(authentication_provider)
```
BertKleewein: >edge_pipeline = pipeline.EdgePipeline(authentication_provider) [](start = 8, length = 62)

I'm glad I saw this.  One auth provider serving two pipelines has implications for sas renewal.  The auth provider needs to notify _both_ pipelines when the token is renewed.  No big deal, but important to get right.

BertKleewein: And, right now, the auth provider isn't notifying anyone on token renewal, hence issue #142, so you won't find any code to change yet.  

---
In reply to: [306955500](https://github.com/Azure/azure-iot-sdk-python-preview/pull/141#discussion_r306955500) [](ancestors = 306955500)

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -859,18 +995,18 @@ def client_class(self):
...
+    def client(self, iothub_pipeline):
         """This client automatically resolves callbacks sent to the pipeline.
         It should be used for the majority of tests.
         """
-        return IoTHubModuleClient(pipeline)
+        return IoTHubModuleClient(iothub_pipeline)
 
     @pytest.fixture
-    def client_manual_cb(self, pipeline_manual_cb):
+    def client_manual_cb(self, mocker, iothub_pipeline_manual_cb):
```
BertKleewein: >mocker [](start = 31, length = 6)

don't need this fixture

PR #147, opened 2019-07-25
Added User-Agent

cartertinney:
* Added User-Agent to all communications
* Created a single source of truth for package version, user agent, and api versions
* Updated tests
* Altered provisioning username to correctly use query parameters


azure-iot-device/azure/iot/device/constant.py
```diff
@@ -0,0 +1,12 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module defines constants for use across the azure-iot-device package
```
cartertinney: I considered naming this module `version.py` instead, but technically the User-Agent isn't really a "version", so it felt weird.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -23,8 +23,6 @@
 )
 from tests.common.pipeline import pipeline_stage_test
 
-logging.basicConfig(level=logging.INFO)
```
cartertinney: I don't think I should have removed this, actually. These logs only appear on failure. Which actually suggests that __EVERY___ test file should probably have this. Separate PR?

scripts/build-release.ps1
```diff
@@ -23,17 +23,24 @@ function Build {
 
     # hashset key is package folder name in repo
 
-    $packages = @{ }
-    $packages["azure-iot-device"] = $env:device_version_part
-    $packages["azure-iot-nspkg"] = $env:nspkg_version_part
-    # TODO add new packages to this list
+    $packages = @{ } # TODO add new packages to this list
+
+    $packages["azure-iot-device"] = [PSCustomObject]@{
```
cartertinney: Essentially what these changes in this file do is allow for different packages to have the package version maintained in a customizable location. Not all packages will need to scale up to having a `constant.py` type of file maintaining a single source of truth. Sometimes just maintaining within the `setup.py` will suffice.

azure-iot-device/azure/iot/device/constant.py
olivakar: what about naming it `client_verison.py` 

---
In reply to: [307432156](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307432156) [](ancestors = 307432156)

azure-iot-device/azure/iot/device/constant.py
```diff
@@ -0,0 +1,12 @@
...
+"""
+
+VERSION = "2.0.0-preview.8"
+USER_AGENT = "py-azure-iot-device/{version}".format(version=VERSION)
```
olivakar: >py- [](start = 14, length = 3)

is this just constructed by us ? or is there a format that we are supposed to follow ? asking because I did not see `py` before anywhere, but may be mistaken. And if there is no format then why don't we use `python` fully ?

azure-iot-device/azure/iot/device/iothub/auth/iotedge_authentication_provider.py
```diff
@@ -112,7 +113,9 @@ def get_trust_bundle(self):
         :raises: IoTEdgeError if unable to retrieve the certificate.
         """
         r = requests.get(
-            self.workload_uri + "trust-bundle", params={"api-version": self.api_version}
+            self.workload_uri + "trust-bundle",
+            params={"api-version": self.api_version},
+            headers={"User-Agent": urllib.parse.quote_plus(constant.USER_AGENT)},
```
BertKleewein: >headers={"User-Agent": urllib.parse.quote_plus(constant.USER_AGENT)}, [](start = 12, length = 69)

Is there a reason we need this, or are you just following some "user-agent everywhere" mantra?  I think either is OK, but it may not be necessary to have this, fewer lines of code to break, yadda yadda yadda.

olivakar:
>         instance, as a PEM certificate in string form.

`PEM file` ? so this returns the contents of the as a PEM file in string form correct ?

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/iotedge_authentication_provider.py:111 in 7909a74. [](commit_id = 7909a747926c5ff0d12de7ccdd35361bc8ebfc3c, deletion_comment = False)

azure-iot-device/azure/iot/device/constant.py
```diff
@@ -0,0 +1,12 @@
...
+IOTHUB_API_VERSION = "2018-06-30"
```
BertKleewein: >IOTHUB_API_VERSION [](start = 0, length = 18)

To save renaming later, I think these are more accurately IOTHUB_DEVICE_API_VERSION and PROVISIONING_DEVICE_API_VERSION.  I don't know if device and service API versions update in lock-step, but I think it might be a mistake to assume that they do.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -48,8 +51,14 @@ def _run_op(self, op):
             else:
                 client_id = op.device_id
 
-            username = "{hostname}/{client_id}/?api-version=2018-06-30".format(
-                hostname=op.hostname, client_id=client_id
+            query_params = [
+                ("api-version", pkg_constant.IOTHUB_API_VERSION),
+                ("DeviceClientType", pkg_constant.USER_AGENT),
```
olivakar: >DeviceClientType [](start = 18, length = 16)

So this becomes a different variable name here as opposed to `User-Agent` ?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: Yes!  Pipeline debugging relies heavily on these logs and I think we should do a much better job of adding logging to our code for post-mortem debugging.

---
In reply to: [307434154](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307434154) [](ancestors = 307434154)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -48,8 +51,14 @@ def _run_op(self, op):
             else:
                 client_id = op.device_id
 
-            username = "{hostname}/{client_id}/?api-version=2018-06-30".format(
-                hostname=op.hostname, client_id=client_id
+            query_params = [
```
olivakar: >query_params [](start = 12, length = 12)

I would like to rename this variable to something different that gives the idea that we have query_param kind of structure in an username field but actually they are not true query_params

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning_mqtt.py
```diff
@@ -43,12 +44,14 @@ def _run_op(self, op):
             # always pass it down because MQTT protocol stage will also want to receive these args.
 
             client_id = op.registration_id
-            client_version = urllib.parse.quote_plus(constant.USER_AGENT)
-            username = "{id_scope}/registrations/{registration_id}/api-version={api_version}&ClientVersion={client_version}".format(
+            query_params = [
+                ("api-version", pkg_constant.PROVISIONING_API_VERSION),
+                ("ClientVersion", pkg_constant.USER_AGENT),
+            ]
+            username = "{id_scope}/registrations/{registration_id}/?{query_params}".format(
```
olivakar: >? [](start = 67, length = 1)

as we found out yesterday that this question mark is not needed for DPS.  To maintain similarity with other SDKs we should remove this. But it is better to add a Note/Comment above.

azure-iot-device/azure/iot/device/constant.py
cartertinney: Given that this is within the azure-iot-device package I figured the "device"ness was implied.

---
In reply to: [307826928](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307826928) [](ancestors = 307826928)

azure-iot-device/azure/iot/device/constant.py
cartertinney: This is whatever we want it to be. It's entirely arbitrary. I chose 'py' to make it shorter, and since 'py' is the accepted abbreviation for Python (pypi, pytest, pycharm, the py library itself, etc.)

---
In reply to: [307825444](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307825444) [](ancestors = 307825444)

azure-iot-device/azure/iot/device/constant.py
cartertinney: User-Agent still realy isn't a "version"

---
In reply to: [307824566](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307824566) [](ancestors = 307824566,307432156)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
cartertinney: Okay let's make a separate PR for this - I'll put it on the backlog

---
In reply to: [307827509](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307827509) [](ancestors = 307827509,307434154)

azure-iot-device/azure/iot/device/iothub/auth/iotedge_authentication_provider.py
cartertinney: Yes, in fact, this is specifically the entire reason for this PR. The lack of User-Agent was causing Edge to throw errors (!!!!)

---
In reply to: [307825867](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307825867) [](ancestors = 307825867)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning_mqtt.py
cartertinney: I'm going to wait until I get an answer from Robledo regarding whether or not the DPS hub prefers this.

---
In reply to: [307828656](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307828656) [](ancestors = 307828656)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
cartertinney: Yes. For IoTHub, in MQTT, the User-Agent is added as 'DeviceClientType', and in DPS it is added as 'ClientVersion'

---
In reply to: [307827069](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#discussion_r307827069) [](ancestors = 307827069)

cartertinney:
>         instance, as a PEM certificate in string form.

I mean... it returns what would be the contents of a .pem file, i.e. the trusted_cert (as we discussed yesterday), in string form. It does NOT return a path to an actual .pem file

---
In reply to: [515524558](https://github.com/Azure/azure-iot-sdk-python-preview/pull/147#issuecomment-515524558) [](ancestors = 515524558)

---
Refers to: azure-iot-device/azure/iot/device/iothub/auth/iotedge_authentication_provider.py:111 in 7909a74. [](commit_id = 7909a747926c5ff0d12de7ccdd35361bc8ebfc3c, deletion_comment = False)

Issue #144, opened 2019-07-24
receive_c2d_message() seems to break when an empty c2d message is sent


bukarev:
I'm running this SDK on Raspbian with Python 3.7.4.

A device has been provisioned using group enrollment and X509 certificate. So I'm running _receive_c2d_message_x509.py_ sample script and sending messages to my device from Azure Portal. It works fine with one exception.

When a message's Body is empty, and only custom properties are set, the Portal reports the message as sent. However, it's not printed out by _receive_c2d_message_x509.py_ script. Any subsequent messages are not printed, either, even if I restart the script. 

When I try to send a d2c message, the following error stack is printed out:

`
ERROR:azure.iot.device.common.pipeline.pipeline_stages_base:Error in <azure.iot.device.common.pipeline.pipeline_stages_mqtt.MQTTClientStage object at 0xb673ffac>._run_op() call
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/azure/iot/device/common/pipeline/pipeline_stages_base.py", line 94, in run_op
    self._run_op(op)
  File "/usr/local/lib/python3.7/site-packages/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py", line 104, in _run_op
    raise e
  File "/usr/local/lib/python3.7/site-packages/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py", line 100, in _run_op
    password=self.sas_token, client_certificate=self.trusted_certificate_chain
  File "/usr/local/lib/python3.7/site-packages/azure/iot/device/common/mqtt_transport.py", line 157, in connect
    self._mqtt_client.tls_set_context(context=ssl_context)
  File "/usr/local/lib/python3.7/site-packages/paho/mqtt/client.py", line 688, in tls_set_context
    raise ValueError('SSL/TLS has already been configured.')
ValueError: SSL/TLS has already been configured.
`

The only way to correct this behaviour is by deleting the device in the Azure Portal and then reprovisioning it.

Seems like a bug on the IoT receiving side?

cartertinney:
We believe this to be an issue with our SDK, and are looking into solutions currently. This is a high priority bug. If you encounter this issue again before a fix is delivered, please try restarting your application rather than deleting and reprovisioning your device.

cartertinney:
This bug has been fixed in #149

cartertinney:
This bug has been fixed in the latest release, 2.0.0-preview.9 (9919643), now available on pip

PR #149, opened 2019-07-26
[BUGFIX] SSL Error on multiple connects

cartertinney:
* Adjusted SSL config to occur at MQTTTransport instantiation, rather than connection
* Made password optional for reconnect

I left a few comments re: some issues I came across in the existing testing. Some may be addressed in this PR, some may need to be addressed later for the sake of getting this bugfix out to respond to issue #144


azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -170,13 +133,49 @@ def test_configures_tls_context_with_client_provided_certificate_chain(
...
+        transport = MQTTTransport(
+            client_id=fake_device_id, hostname=fake_hostname, username=fake_username
+        )
+        assert transport._op_manager._pending_operation_callbacks == {}
+        assert transport._op_manager._unknown_operation_completions == {}
+
+
+@pytest.mark.describe("MQTTTransport - .connect()")
+class TestConnect(object):
     @pytest.mark.it("Sets username and password")
```
cartertinney: >Sets [](start = 21, length = 4)

I noticed these tests aren't sufficiently testing the optionality of the password. These could use some shoring up (same for reconnect). I'll be doing this before I merge this.

cartertinney:
> class TestMQTTProviderRunOpWithConnect(object):

None of these tests actually... check... the call to transport.connect....

I will also fix that.

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py:313 in c5c238c. [](commit_id = c5c238cab2ae03f54452eeac21d6c86fd4b3cf11, deletion_comment = False)

cartertinney:
> # license information.

Interesting question re: testing wrt this module.

The bug being fixed by this PR (connect called multiple times) isn't really the type of thing that these unittests cover, since it's a bug that results from a series of calls across units, not just a singular one. (i.e. the problem was that things broke if `connect()` was called twice).

As a result, I haven't really been able to add a test to show that this works now. But maybe that's not the place of these unit tests? Perhaps connect -> connect again is a case for E2E tests?

How do ya'll want to capture this?

---
Refers to: azure-iot-device/tests/common/test_mqtt_transport.py:4 in c5c238c. [](commit_id = c5c238cab2ae03f54452eeac21d6c86fd4b3cf11, deletion_comment = False)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -25,39 +25,47 @@ class MQTTTransport(object):
...
-        self._mqtt_client = mqtt.Client(
+        # Instantiate client
+        mqtt_client = mqtt.Client(
             client_id=self._client_id, clean_session=False, protocol=mqtt.MQTTv311
         )
-        self._mqtt_client.enable_logger(logging.getLogger("paho"))
+        mqtt_client.enable_logger(logging.getLogger("paho"))
 
+        # Configure TLS/SSL
+        ssl_context = self._create_ssl_context()
```
cartertinney: I removed the `mqtt_client.tls_set_insecure(False)` because it already defaults to `False`, and having the call at all in the code is more dangerous than not having it (since all it would take to enable it to change `False` to `True`). Furthermore, the API itself recommends not including it in production code for this reason.

olivakar: +1

---
In reply to: [307945405](https://github.com/Azure/azure-iot-sdk-python-preview/pull/149#discussion_r307945405) [](ancestors = 307945405)

olivakar:
>     def connect(self, password=None, client_certificate=None):

if we are instantiating the mqtt itself with a client provided certificate then is this needed here ?
The ssl context should have been set already right ?

---
Refers to: azure-iot-device/azure/iot/device/common/mqtt_transport.py:157 in c5c238c. [](commit_id = c5c238cab2ae03f54452eeac21d6c86fd4b3cf11, deletion_comment = False)

olivakar:
>     def connect(self, password=None, client_certificate=None):

also can we please mention that the password is optional because there is a possibility that the client has been connected via some X509 certificate..
same type of docstring on reconnect as well.

---
Refers to: azure-iot-device/azure/iot/device/common/mqtt_transport.py:157 in c5c238c. [](commit_id = c5c238cab2ae03f54452eeac21d6c86fd4b3cf11, deletion_comment = False)

cartertinney:
>     def connect(self, password=None, client_certificate=None):

Whoops, I forgot to remove this. Good catch.

---
In reply to: [516099098](https://github.com/Azure/azure-iot-sdk-python-preview/pull/149#issuecomment-516099098) [](ancestors = 516099098)

---
Refers to: azure-iot-device/azure/iot/device/common/mqtt_transport.py:157 in c5c238c. [](commit_id = c5c238cab2ae03f54452eeac21d6c86fd4b3cf11, deletion_comment = False)

cartertinney:
>     def connect(self, password=None, client_certificate=None):

The docstring already does suggest it is optional.

---
In reply to: [516114183](https://github.com/Azure/azure-iot-sdk-python-preview/pull/149#issuecomment-516114183) [](ancestors = 516114183)

---
Refers to: azure-iot-device/azure/iot/device/common/mqtt_transport.py:157 in c5c238c. [](commit_id = c5c238cab2ae03f54452eeac21d6c86fd4b3cf11, deletion_comment = False)

cartertinney:
> # license information.

As per discussion offline, this will be captured in an E2E test. I have created a task for it on the workboard.

---
In reply to: [515631833](https://github.com/Azure/azure-iot-sdk-python-preview/pull/149#issuecomment-515631833) [](ancestors = 515631833)

---
Refers to: azure-iot-device/tests/common/test_mqtt_transport.py:4 in c5c238c. [](commit_id = c5c238cab2ae03f54452eeac21d6c86fd4b3cf11, deletion_comment = False)

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
```diff
@@ -152,8 +152,7 @@ def test_connect_calls_connect_on_provider(
             assert mock_mqtt_transport.connect.call_args[1]["password"] is not None
             assert_for_symmetric_key(mock_mqtt_transport.connect.call_args[1]["password"])
         elif params_security_clients["client_class"].__name__ == "X509SecurityClient":
-            assert mock_mqtt_transport.connect.call_args[1]["client_certificate"] is not None
-            assert_for_client_x509(mock_mqtt_transport.connect.call_args[1]["client_certificate"])
+            assert mock_mqtt_transport.connect.call_args[1]["password"] is None
```
cartertinney: These changes need to be made to keep the tests running. However, the fact that they even broke is because tests are not following the preferred unit testing model. These top level pipeline interface tests were actually also testing what happened at the bottom of the pipeline. These are totally separate units.

In order to make the tests pass, I have altered these tests, however, this suggests coverage is not on par with the sister class in the iothub package.

PR #150, opened 2019-07-29
[INFRASTRUCTURE] Added logging to all test modules

cartertinney:
* All test modules now will capture logs, which will be only displayed in the output upon failure


azure-iot-device/tests/common/test_asyncio_compat.py
```diff
@@ -7,8 +7,11 @@
 import pytest
 import asyncio
 import sys
+import logging
 from azure.iot.device.common import asyncio_compat
 
+logging.basicConfig(level=logging.INFO)
```
BertKleewein: >INFO [](start = 34, length = 4)

do we want INFO or DEBUG?  I think we have a ton of logging at INFO but almost none at DEBUG, so it doesn't matter much today, but do we want to shift more to DEBUG and make INFO less noisy?

Issue #152, opened 2019-07-30
New Message without explictly set correlation_id AND message_id fails when sent


bf1967:
I'm working with the the send_d2c_message.py from the samples folder.

If I don't explicitly set the correlation_id AND the message_id of the message the message fails to send.  Is this behavior by design or a bug?    If one or the other is set it works fine.

Calling send_d2c_message with just a message payload works just fine.  I'm trying to add custom_properties to my message and discovered this behavior.

I can provide logs if needed.

olivakar:
We are looking into it.

cartertinney:
This bug has been fixed in the latest release, 2.0.0-preview.9 (9919643), now available on pip

bf1967:
Your fix did the trick.   Thanks for the quick turnaround @olivakar @cartertinney.

PR #153, opened 2019-07-30
Properties bug

olivakar:
https://github.com/Azure/azure-iot-sdk-python-preview/issues/152

### DIFF
--- a/azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
+++ b/azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
@@ -257,7 +257,8 @@ def encode_properties(message_to_send, topic):
     topic += system_properties_encoded
 
     if message_to_send.custom_properties and len(message_to_send.custom_properties) > 0:
-        topic += "&"
+        if system_properties and len(system_properties) > 0:
+            topic += "&"
         user_properties_encoded = urllib.parse.urlencode(message_to_send.custom_properties)
         topic += user_properties_encoded
 
--- a/azure-iot-device/samples/legacy-samples/send_telemetry.py
+++ b/azure-iot-device/samples/legacy-samples/send_telemetry.py
@@ -17,8 +17,8 @@ device_client = IoTHubDeviceClient.create_from_connection_string(conn_str)
 # Connect the client.
 device_client.connect()
 
-# send 5 messages with a 1 second pause between each message
-for i in range(1, 6):
+# send 2 messages with 2 system properties & 1 custom property with a 1 second pause between each message
+for i in range(1, 3):
     print("sending message #" + str(i))
     msg = Message("test wind speed " + str(i))
     msg.message_id = uuid.uuid4()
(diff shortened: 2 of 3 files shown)


azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -257,7 +257,8 @@ def encode_properties(message_to_send, topic):
     topic += system_properties_encoded
 
     if message_to_send.custom_properties and len(message_to_send.custom_properties) > 0:
-        topic += "&"
+        if system_properties and len(system_properties) > 0:
```
BertKleewein: >if system_properties and len(system_properties) > 0: [](start = 8, length = 52)

There should be a unit test to validate this

azure-iot-device/samples/legacy-samples/send_telemetry.py
```diff
@@ -17,8 +17,8 @@
 # Connect the client.
 device_client.connect()
 
-# send 5 messages with a 1 second pause between each message
-for i in range(1, 6):
+# send 2 messages with 2 system properties & 1 custom property with a 1 second pause between each message
```
BertKleewein: We definitely need E2E tests for these scenarios (and probably unit tests to make sure the topics are built correctly), but do we need this in a sample?

cartertinney: I also agree that this is probably more suited to an E2E test, and perhaps the sample should stay the same.

---
In reply to: [308970700](https://github.com/Azure/azure-iot-sdk-python-preview/pull/153#discussion_r308970700) [](ancestors = 308970700)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
BertKleewein: And,  for what it's worth, I've always thought that mqtt_topic_iothub.py deserves to have its own set of unit tests.

---
In reply to: [308970175](https://github.com/Azure/azure-iot-sdk-python-preview/pull/153#discussion_r308970175) [](ancestors = 308970175)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
```diff
@@ -394,13 +450,99 @@ def test_operation_succeeds(self, params, stage, stages_configured_for_both, op)
...
+    {
+        "name": "send telemetry with multiple system and multiple user properties",
+        "stage_type": "device",
+        "op_class": pipeline_ops_iothub.SendD2CMessageOperation,
+        "op_init_kwargs": {
+            "message": create_message_with_system_and_user_properties(
+                fake_message_body, is_multiple=True
+            )
+        },
+        # For more than 1 user property the order could be different, creating 2 different topics
```
BertKleewein: >For more than 1 user property the order could be different, [](start = 10, length = 59)

You should document which version of Python puts these out-of-order.  IIRC, it's 3.4, right?  That will be EOL soon enough.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
olivakar: the tests are now params of the iothub mqtt stages file..a separate test file does not exist 

---
In reply to: [308971540](https://github.com/Azure/azure-iot-sdk-python-preview/pull/153#discussion_r308971540) [](ancestors = 308971540,308970175)

azure-iot-device/samples/legacy-samples/send_telemetry.py
olivakar: I think as long as there are no E2E tests let the sample be as it is

---
In reply to: [308971106](https://github.com/Azure/azure-iot-sdk-python-preview/pull/153#discussion_r308971106) [](ancestors = 308971106,308970700)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
olivakar: actually for me it happened for 2.7 , 3.4 and 3.5 , so better not to mention explicitly.

---
In reply to: [309384147](https://github.com/Azure/azure-iot-sdk-python-preview/pull/153#discussion_r309384147) [](ancestors = 309384147)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
cartertinney: At some point we should probably start testing them separately though. I don't think this PR is the right place for it, but it would be good to do. I'll add a task to the taskboard.

---
In reply to: [309413760](https://github.com/Azure/azure-iot-sdk-python-preview/pull/153#discussion_r309413760) [](ancestors = 309413760,308971540,308970175)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
```diff
@@ -123,6 +134,51 @@ def apply_fake_pipeline_thread(fake_pipeline_thread):
...
+            content_type=fake_content_type,
+        )
+    else:
+        msg = Message(message_content, output_name=fake_output_name, message_id=fake_message_id)
+
+    msg.custom_properties[fake_message_user_property_1_key] = fake_message_user_property_1_value
+    if is_multiple:
+        msg.custom_properties[fake_message_user_property_2_key] = fake_message_user_property_2_value
+    return msg
+
```
cartertinney: It would probably be better for these to be handled as a fixture. But I suspect that's likely going to be too difficult to do with the current model of testing employed in this module.

PR #155, opened 2019-08-01
Correct URL to the samples directory

bf1967:


cartertinney:
Included as part of #170

PR #158, opened 2019-08-05
[REFACTOR] Consistent Fail Op Pattern

cartertinney:
* Added optional `error` parameter to `operation_flow.complete_op()` that sets error
* Changed most fail completions in stages to use this new parameter instead of a manual error set
* Some stages would have required extensive refactoring, so were not changed. 
* Removed ALL instances of raising an error in `PipelineStage._execute_op()` in order to be caught and sent into the wrapping `PipelineStage.run_op()` to do an op completion with error. This path is now ONLY for unexpected exceptions.

### DIFF
--- a/azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
+++ b/azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
@@ -100,6 +100,11 @@ def complete_op(stage, op, error=None):
     (such as a try/except wrapper) which are strongly advised.
     """
     if error:
+        if op.error:
+            # This shouldn't happen. If it does, something is wrong.
+            # You should EITHER set the error on the op, and not pass the error as param,
+            # OR only pass the error as a param.
+            logger.warning("Ovewriting {} error with {} error".format(op.error, error))
         op.error = error
 
     logger.info(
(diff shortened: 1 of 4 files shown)


azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -92,13 +92,21 @@ def pass_op_to_next_stage(stage, op):
 
 
 @pipeline_thread.runs_on_pipeline_thread
-def complete_op(stage, op):
+def complete_op(stage, op, error=None):
```
cartertinney: >error=None [](start = 27, length = 10)

Although I was advocating for the simplicity of something like this, it now does create two different ways of doing the same thing (and creates the edge case highlighted below). Granted, that edge case did technically already exist, and at least there's now a way to log it.

If anyone feels strongly against this, I would be open to returning to the old method of manually adding error before completion. The most important thing in this PR is establishing the unique paths of expected vs unexpected exceptions.

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -92,13 +92,21 @@ def pass_op_to_next_stage(stage, op):
...
     """
     Helper function to complete an operation by calling its callback function thus
     returning the result of the operation back up the pipeline.  This is perferred to
     calling the operation's callback directly as it provides several layers of protection
     (such as a try/except wrapper) which are strongly advised.
     """
+    if error:
+        if op.error:
```
BertKleewein: >if op.error [](start = 8, length = 11)

i agree that we need to narrow this down to a single way of doing things, but, I suggest `if op.error and op.error not == error:` to make the warning a little more explicit.

PR #159, opened 2019-08-06
Additional MQTTTransportStage testing

cartertinney:
* Made tests more granular
* Captured pending connection operations in testing
* Ensured expected failures cause an operation to fail as soon as possible
* Fixed a handler bug where error was not being passed


azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -107,28 +107,91 @@ def transport(mocker):
...
+    return pipeline_ops_mqtt.MQTTUnsubscribeOperation(topic=fake_topic, callback=mocker.MagicMock())
+
+
+@pytest.fixture
+def create_transport(stage, transport, op_set_connection_args):
+    stage.run_op(op_set_connection_args)
+
+
+# TODO: This should be a package level class inherited by all .run_op() tests in all stages
+class RunOpTests(object):
```
cartertinney: By inheriting these common tests around error handling, we can ensure all Run Op tests have them.

Unfortunately, this does mean that these tests end up running first within a test class, which makes reading the requirements a bit odd (they should go at the end of the other tests IMO). I'm sure there's a way this could be altered to run in the correct order via hooks, but I didn't want to waste any additional time on it right now.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -107,28 +107,91 @@ def transport(mocker):
...
+
+
 @pytest.mark.describe(
     "MQTTTransportStage - .run_op() -- called with pipeline_ops_mqtt.SetMQTTConnectionArgsOperation"
 )
-class TestMQTTProviderRunOpWithSetConnectionArgs(object):
+class TestMQTTProviderRunOpWithSetConnectionArgs(RunOpTests):
     @pytest.mark.it("Creates an MQTTTransport object")
     def test_creates_transport(self, stage, transport, op_set_connection_args):
         stage.run_op(op_set_connection_args)
```
cartertinney: >run_op [](start = 14, length = 6)

Turns out I was able to test only on the wrapping `.run_op()` method rather than the inner `._execute_op()` after all. Just took the right approach to writing the tests.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -350,7 +464,7 @@ def test_connected_handler(self, stage, create_transport, mocker):
         stage.transport.on_mqtt_connected_handler()
         assert stage.previous.on_connected.call_count == 1
 
-    @pytest.mark.it("Completes an active connect op when the transport connected event fires")
+    @pytest.mark.it("Completes a pending connect op when the transport connected event fires")
     def test_completes_active_connect_op(self, stage, create_transport, callback):
         op = pipeline_ops_base.ConnectOperation(callback=callback)
         callback.reset_mock()
```
cartertinney: Using a unique `mocker.MagicMock()` would prevent the need for having to do this.

azure-iot-device/tests/common/pipeline/fixtures.py
```diff
@@ -14,9 +14,16 @@
 )
 
 
+# TODO: remove this fixture
```
BertKleewein: YES YES YES.  I am filled with shame

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -28,20 +28,20 @@ class MQTTTransportStage(PipelineStage):
...
-        if self._active_disconnect_op:
-            ops_to_cancel.add(self._active_disconnect_op)
-            self._active_disconnect_op = None
+            ops_to_cancel.append(self._pending_connection_op)
+            self._pending_connection_op = None
+        if self._pending_connection_op:
+            ops_to_cancel.append(self._pending_connection_op)
+            self._pending_connection_op = None
 
         for op in ops_to_cancel:
```
BertKleewein: >   for op in ops_to_cancel: [](start = 5, length = 27)

There will only ever be one op in this list.  when you collapsed _active_connect_op and _active_disconnect_op, you removed the need to have a list here.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -28,20 +28,20 @@ class MQTTTransportStage(PipelineStage):
     """
 
     @pipeline_thread.runs_on_pipeline_thread
-    def _cancel_active_connect_disconnect_ops(self):
+    def _cancel_pending_connection_op(self):
```
BertKleewein: There is some overlap between the code in this layer and the code in my latest PR.  With my latest PR, there should never be 2 pending connection ops issued from above.  
Regardless, I think this code is still OK to have here.  Maybe we can remove it at some point, but I  think we're better off leaving the extra protection right now.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -159,168 +222,219 @@ def test_sets_transport_attribute_on_root(self, stage, transport, op_set_connect
         stage.run_op(op_set_connection_args)
         assert stage.previous.transport == transport.return_value
 
-    @pytest.mark.it("Completes with success if no exception")
+    @pytest.mark.it("Sets the pending connection op tracker to None")
+    def test_pending_conn_op(self, stage, transport, op_set_connection_args):
+        stage.run_op(op_set_connection_args)
+        assert stage._pending_connection_op is None
```
BertKleewein: grumble, grumble, testing a private field, blah blah blah.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -159,168 +222,219 @@ def test_sets_transport_attribute_on_root(self, stage, transport, op_set_connect
...
+    @pytest.mark.it("Cancels any already pending connection operation")
+    @pytest.mark.parametrize(
+        "pending_connection_op",
+        [
+            pytest.param(pipeline_ops_base.ConnectOperation(), id="Pending ConnectOperation"),
+            pytest.param(pipeline_ops_base.ReconnectOperation(), id="Pending ReconnectOperation"),
+            pytest.param(pipeline_ops_base.DisconnectOperation(), id="Pending DisconnectOperation"),
+        ],
+    )
+    def test_pending_operation_cancelled(
```
BertKleewein: >test_pending_operation_cancelled [](start = 8, length = 32)

this is the behavior that matters.  tests_sets_pending_operation and test_pending_conn_op are testing implementation details that support this functionality.  I think this test alone is sufficient and those other 2 are not necessary.

BertKleewein: oh, wait, you have one test that verifies that _pending_connection_op gets set, and another test that verifies that the pending op gets cancelled.  I can see why you broke these up, but I still don't like how the private attribute leaks into the test.  I don't like it, but I don't think it's worth obsessing over either.  

---
In reply to: [311657202](https://github.com/Azure/azure-iot-sdk-python-preview/pull/159#discussion_r311657202) [](ancestors = 311657202)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: That's what I thought. I'll remove this now.

---
In reply to: [311645622](https://github.com/Azure/azure-iot-sdk-python-preview/pull/159#discussion_r311645622) [](ancestors = 311645622)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: Yeah, I think from the perspective of this module, there could be 2 pending ops, so it's important to have this logic that cancels anything already existing before setting a new one. Just to be safe.

---
In reply to: [311646559](https://github.com/Azure/azure-iot-sdk-python-preview/pull/159#discussion_r311646559) [](ancestors = 311646559)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
cartertinney: Yeah, but how else are you going to test this? The entire mechanics of pending operations is handled in private.

At least we're looking at the private field from within the relevant unit. I'd be much more worried if we had to access private fields like this while testing a different unit.

---
In reply to: [311656186](https://github.com/Azure/azure-iot-sdk-python-preview/pull/159#discussion_r311656186) [](ancestors = 311656186)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
cartertinney: I could see the argument to not having a specific test (i.e. requirement) for things related to private attributes (i.e. fold these tests together), but the private attribute does have to be at asserted upon somewhere, otherwise we don't know it's being set/unset correctly.

---
In reply to: [311658611](https://github.com/Azure/azure-iot-sdk-python-preview/pull/159#discussion_r311658611) [](ancestors = 311658611,311657202)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -180,41 +194,49 @@ def _on_mqtt_connection_failure(self, cause):
...
-            op.error = cause
+            op = self._pending_connection_op
+            self._pending_connection_op = None
+
+            if cause:
+                # Only create a ConnnectionDroppedError if there is a cause,
+                # i.e. unexpected disconnect.
+                try:
+                    six.raise_from(errors.ConnectionDroppedError, cause)
+                except errors.ConnectionDroppedError as e:
```
cartertinney: So, this problem:

Native exception chaining was introduced in Python 3, thus isn't compatible with 2. This use of six allows us to use it in 3, while ignoring it in 2. But I wonder if it might be better to use our own implementation of this concept that will be compatible across versions.

i.e. we define our custom exceptions with a `.cause` attribute, and put a space for it in the constructor. Thus we can chain in Python 2.

This will ensure we have consistent tools for investigating errors across versions - it would suck for everything to be harder to investigate in Python 2.

Not suggesting we do it in this PR or anything, this is quite clearly part of a separate PR (or it's own one entirely), but this is something we should take into consideration when putting together our larger error story in the coming weeks.

cartertinney:
>     def _on_mqtt_connection_failure(self, cause):

Cause is mandatory here, correct? That makes sense to me - if a connection fails, there should always a reason. But just wanted to double check.

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py:191 in 7dc749c. [](commit_id = 7dc749c09bc53e00868f2ee797ad155132972858, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -73,44 +67,58 @@ def _execute_op(self, op):
...
+            # Currently, this means that if, say, a connect operation is the pending op and is executed
+            # but another connection op is begins by the time the CONACK is received, the original
+            # operation will be cancelled, but the CONACK for it will still be received, and complete the
+            # NEW operation. This is not desirable, but it is how things currently work.
+
+            # We are however, checking the type, so the CONACK from a cancelled Connect, cannot successfully
+            # complete a Disconnect operation.
+            self._pending_connection_op = None
+
             self.pipeline_root.transport = self.transport
```
cartertinney: >self.pipeline_root.transport = self.transport [](start = 12, length = 45)

By the way, why does this happen? I'm not clear on the purpose.

BertKleewein: removed in my PR

---
In reply to: [311703157](https://github.com/Azure/azure-iot-sdk-python-preview/pull/159#discussion_r311703157) [](ancestors = 311703157)

BertKleewein:
>     def _on_mqtt_connection_failure(self, cause):

yes, mandatory


---
In reply to: [519205017](https://github.com/Azure/azure-iot-sdk-python-preview/pull/159#issuecomment-519205017) [](ancestors = 519205017)

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py:191 in 7dc749c. [](commit_id = 7dc749c09bc53e00868f2ee797ad155132972858, deletion_comment = False)

PR #160, opened 2019-08-07
move ensure connection down and add separate stage to serialize conne…

BertKleewein:
…ct ops

This accomplishes 4 things:
1. more prep work for retry
2. prep work for renewing SAS token.  
3. brings us closer in line with track-2 work.
4. more code under test

The primary goal here is to separate the notions of "this operation needs the transport to be connected" and "connection operations need be serialized" into 2 different stages.  This is because the old code had these 2 things intertwined in a way that made adding new functionality difficult.  In particular, I'm adding support for reconnecting (this change) and I will be adding support to make sure there are no ops in progress before reconnecting (future change).

### DIFF
--- a/azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
+++ b/azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
@@ -163,16 +163,25 @@ class PipelineRootStage(PipelineStage):
     the pipeline exist.  This is also where clients can add event handlers to receive
     events from the pipeline.
 
-    :ivar on_pipeline_event: Handler which can be set by users of the pipeline to
+    :ivar on_pipeline_event_handler: Handler which can be set by users of the pipeline to
       receive PipelineEvent objects.  This is how users receive any "unsolicited"
       events from the pipeline (such as C2D messages).  This function is called with
       a PipelineEvent object every time any such event occurs.
-    :type on_pipeline_event: Function
+    :type on_pipeline_event_handler: Function
+    :ivar on_connected_handler: Handler which can be set by users of the pipeline to
+      receive events every time the underlying transport connects
+    :type on_connected_handler: Function
+    :ivar on_disconnected_handler: Handler which can be set by users of the pipeline to
+      receive events every time the underlying transport disconnects
+    :type on_disconnected_handler: Function
     """
 
     def __init__(self):
         super(PipelineRootStage, self).__init__()
-        self.on_pipeline_event = None
+        self.on_pipeline_event_handler = None
+        self.on_connected_handler = None
+        self.on_disconnected_handler = None
+        self.connected = False
 
     def run_op(self, op):
         op.callback = pipeline_thread.invoke_on_callback_thread_nowait(op.callback)
@@ -209,17 +218,38 @@ class PipelineRootStage(PipelineStage):
     def _handle_pipeline_event(self, event):
         """
         Override of the PipelineEvent handler.  Because this is the root of the pipeline,
-        this function calls the on_pipeline_event handler to pass the event to the
+        this function calls the on_pipeline_event_handler to pass the event to the
         caller.
 
         :param PipelineEvent event: Event to be handled, i.e. returned to the caller
           through the handle_pipeline_event (if provided).
         """
-        if self.on_pipeline_event:
-            # already protected by try/except in handle_pipeline_event()
-            self.on_pipeline_event(event)
+        if self.on_pipeline_event_handler:
+            pipeline_thread.invoke_on_callback_thread_nowait(self.on_pipeline_event_handler)(event)
         else:
             logger.warning("incoming pipeline event with no handler.  dropping.")
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def on_connected(self):
+        logger.info(
+            "{}: on_connected.  on_connected_handler={}".format(
+                self.name, self.on_connected_handler
+            )
+        )
+        self.connected = True
+        if self.on_connected_handler:
+            pipeline_thread.invoke_on_callback_thread_nowait(self.on_connected_handler)()
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def on_disconnected(self):
+        logger.info(
+            "{}: on_disconnected.  on_disconnected_handler={}".format(
+                self.name, self.on_disconnected_handler
+            )
+        )
+        self.connected = False
+        if self.on_disconnected_handler:
+            pipeline_thread.invoke_on_callback_thread_nowait(self.on_disconnected_handler)()
 
 
 class EnsureConnectionStage(PipelineStage):
@@ -228,15 +258,16 @@ class EnsureConnectionStage(PipelineStage):
     it needs to be connected.
     """
 
-    def __init__(self):
-        super(EnsureConnectionStage, self).__init__()
-        self.connected = False
-
     @pipeline_thread.runs_on_pipeline_thread
     def _execute_op(self, op):
         # Any operation that requires a connection can trigger a connection if
         # we're not connected.
-        if op.needs_connection and not self.connected:
+        if op.needs_connection and not self.pipeline_root.connected:
+            logger.info(
+                "{}({}): Op needs connection.  Queueing this op and starting a ConnectionOperation".format(
+                    self.name, op.name
+                )
+            )
             self._do_connect(op)
 
         # Finally, if this stage doesn't need to do anything else with this operation,
@@ -251,14 +282,11 @@ class EnsureConnectionStage(PipelineStage):
         """
         # function that gets called after we're connected.
         @pipeline_thread.runs_on_pipeline_thread
-        def on_connected(op_connect):
-            logger.info(
-                "{}({}): connection is complete: {}".format(self.name, op.name, op_connect.error)
-            )
+        def on_connect_op_complete(op_connect):
             if op_connect.error:
                 logger.info(
-                    "{}({}): completing with failure because of connection failure".format(
-                        self.name, op.name
+                    "{}({}): Connection failed.  Completing with failure because of connection failure: {}".format(
+                        self.name, op.name, op_connect.error
                     )
                 )
                 op.error = op_connect.error
@@ -272,18 +300,8 @@ class EnsureConnectionStage(PipelineStage):
         # call down to the next stage to connect.
         logger.info("{}({}): calling down with Connect operation".format(self.name, op.name))
         operation_flow.pass_op_to_next_stage(
-            self, pipeline_ops_base.ConnectOperation(callback=on_connected)
+            self, pipeline_ops_base.ConnectOperation(callback=on_connect_op_complete)
         )
-
-    @pipeline_thread.runs_on_pipeline_thread
-    def on_connected(self):
-        self.connected = True
-        PipelineStage.on_connected(self)
-
-    @pipeline_thread.runs_on_pipeline_thread
-    def on_disconnected(self):
-        self.connected = False
-        PipelineStage.on_disconnected(self)
 
 
 class SerializeConnectOpsStage(PipelineStage):
@@ -299,7 +317,6 @@ class SerializeConnectOpsStage(PipelineStage):
         super(SerializeConnectOpsStage, self).__init__()
         self.queue = queue.Queue()
         self.blocked = False
-        self.connected = False
 
     @pipeline_thread.runs_on_pipeline_thread
     def _execute_op(self, op):
@@ -307,13 +324,13 @@ class SerializeConnectOpsStage(PipelineStage):
         # to complete), we queue up all operations until after the connect completes.
         if self.blocked:
             logger.info(
-                "{}({}): pipeline is blocked waiting for connect/disconnect/reconnect.  queueing.".format(
+                "{}({}): pipeline is blocked waiting for a prior connect/disconnect/reconnect to complete.  queueing.".format(
                     self.name, op.name
                 )
             )
             self.queue.put_nowait(op)
 
-        elif isinstance(op, pipeline_ops_base.ConnectOperation) and self.connected:
+        elif isinstance(op, pipeline_ops_base.ConnectOperation) and self.pipeline_root.connected:
             logger.info(
                 "{}({}): Transport is already connected.  Completing early".format(
                     self.name, op.name
@@ -321,7 +338,10 @@ class SerializeConnectOpsStage(PipelineStage):
             )
             operation_flow.complete_op(stage=self, op=op)
 
-        elif isinstance(op, pipeline_ops_base.DisconnectOperation) and not self.connected:
+        elif (
+            isinstance(op, pipeline_ops_base.DisconnectOperation)
+            and not self.pipeline_root.connected
+        ):
             logger.info(
                 "{}({}): Transport is already disconnected.  Completing early".format(
                     self.name, op.name
@@ -347,7 +367,7 @@ class SerializeConnectOpsStage(PipelineStage):
                 op.callback = old_callback
                 self._unblock(op, op.error)
                 logger.info(
-                    "{}({}): unblock is comlete.  completing op that caused unblock".format(
+                    "{}({}): unblock is complete.  completing op that caused unblock".format(
                         self.name, op.name
                     )
                 )
@@ -362,18 +382,18 @@ class SerializeConnectOpsStage(PipelineStage):
     @pipeline_thread.runs_on_pipeline_thread
     def _block(self, op):
         """
-        block this stage while we're waiting for the connection to complete.
+        block this stage while we're waiting for the connect/disconnect/reconnect operation to complete.
         """
-        logger.info("{}({}): enabling block".format(self.name, op.name))
+        logger.info("{}({}): blocking".format(self.name, op.name))
         self.blocked = True
 
     @pipeline_thread.runs_on_pipeline_thread
     def _unblock(self, op, error):
         """
-        Unblock this stage after the connection is complete.  This also means
+        Unblock this stage after the connect/disconnect/reconnect operation is complete.  This also means
         releasing all the operations that were queued up.
         """
-        logger.info("{}({}): disabling block and releasing queued ops.".format(self.name, op.name))
+        logger.info("{}({}): unblocking and releasing queued ops.".format(self.name, op.name))
         self.blocked = False
         logger.info(
             "{}({}): processing {} items in queue".format(self.name, op.name, self.queue.qsize())
@@ -401,16 +421,6 @@ class SerializeConnectOpsStage(PipelineStage):
                 )
                 # call run_op directly here so operations go through this stage again (especiall connect/disconnect ops)
                 self.run_op(op_to_release)
-
-    @pipeline_thread.runs_on_pipeline_thread
-    def on_connected(self):
-        self.connected = True
-        PipelineStage.on_connected(self)
-
-    @pipeline_thread.runs_on_pipeline_thread
-    def on_disconnected(self):
-        self.connected = False
-        PipelineStage.on_disconnected(self)
 
 
 class CoordinateRequestAndResponseStage(PipelineStage):
--- a/azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
+++ b/azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
@@ -223,73 +223,141 @@ class PipelineRootStage(PipelineStage):
 
 
 class EnsureConnectionStage(PipelineStage):
-    # TODO: additional documentation and tests for this class are not being implemented because a significant rewriting to support more scenarios is pending
     """
     This stage is responsible for ensuring that the protocol is connected when
-    it needs to be connected, and it's responsible for queueing operations
-    while we're waiting for the connect operation to complete.
-
-    Note: this stage will likely be replaced by a more full-featured stage to handle
-    other "block while we're setting something up" operations, such as subscribing to
-    twin responses.  That is another example where we want to ensure some state and block
-    requests until that state is achieved.
+    it needs to be connected.
     """
 
     def __init__(self):
         super(EnsureConnectionStage, self).__init__()
         self.connected = False
-        self.queue = queue.Queue()
-        self.blocked = False
 
     @pipeline_thread.runs_on_pipeline_thread
     def _execute_op(self, op):
-        # If this stage is currently blocked (because we're waiting for a connection
-        # to complete, we queue up all operations until after the connect completes.
-        if self.blocked:
-            logger.info(
-                "{}({}): pipeline is blocked waiting for connect.  queueing.".format(
-                    self.name, op.name
-                )
-            )
-            self.queue.put_nowait(op)
-
-        # If we get a request to connect, we either complete immediately (if we're already
-        # connected) or we do the connect operation, which is pulled out into a helper
-        # function because starting the connection also means blocking this stage until
-        # the connect is complete.
-        elif isinstance(op, pipeline_ops_base.ConnectOperation):
-            if self.connected:
-                logger.info(
-                    "{}({}): protocol client is already conencted.  completing early.".format(
-                        self.name, op.name
-                    )
-                )
-                operation_flow.complete_op(self, op)
-            else:
-                self._do_connect(op)
-
-        # If we get a request to disconnect, we either complete the request immediately
-        # (if we're already disconencted) or we pass the disconnect request down.
-        elif isinstance(op, pipeline_ops_base.DisconnectOperation):
-            if not self.connected:
-                logger.info(
-                    "{}({}): procotol client is already disconencted.  completing early.".format(
-                        self.name, op.name
-                    )
-                )
-                operation_flow.complete_op(self, op)
-            else:
-                operation_flow.pass_op_to_next_stage(self, op)
-
-        # Any other operation that requires a connection can trigger a connection if
+        # Any operation that requires a connection can trigger a connection if
         # we're not connected.
-        elif op.needs_connection and not self.connected:
+        if op.needs_connection and not self.connected:
             self._do_connect(op)
 
         # Finally, if this stage doesn't need to do anything else with this operation,
         # it just passes it down.
         else:
             operation_flow.pass_op_to_next_stage(self, op)
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _do_connect(self, op):
+        """
+        Start connecting the transport in response to some operation
+        """
+        # function that gets called after we're connected.
+        @pipeline_thread.runs_on_pipeline_thread
+        def on_connected(op_connect):
+            logger.info(
+                "{}({}): connection is complete: {}".format(self.name, op.name, op_connect.error)
+            )
+            if op_connect.error:
+                logger.info(
+                    "{}({}): completing with failure because of connection failure".format(
+                        self.name, op.name
+                    )
+                )
+                op.error = op_connect.error
+                operation_flow.complete_op(stage=self, op=op)
+            else:
+                logger.info(
+                    "{}({}): connection is complete.  Continuing with op".format(self.name, op.name)
+                )
+                operation_flow.pass_op_to_next_stage(stage=self, op=op)
+
+        # call down to the next stage to connect.
+        logger.info("{}({}): calling down with Connect operation".format(self.name, op.name))
+        operation_flow.pass_op_to_next_stage(
+            self, pipeline_ops_base.ConnectOperation(callback=on_connected)
+        )
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def on_connected(self):
+        self.connected = True
+        PipelineStage.on_connected(self)
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def on_disconnected(self):
+        self.connected = False
+        PipelineStage.on_disconnected(self)
+
+
+class SerializeConnectOpsStage(PipelineStage):
+    """
+    This stage is responsible for serializing connect, disconnect, and reconnect ops on
+    the pipeline, such that only a single one of these ops can go past this stage at a
+    time.  This way, we don't have to worry about cases like "what happens if we try to
+    disconnect if we're in the middle of reconnecting."  This stage will wait for the
+    reconnect to complete before letting the disconnect past.
+    """
+
+    def __init__(self):
+        super(SerializeConnectOpsStage, self).__init__()
+        self.queue = queue.Queue()
+        self.blocked = False
+        self.connected = False
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _execute_op(self, op):
+        # If this stage is currently blocked (because we're waiting for a connection, etc,
+        # to complete), we queue up all operations until after the connect completes.
+        if self.blocked:
+            logger.info(
+                "{}({}): pipeline is blocked waiting for connect/disconnect/reconnect.  queueing.".format(
+                    self.name, op.name
+                )
+            )
+            self.queue.put_nowait(op)
+
+        elif isinstance(op, pipeline_ops_base.ConnectOperation) and self.connected:
+            logger.info(
+                "{}({}): Transport is already connected.  Completing early".format(
+                    self.name, op.name
+                )
+            )
+            operation_flow.complete_op(stage=self, op=op)
+
+        elif isinstance(op, pipeline_ops_base.DisconnectOperation) and not self.connected:
+            logger.info(
+                "{}({}): Transport is already disconnected.  Completing early".format(
+                    self.name, op.name
+                )
+            )
+            operation_flow.complete_op(stage=self, op=op)
+
+        elif (
+            isinstance(op, pipeline_ops_base.DisconnectOperation)
+            or isinstance(op, pipeline_ops_base.ConnectOperation)
+            or isinstance(op, pipeline_ops_base.ReconnectOperation)
+        ):
+            self._block(op)
+            old_callback = op.callback
+
+            @pipeline_thread.runs_on_pipeline_thread
+            def on_operation_complete(op):
+                logger.info(
+                    "{}({}): complete.  Unblocking queue with error: {}".format(
+                        self.name, op.name, op.error
+                    )
+                )
+                op.callback = old_callback
+                self._unblock(op, op.error)
+                logger.info(
+                    "{}({}): unblock is comlete.  completing op that caused unblock".format(
+                        self.name, op.name
+                    )
+                )
+                operation_flow.complete_op(stage=self, op=op)
+
+            op.callback = on_operation_complete
+            operation_flow.pass_op_to_next_stage(stage=self, op=op)
+
+        else:
+            operation_flow.pass_op_to_next_stage(stage=self, op=op)
 
     @pipeline_thread.runs_on_pipeline_thread
     def _block(self, op):
@@ -303,17 +371,20 @@ class EnsureConnectionStage(PipelineStage):
     def _unblock(self, op, error):
         """
         Unblock this stage after the connection is complete.  This also means
-        releasing all the queued up operations that we were waiting for the
-        connect operation to complete.
+        releasing all the operations that were queued up.
         """
         logger.info("{}({}): disabling block and releasing queued ops.".format(self.name, op.name))
         self.blocked = False
         logger.info(
             "{}({}): processing {} items in queue".format(self.name, op.name, self.queue.qsize())
         )
-        # loop through our queue and release all the blocked operations
-        while not self.queue.empty():
-            op_to_release = self.queue.get_nowait()
+        # Loop through our queue and release all the blocked operations
+        # Put a new Queue in self.queue because releasing ops might put them back in the
+        # queue, especially if there's a ConnectOperation in the list of ops to release
+        old_queue = self.queue
+        self.queue = queue.Queue()
+        while not old_queue.empty():
+            op_to_release = old_queue.get_nowait()
             if error:
                 # if we're unblocking the queue because something (like a connect operation) failed,
                 # then we fail all of the blocked operations with the same error.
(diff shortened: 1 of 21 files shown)


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
     @pipeline_thread.runs_on_pipeline_thread
     def _execute_op(self, op):
-        # If this stage is currently blocked (because we're waiting for a connection
-        # to complete, we queue up all operations until after the connect completes.
+        # If this stage is currently blocked (because we're waiting for a connection, etc,
+        # to complete), we queue up all operations until after the connect completes.
         if self.blocked:
             logger.info(
-                "{}({}): pipeline is blocked waiting for connect.  queueing.".format(
+                "{}({}): pipeline is blocked waiting for connect/disconnect/reconnect.  queueing.".format(
```
olivakar: >waiting for connect [](start = 45, length = 19)

`waiting for a previous connect/disconnect/reconnect` may sound better.
Is there any possibility of knowing what was the exact previous operation ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
+    @pipeline_thread.runs_on_pipeline_thread
+    def _do_connect(self, op):
+        """
+        Start connecting the transport in response to some operation
+        """
+        # function that gets called after we're connected.
+        @pipeline_thread.runs_on_pipeline_thread
+        def on_connected(op_connect):
+            logger.info(
+                "{}({}): connection is complete: {}".format(self.name, op.name, op_connect.error)
```
cartertinney: >op_connect.error [](start = 80, length = 16)

Why would there be an error on connection complete?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
     def __init__(self):
         super(EnsureConnectionStage, self).__init__()
         self.connected = False
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _execute_op(self, op):
+        # Any operation that requires a connection can trigger a connection if
+        # we're not connected.
+        if op.needs_connection and not self.connected:
+            self._do_connect(op)
```
cartertinney: Maybe a log message here indicating that the operation needs a connection, but is not connected? I don't think that's captured anywhere explicitly in the logs.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
+        """
+        # function that gets called after we're connected.
+        @pipeline_thread.runs_on_pipeline_thread
+        def on_connected(op_connect):
+            logger.info(
+                "{}({}): connection is complete: {}".format(self.name, op.name, op_connect.error)
+            )
+            if op_connect.error:
+                logger.info(
+                    "{}({}): completing with failure because of connection failure".format(
```
cartertinney: And this log seems like it SHOULD be the one to display the error.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Yeah, something like "pipeline is blocked waiting for a prior connect/disconnect/reconnect to complete"

---
In reply to: [311696139](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311696139) [](ancestors = 311696139)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
+    time.  This way, we don't have to worry about cases like "what happens if we try to
+    disconnect if we're in the middle of reconnecting."  This stage will wait for the
+    reconnect to complete before letting the disconnect past.
+    """
+
+    def __init__(self):
+        super(SerializeConnectOpsStage, self).__init__()
         self.queue = queue.Queue()
         self.blocked = False
+        self.connected = False
```
cartertinney: Should this maybe be maintained on the Root rather than on multiple individual stages?

cartertinney:
>         block this stage while we're waiting for the connection to complete.

operation, not connection

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py:365 in 5c3c8b4. [](commit_id = 5c3c8b41fdf51d069cd7f977c4096cc409fd8e8f, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
                 logger.info(
-                    "{}({}): procotol client is already disconencted.  completing early.".format(
+                    "{}({}): unblock is comlete.  completing op that caused unblock".format(
                         self.name, op.name
                     )
                 )
-                operation_flow.complete_op(self, op)
-            else:
-                operation_flow.pass_op_to_next_stage(self, op)
+                operation_flow.complete_op(stage=self, op=op)
```
cartertinney: >operation_flow.complete_op(stage=self, op=op) [](start = 16, length = 45)

Doesn't this mean that the operation gets "completed" twice?

It gets completed below, thus triggering this callback, and then completes again here in the callback, in order to complete the ORIGINAL callback. This seems odd, will generate potentially confusing logs, and could result in issues if "complete_op" ever does more than just triggering a callback.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -303,17 +371,20 @@ def _block(self, op):
...
         )
-        # loop through our queue and release all the blocked operations
-        while not self.queue.empty():
-            op_to_release = self.queue.get_nowait()
+        # Loop through our queue and release all the blocked operations
+        # Put a new Queue in self.queue because releasing ops might put them back in the
+        # queue, especially if there's a ConnectOperation in the list of ops to release
+        old_queue = self.queue
+        self.queue = queue.Queue()
+        while not old_queue.empty():
```
cartertinney: >while not old_queue.empty(): [](start = 8, length = 28)

What happens if new ops are added to the new queue while we're still resolving the ops old queue? Or does the threading model protect us from this? It would seem to me that the ops in the old queue should ALL be resolved before anything in the new queue, since they predate it.

e.g. 5 Publish ops are queued while waiting for a ConnectOp to resolve. The connect is resolved, and we begin resolving the 5 publishes from the old_queue, while a new empty queue is made. While we're resolving Publish #2, a DisconnectOp is sent down the pipe. The stage is unblocked already, since it's that happens before the resolution of all the old queued ops, so does the DisconnectOp block on our remaining Publishes or not?

If it does, cool. If it doesn't, then what happens to these Publish ops that are caught in limbo? And what happens if a ConnectOp is queued immediately after the DisconnectOp?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -303,17 +371,20 @@ def _block(self, op):
...
+            op_to_release = old_queue.get_nowait()
             if error:
                 # if we're unblocking the queue because something (like a connect operation) failed,
                 # then we fail all of the blocked operations with the same error.
```
cartertinney: Is this desired? Shouldn't we retry here? Or will that be part of a future PR?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -325,50 +396,12 @@ def _unblock(self, op, error):
                 op_to_release.error = error
```
cartertinney: >op_to_release.error = error [](start = 16, length = 27)

Doesn't need to be done in this PR, but as part of a coherent error story, we might want to chain a different error from the original to indicate this is failing indirectly due to an error in a different op

cartertinney:
>         PipelineStage.on_disconnected(self)

These are unused

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py:413 in 5c3c8b4. [](commit_id = 5c3c8b41fdf51d069cd7f977c4096cc409fd8e8f, deletion_comment = False)

cartertinney:
I've left some comments on implementation, once I'm clear on a few questions re: expected behavior that were raised there, I will review the test code.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
+            operation_flow.pass_op_to_next_stage(self, op)
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _do_connect(self, op):
+        """
+        Start connecting the transport in response to some operation
+        """
+        # function that gets called after we're connected.
+        @pipeline_thread.runs_on_pipeline_thread
+        def on_connected(op_connect):
```
olivakar: I am guessing this will be called even if there is a failure. so may be this method can be renamed to something that handles post connection stuff. currently this feels to be a successful connection call back.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -303,17 +371,20 @@ def _block(self, op):
     def _unblock(self, op, error):
         """
         Unblock this stage after the connection is complete.  This also means
-        releasing all the queued up operations that we were waiting for the
-        connect operation to complete.
+        releasing all the operations that were queued up.
         """
         logger.info("{}({}): disabling block and releasing queued ops.".format(self.name, op.name))
```
olivakar: >disabling block [](start = 29, length = 15)

`unblocking` will sound better i think

cartertinney: +1

---
In reply to: [311821517](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311821517) [](ancestors = 311821517)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -223,73 +223,141 @@ def _handle_pipeline_event(self, event):
...
-
-        # If we get a request to disconnect, we either complete the request immediately
-        # (if we're already disconencted) or we pass the disconnect request down.
-        elif isinstance(op, pipeline_ops_base.DisconnectOperation):
-            if not self.connected:
+                op.callback = old_callback
+                self._unblock(op, op.error)
                 logger.info(
-                    "{}({}): procotol client is already disconencted.  completing early.".format(
+                    "{}({}): unblock is comlete.  completing op that caused unblock".format(
```
olivakar: >comlete [](start = 40, length = 7)

spell

olivakar:
>         logger.info("{}({}): enabling block".format(self.name, op.name))

similar to below `blocking` would sound nicer

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py:367 in 5c3c8b4. [](commit_id = 5c3c8b41fdf51d069cd7f977c4096cc409fd8e8f, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: If the connection failed.  The operation is done, but it failed.  I think i can re-phrase this message better.

---
In reply to: [311697780](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311697780) [](ancestors = 311697780)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: Yes, I was debating about that.  Good idea.

---
In reply to: [311702803](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311702803) [](ancestors = 311702803)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: That's exactly what it means, and you're right.  The other option is to copy every single operation that goes down the pipeline.  I'll have to think about this more, I didn't even consider this as a point of confusion because it's a fairly common practice in driver development (though with different terms).

---
In reply to: [311710166](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311710166) [](ancestors = 311710166)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: retry needs to be handled in a different stage and is part of a different PR.

---
In reply to: [311715312](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311715312) [](ancestors = 311715312)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: The answer to this depends on what you mean by "resolved."  That is not a term that we've used here.  We can only ensure that ops get passed down to the next lower level.  We never wait for ops to complete (e.g. ACK comes back) unless we absolutely must (such as waiting for a CONACK before publishing).

Here are a few examples that I hope are self-explanatory.

[http://www.plantuml.com/plantuml/svg/0/pLPDRy8m3Br7wd_maKdZG7P3wiHbFoIafeqMqgf91zkldzFIxr8oYSD0Wg8ysy_FBnO7wqJfghEEM1p3nrMS2usmsuCe2f2vCIYT1M4IAAgZLZP32noSEjFt9LE4lGZbb4dXb9Tj0BYS7dqcXn374QkA0alOleF5KWcjVX3sK5TkGoBM7otYZZgNdrGdOYQdhALACxS7hKxeoJJ7a9j19CMb2CeaIWfNqqK3BXEkFR5D52OH4zAfBqBruHD2EhzG1pFIAutTzDhkWb0yOas92QfLr5PI8YQUwrsXAbFt4x44_zZFW6i0YazGJNkBNBkUManNAGZ5XwYHGpcH4ZHtONlJlW9U9HRTUE0abBQr0PzimNzlbxhp9rbbFhgbxFFHhQ4cF4UZVLiSRKONNljBN--QoX1hkkrIQDql0J_fjCohejbUT6RRIfkNY06zGcpWSS5MyE0b2MT9xbW2JJ8dVHDnqqd-cFt0Fq4fUNXAFahPzA75DtsE9Av_7OLtbKySkivMrUvywb7BNdqWz42NG2HVGAwik5Lf5jVTI9tLnHYKcmVdvhFS_dsW9_e3y0i0](http://www.plantuml.com/plantuml/svg/0/pLPDRy8m3Br7wd_maKdZG7P3wiHbFoIafeqMqgf91zkldzFIxr8oYSD0Wg8ysy_FBnO7wqJfghEEM1p3nrMS2usmsuCe2f2vCIYT1M4IAAgZLZP32noSEjFt9LE4lGZbb4dXb9Tj0BYS7dqcXn374QkA0alOleF5KWcjVX3sK5TkGoBM7otYZZgNdrGdOYQdhALACxS7hKxeoJJ7a9j19CMb2CeaIWfNqqK3BXEkFR5D52OH4zAfBqBruHD2EhzG1pFIAutTzDhkWb0yOas92QfLr5PI8YQUwrsXAbFt4x44_zZFW6i0YazGJNkBNBkUManNAGZ5XwYHGpcH4ZHtONlJlW9U9HRTUE0abBQr0PzimNzlbxhp9rbbFhgbxFFHhQ4cF4UZVLiSRKONNljBN--QoX1hkkrIQDql0J_fjCohejbUT6RRIfkNY06zGcpWSS5MyE0b2MT9xbW2JJ8dVHDnqqd-cFt0Fq4fUNXAFahPzA75DtsE9Av_7OLtbKySkivMrUvywb7BNdqWz42NG2HVGAwik5Lf5jVTI9tLnHYKcmVdvhFS_dsW9_e3y0i0)


[http://www.plantuml.com/plantuml/svg/0/xLV1ReCm3BqZyG_MTcaF75PjErHgLKtQ_YCbRecQ9eY4iUthvy1e8fgoxZ2fqmedmB6Vd_CUi3AMLRRUopHPBE3vZUrBYN0_XrmhXRdDbcMzbi8KsP8BCxm94wivUuSi0oOb-0W3J76eq5zM21jTvSYdi5YaILsMM46s18ELO59y8CoXnT9NJXF_A1gtbZhV2RLD4vPRySei-g31OgaRgdPKO0H4t_NTQ9Hl-AnaipJfa0oYldA0gVCSaPiqKPgwgSIsi7EGOeDMxB5h6BI2cpRB3GZ5HKvz6sWAL60hfaofgQ0mVKxa0JtjiX5UQzMMI1EEFwGqeEkRgOMK_TQ0mnfHmdr0pxdZAUZwzvcQ3QAwxNCAsg6BFkn0E6oOaACXiqE8EmELFvojygPW5XhQaKWoVuU6AUiqhxJAEb67qhyA_0o1_mDDXeiR9gpp__OrKIBvbjg2BWrOJUIutMWnbX9kNrwUw0WTCKs03kFUBHmcfL2xQSGZVGe2liNMAvrLTgixltIx-kkfQI4wL0xbIOHq7noI4tFsgepRIgIlg0b4yrmDR-XcWPSzKZb6ysV6snkiQt8jOGl_8NArfZzcJOV9NQupNPGJ17mTgpkAr6iZg701lX35DM8Bexwsei7IJyud](http://www.plantuml.com/plantuml/svg/0/xLV1ReCm3BqZyG_MTcaF75PjErHgLKtQ_YCbRecQ9eY4iUthvy1e8fgoxZ2fqmedmB6Vd_CUi3AMLRRUopHPBE3vZUrBYN0_XrmhXRdDbcMzbi8KsP8BCxm94wivUuSi0oOb-0W3J76eq5zM21jTvSYdi5YaILsMM46s18ELO59y8CoXnT9NJXF_A1gtbZhV2RLD4vPRySei-g31OgaRgdPKO0H4t_NTQ9Hl-AnaipJfa0oYldA0gVCSaPiqKPgwgSIsi7EGOeDMxB5h6BI2cpRB3GZ5HKvz6sWAL60hfaofgQ0mVKxa0JtjiX5UQzMMI1EEFwGqeEkRgOMK_TQ0mnfHmdr0pxdZAUZwzvcQ3QAwxNCAsg6BFkn0E6oOaACXiqE8EmELFvojygPW5XhQaKWoVuU6AUiqhxJAEb67qhyA_0o1_mDDXeiR9gpp__OrKIBvbjg2BWrOJUIutMWnbX9kNrwUw0WTCKs03kFUBHmcfL2xQSGZVGe2liNMAvrLTgixltIx-kkfQI4wL0xbIOHq7noI4tFsgepRIgIlg0b4yrmDR-XcWPSzKZb6ysV6snkiQt8jOGl_8NArfZzcJOV9NQupNPGJ17mTgpkAr6iZg701lX35DM8Bexwsei7IJyud)



---
In reply to: [311714638](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311714638) [](ancestors = 311714638)

BertKleewein: And, for your last question about calling disconnect then immediately calling connect:

[http://www.plantuml.com/plantuml/svg/0/lPFBRi8m44NtaV8FiwGBraX8GTdm8OCz0ek9RMAxBVrw1imZXK1JXCZ2Y-Z8z_gCNOU8RKmDbqLLmVmR6yy4aocWzw2GEO0sGJbhIKL0go6ko89f6j868_7sI9sGgYgBPUkIxzrXz54M85_odbeOpo1GQv3D3y4K3bKwS4OlH2--OAVMnYupZ2gQJwbr0NisOVSbuNrvPFLj--7LX_K-ozIQHAU3JQ94ly9pGyQjh3T3sFzOe5hdc8nUHNJea9GYagG_MYLxC9afJOD4FclWHPEx7vPdTC_3Gz7_iFmEnQUJ7BFambmVf9PLtkyE](http://www.plantuml.com/plantuml/svg/0/lPFBRi8m44NtaV8FiwGBraX8GTdm8OCz0ek9RMAxBVrw1imZXK1JXCZ2Y-Z8z_gCNOU8RKmDbqLLmVmR6yy4aocWzw2GEO0sGJbhIKL0go6ko89f6j868_7sI9sGgYgBPUkIxzrXz54M85_odbeOpo1GQv3D3y4K3bKwS4OlH2--OAVMnYupZ2gQJwbr0NisOVSbuNrvPFLj--7LX_K-ozIQHAU3JQ94ly9pGyQjh3T3sFzOe5hdc8nUHNJea9GYagG_MYLxC9afJOD4FclWHPEx7vPdTC_3Gz7_iFmEnQUJ7BFambmVf9PLtkyE)


---
In reply to: [312164161](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312164161) [](ancestors = 312164161,311714638)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: changed.  I could add code to keep track of the current operation, but I don't want to do that just for logging because then I need to make sure it's always correct which means a whole bunch of tests and I just don't want to go there.  

---
In reply to: [311702297](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r311702297) [](ancestors = 311702297,311696139)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
```diff
@@ -131,6 +131,7 @@ def apply_fake_pipeline_thread(fake_pipeline_thread):
     handled_ops=ops_handled_by_this_stage,
     all_events=all_common_events + all_iothub_events,
     handled_events=events_handled_by_this_stage,
+    extra_initializer_defaults={"feature_to_topic": dict},
```
cartertinney: Use of `{}` is preferred to `dict()` which is preferable to `dict`

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning_mqtt.py
```diff
@@ -79,6 +79,7 @@ def apply_fake_pipeline_thread(fake_pipeline_thread):
     handled_ops=ops_handled_by_this_stage,
     all_events=all_common_events + all_provisioning_events,
     handled_events=events_handled_by_this_stage,
+    extra_initializer_defaults={"action_to_topic": dict},
```
cartertinney: As in the other comment, `{}` is preferred

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
cartertinney: Given that this dict maps features to topics, for use in other functions, and those tests test for correctness of topics chosen, I feel like this should probably be a private attribute that isn't tested. There's no real requirements value in proving that a mapping dictionary exists called `feature_to_topic`

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning_mqtt.py
cartertinney: Same as the comment in `test_pipeline_stages_iothub_mqtt.py`

cartertinney:
>         operation_flow.pass_op_to_next_stage(self, op)

This is dead code as far as I can tell, it seems totally unused.

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py:198 in 4abece0. [](commit_id = 4abece0e32401edbad0d0ed7f5a57010d0677ddd, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -209,111 +218,193 @@ def append_stage(self, new_next_stage):
     def _handle_pipeline_event(self, event):
```
cartertinney: >def _handle_pipeline_event(self, event): [](start = 4, length = 40)

Is this also dead? I don't see it being used anywhere

cartertinney:
> @pytest.mark.it("Calls operation callback in callback thread")

Specify when this occurs (i.e. when running an operation)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:66 in 51aba2e. [](commit_id = 51aba2e27964837c9b29f0b434ada6683c9c39a8, deletion_comment = False)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+    },
+]
+
+
+@pytest.mark.parametrize(
+    "params",
+    ops_that_cause_connection,
+    ids=[x["op_class"].__name__ for x in ops_that_cause_connection],
+)
+@pytest.mark.describe("EnsureConnectionStage - .run_op() -- called with connect operation")
```
cartertinney: >connect operation [](start = 72, length = 17)

This seems to contradict the above parametrization.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+    def test_sends_connect(self, params, op, stage):
+        stage.pipeline_root.connected = False
+
+        stage.run_op(op)
+
+        assert stage.next.run_op.call_count == 1
+        assert isinstance(stage.next.run_op.call_args[0][0], pipeline_ops_base.ConnectOperation)
+
+    @pytest.mark.it(
+        "Calls the op's callback with the error from the ConnectOperation is that operation fails"
```
cartertinney: >is [](start = 74, length = 2)

if

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+        "op_class": pipeline_ops_mqtt.MQTTUnsubscribeOperation,
+        "op_init_kwargs": {"topic": fake_topic},
+    },
+]
+
+
+@pytest.mark.parametrize(
+    "params",
+    ops_that_cause_connection,
+    ids=[x["op_class"].__name__ for x in ops_that_cause_connection],
```
cartertinney: What about all the other operations that could pass through this stage?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+    @pytest.fixture
+    def stage(self, mocker):
+        stage = make_mock_stage(
+            mocker=mocker, stage_to_make=pipeline_stages_base.EnsureConnectionStage
+        )
+        stage.next.run_op = mocker.MagicMock()
+        return stage
+
+    @pytest.mark.it("Passes the operation down the pipline when the transport is already connected")
+    def test_operation_alrady_connected(self, params, op, stage):
```
cartertinney: This test class is kind of tricky, because technically this method has four units/paths:

1) Called with an op that needs connection, and we are not connected
2) Called with an op that needs connection, and we are connected
3) Called with an op that does not need connection, and we are not connected
4) Called with an op that does not need connection, and we are connected

In practice, units 2-4 are all the exact same path, which only requires a single test, so for simplicity you've thrown that single test in with all the others from unit 1. This can be a little confusing while reading the output, but honestly, it's not the end of the world, since we DO have the coverage.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+
+        stage.run_op(op)
+
+        assert stage.next.run_op.call_count == 1
+        assert stage.next.run_op.call_args[0][0] == op
+
+    @pytest.mark.it(
+        "Sends a ConnectOperation instead of the op down the pipeline if the transport is not connected"
+    )
+    def test_sends_connect(self, params, op, stage):
```
cartertinney: >params [](start = 33, length = 6)

If you put the parametrization directly on the fixture rather than the whole class, you wouldn't need to include the `params` args in all of these tests

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+    @pytest.fixture
+    def stage(self, mocker):
+        stage = make_mock_stage(
+            mocker=mocker, stage_to_make=pipeline_stages_base.SerializeConnectOpsStage
+        )
+        stage.next.run_op = mocker.MagicMock()
+        return stage
+
+    @pytest.fixture
+    def connect_op(self, mocker, params):
```
cartertinney: I am honestly unclear what the `params` argument refers to

cartertinney: this should probably also be `connection_op` rather than `connect_op` since it could apply to not just `ConnectOperation`, but also `DisconnectOperation` and `ReconnectOperation`.

although admittedly, that terminology isn't much better (especially if we end up eventually turning the transport into something called a "connection") as per the Track 2 stuff.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+    {"op_class": pipeline_ops_base.DisconnectOperation, "connected_flag_required_to_run": True},
+    {"op_class": pipeline_ops_base.ReconnectOperation, "connected_flag_required_to_run": True},
+]
+
+
+class FakeOperation(pipeline_ops_base.PipelineOperation):
+    pass
+
+
+@pytest.mark.describe("SerializeConnectOpsStage - .run_op() -- called with connect operation")
```
cartertinney: >connect [](start = 75, length = 7)

I'm guessing this also refers to "connection operation", i.e. operations that are related to management of the connection, rather than an actual `ConnectOperation`?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+        "Immediately completes a DisconnectOperation if the transport is already disconnected"
+    )
+    def test_disconnect_while_disconnected(self, stage, mocker):
+        op = pipeline_ops_base.DisconnectOperation(callback=mocker.MagicMock())
+        stage.pipeline_root.connected = False
+        stage.run_op(op)
+        assert_callback_succeeded(op=op)
+
+    @pytest.mark.it(
+        "Immediately passes the operation down if an operation is not alrady being serialized"
```
cartertinney: >serialized [](start = 83, length = 10)

Probably just my lack of knowledge, but I'm unclear exactly what `serialized` means in this context. I get what this is referring to - an already running connection operation. blocking all other operations, I just don't see how that means the operation is "being serialized".

Also probably worth specifying "another connection operation is not already being serialized", since only connection operations can block.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+    def test_passes_op_when_not_blocked(self, stage, mocker, fake_op):
+        stage.run_op(fake_op)
+        assert stage.next.run_op.call_count == 1
+        assert stage.next.run_op.call_args[0][0] == fake_op
+
+    @pytest.mark.parametrize(
+        "params", connect_ops, ids=[x["op_class"].__name__ for x in connect_ops]
+    )
+    @pytest.mark.it(
+        "Does not pass the operation down if a different operation is currently being serialized"
```
cartertinney: "Does not immediately pass the operation down", since, it will, of course, eventually get passed down.

Arguably this, and the next test could be folded together for a greater degree of clarity.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+        stage.run_op(fake_op)
+
+        assert stage.next.run_op.call_count == 1
+        assert stage.next.run_op.call_args[0][0] == connect_op
+
+    @pytest.mark.parametrize(
+        "params", connect_ops, ids=[x["op_class"].__name__ for x in connect_ops]
+    )
+    @pytest.mark.it(
+        "Waits for the currently serialized operation to complete before passing the op down"
```
cartertinney: complete successfully

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+            stage.run_op(op)
+
+        connect_op.error = fake_exception
+        operation_flow.complete_op(stage=stage.next, op=connect_op)
+
+        for op in fake_ops:
+            assert_callback_failed(op=op, error=fake_exception)
+
+    @pytest.mark.it(
+        "Does not pass down operations in the queue if an operation in the queue gets serialized"
```
cartertinney: >Does not pass down operations in the queue if an operation in the queue gets serialized [](start = 9, length = 87)

This is a hard one to phrase!

Does not pass down further operations in the queue once an operation in the queue begins serialization?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+                    "first_connect_op": pipeline_ops_base.ReconnectOperation,
+                    "mid_connect_flag": True,
+                    "second_connect_op": pipeline_ops_base.ConnectOperation,
+                },
+                id="Reconnect followed by Connect",
+            ),
+        ],
+    )
+    @pytest.mark.it(
+        "Immediately completes a second op which was waiting for a first op that succeeded"
```
cartertinney: >"Immediately completes a second op which was waiting for a first op that succeeded" [](start = 8, length = 83)

Wow, this one is __really__ hard too. We have to capture the relationship between the operations, and their use of the queue

Maybe "Immediately completes a queued operation whose desired result has already been completed by a previous operation"?

There should also probably be a test cate with intermediate operations i.e. Connect->Publish->Connect, since this doesn't only apply to ops in direct succession

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -107,12 +99,441 @@ def mock_execute_op(self, op):
...
+        assert stage.next.run_op.call_count == 1
+        assert stage.next.run_op.call_args[0][0] == first_connect_op
+
+        # complete first_connect_op
+        stage.pipeline_root.connected = params["mid_connect_flag"]
+        operation_flow.complete_op(stage=stage.next, op=first_connect_op)
+
+        # second connect_op should be completed without having been passed down.
+        assert stage.next.run_op.call_count == 1
+        assert_callback_succeeded(op=second_connect_op)
```
cartertinney: I'd also like to see a test case for the case we discussed offline the other day:

1) Connection operations triggers op serialization and starts block
2) Several operations pile up in the queue
3) Connection operation completes, queue is unblocked
4) While in the process of emptying the queue, a new operation is sent down

As per our discussion, you said this new operation should be blocked, or not allowed to be sent down, or something to that effect, until the queue has successfully emptied.

cartertinney:
> # Licensed under the MIT License. See License.txt in the project root for

Given that this test module needs extensive revision, I'm not expecting complete and total test coverage here. The tests you've added seem appropriately thorough, given the existing testing model, that's sufficient for me at this time.

---
Refers to: azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py:3 in 51aba2e. [](commit_id = 51aba2e27964837c9b29f0b434ada6683c9c39a8, deletion_comment = False)

cartertinney:
I have now reviewed the testing code. There's a few missing cases that I'd like to see coverage for.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: yes I would.  The runner requires params on the test function even if it's not used.  Otherwise it assume you forgot to include it and fails with error.

---
In reply to: [312670685](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312670685) [](ancestors = 312670685)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
BertKleewein: out of scope.  If you believe this is important, open a task in TFS.

---
In reply to: [312593187](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312593187) [](ancestors = 312593187)

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning_mqtt.py
BertKleewein: same as the response in test_pipeline_stages_iothub_mqtt.py

---
In reply to: [312593655](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312593655) [](ancestors = 312593655)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: Yeah, what about them?  What do you mean?  add_base_pipeline_stage_tests() makes sure they pass through.

---
In reply to: [312665049](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312665049) [](ancestors = 312665049)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: Added in TFS task 5117988

---
In reply to: [312667014](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312667014) [](ancestors = 312667014)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: Added in TFS task 5117988

---
In reply to: [312678963](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312678963) [](ancestors = 312678963)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: Added in TFS task 5117988

---
In reply to: [312679508](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312679508) [](ancestors = 312679508)

BertKleewein:
>         operation_flow.pass_op_to_next_stage(self, op)

PIpelineRootStage.run_op() calls PipelineStage.run_op() which calls self._execute_op()

---
In reply to: [520074698](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#issuecomment-520074698) [](ancestors = 520074698)

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py:198 in 4abece0. [](commit_id = 4abece0e32401edbad0d0ed7f5a57010d0677ddd, deletion_comment = False)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: yes, this is confusing.  I changed all the mark.it text to talk about operations that are blocking the stage.  I think this is much clearer.

---
In reply to: [312674162](https://github.com/Azure/azure-iot-sdk-python-preview/pull/160#discussion_r312674162) [](ancestors = 312674162)

PR #161, opened 2019-08-08
clarifying service SDK is not supported yet

gregman-msft:
some confusion from myself and others trying to figure out if DPS client is supported by v2, so this change may help clarify


gregman-msft:
@olivakar

olivakar:
We will change the read me to clarify this point however we can not change the name as the name we landed on was `Azure IoT Hub Provisioning SDK` for the service part. I will close this PR.

PR #164, opened 2019-08-14
[INFRASTRUCTURE] Override inherited methods with shim implementations

cartertinney:
* Dynamically patch all iothub and provisioning clients with overrides to inherited methods, which simply call into the parent method.
* New override methods contain the docs of the parent class
* This allows us to keep our codebase the same, while providing all the necessary implementation on child classes for automatic doc generation
* This does NOT interfere with static code analysis tools (e.g. intellisense) because it results in no static changes to the class definitions

EDIT: This is currently only implemented for Python 3.5+, still need to fix this for compat scenarios

### DIFF
--- a/azure-iot-device/azure/iot/device/patch.py
+++ b/azure-iot-device/azure/iot/device/patch.py
@@ -45,12 +45,11 @@ def add_shims_for_inherited_methods(target_class):
 
     # Import the class we're adding methods to, so that functions defined in this scope can use super()
     class_module = inspect.getmodule(target_class)
-    exec(
-        "from {module} import {target_class} as {alias}".format(
-            module=class_module.__name__, target_class=target_class.__name__, alias=classname_alias
-        ),
-        globals(),
+    import_cmdstr = "from {module} import {target_class} as {alias}".format(
+        module=class_module.__name__, target_class=target_class.__name__, alias=classname_alias
     )
+    logger.info("exec: " + import_cmdstr)
+    exec(import_cmdstr, globals())
 
     for method in all_methods:
         method_name = method[0]
@@ -99,7 +98,7 @@ def add_shims_for_inherited_methods(target_class):
                 ret_syntax = "return"
 
             # Dynamically define a new function, with the same name, that invokes the method of the parent class
-            new_fn_str = "{def_syntax} {method_name}{signature}: {ret_syntax} super({leaf_class}, {object_or_type}).{method_name}{invocation}".format(
+            fn_def_cmdstr = "{def_syntax} {method_name}{signature}: {ret_syntax} super({leaf_class}, {object_or_type}).{method_name}{invocation}".format(
                 def_syntax=def_syntax,
                 method_name=method_name,
                 signature=str(method_sig),
@@ -108,26 +107,28 @@ def add_shims_for_inherited_methods(target_class):
                 object_or_type=obj_or_type,
                 invocation=str(invoke_params),
             )
-            logger.info("Adding shim: " + new_fn_str)
-            exec(new_fn_str, globals())
+            logger.info("exec: " + fn_def_cmdstr)
+            exec(fn_def_cmdstr, globals())
 
             # Add function to leaf/child class as a method.
-            exec("{method_name}.__doc__ = method_obj.__doc__".format(method_name=method_name))
+            set_doc_cmdstr = "{method_name}.__doc__ = method_obj.__doc__".format(
+                method_name=method_name
+            )
+            logger.info("exec: " + set_doc_cmdstr)
+            exec(set_doc_cmdstr)
 
             # Add as a classmethod if the method is a classmethod
             if inspect.ismethod(method_obj):
-                exec(
-                    "setattr(target_class, method_name, classmethod({method_name}))".format(
-                        method_name=method_name
-                    )
+                attach_shim_cmdstr = "setattr(target_class, method_name, classmethod({method_name}))".format(
+                    method_name=method_name
                 )
             # Add regularly if the method is an instance method
             else:
-                exec(
-                    "setattr(target_class, method_name, {method_name})".format(
-                        method_name=method_name
-                    )
+                attach_shim_cmdstr = "setattr(target_class, method_name, {method_name})".format(
+                    method_name=method_name
                 )
+            logger.info("exec: " + attach_shim_cmdstr)
+            exec(attach_shim_cmdstr)
 
     # NOTE: the __qualname__ attributes of these new shim methods point to THIS module, not the class itself.
     # This shouldn't matter, but in case it does, I am documenting that fact here.
--- a/azure-iot-device/azure/iot/device/patch.py
+++ b/azure-iot-device/azure/iot/device/patch.py
@@ -1,0 +1,119 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module provides patches used to dynamically modify items from the libraries"""
+
+import inspect
+
+
+def add_shims_for_inherited_methods(target_class):
+    """Dynamically add overriding shim methods for all public inherited methods on a child class,
+    which simply call into the parent class implementation of the same method.
+
+    These shim methods will include the same docstrings as the method from the parent class.
+
+    :param target_class: The class to add shim methods to
+    """
+
+    # Depending on how the method was defined, it could be either a function or a method.
+    # Thus we need to find the union of the two sets.
+    # Here instance methods are considered functions because they are not yet bound to an instance
+    # of the class. Classmethods on the other hand, are already bound, and show up as methods.
+    # It also is worth noting that async functions/methods ARE picked up by this introspection.
+    class_functions = inspect.getmembers(target_class, predicate=inspect.isfunction)
+    class_methods = inspect.getmembers(target_class, predicate=inspect.ismethod)
+    all_methods = class_functions + class_methods
+
+    # Import the class we're adding methods to, so that functions defined in this scope can use super()
+    class_module = inspect.getmodule(target_class)
+    exec(
+        "from {module} import {target_class}".format(
+            module=class_module.__name__, target_class=target_class.__name__
+        ),
+        globals(),
+    )
+
+    for method in all_methods:
+        method_name = method[0]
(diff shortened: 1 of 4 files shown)


azure-iot-device/azure/iot/device/patch.py
```diff
@@ -0,0 +1,116 @@
...
+import six
+
+
+def add_shims_for_inherited_methods(target_class):
+    """Dynamically add overriding shim methods for all public inherited methods on a child class,
+    which simply call into the parent class implementation of the same method.
+
+    These shim methods will include the same docstrings as the method from the parent class.
+
+    :param target_class: The class to add shim methods to
```
olivakar: >target_class [](start = 11, length = 12)

also note that is most probably the child class right ?

azure-iot-device/azure/iot/device/patch.py
```diff
@@ -0,0 +1,116 @@
...
+        "from {module} import {target_class}".format(
+            module=class_module.__name__, target_class=target_class.__name__
+        ),
+        globals(),
+    )
+
+    for method in all_methods:
+        method_name = method[0]
+        method_obj = method[1]
+        originating_module = inspect.getmodule(method_obj)
```
olivakar: >originating_module [](start = 8, length = 18)

what is the difference between this module and the class module ?

azure-iot-device/azure/iot/device/patch.py
```diff
@@ -0,0 +1,133 @@
...
+            module=class_module.__name__, target_class=target_class.__name__, alias=classname_alias
+        ),
+        globals(),
+    )
+
+    for method in all_methods:
+        method_name = method[0]
+        method_obj = method[1]
+        # We can index on 0 here because the list comprehension will always be exactly 1 element
+        method_attribute = [att for att in class_attributes if att.name == method_name][0]
```
BertKleewein: Nice trick.  makes me wonder if there's a clearer way to do this.

BertKleewein: Too bad there's not a builtin for this.  I've also found next(att for att in class_attributes if att.name == mathod_name)) but that doesn't add any clarity as far as I'm concerned

---
In reply to: [315324838](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315324838) [](ancestors = 315324838)

cartertinney: Yeah, I tried to find something more elegant, but unfortunately, there doesn't seem to be anything. I think the core problem is that there's no real way to guarantee that no more than one item in the list will meet the select condition.

The only alternative I can think of involves a for loop that checks the condition issuing a `break` after the correct attribute is found.

```python
method_attribute = None
for att in class_attributes:
    if att.name == method_name:
        method_attribute = att
        break
```

This is I suppose, more correct, but takes up more lines. Perhaps it's clearer?

---
In reply to: [315326449](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315326449) [](ancestors = 315326449,315324838)

azure-iot-device/azure/iot/device/patch.py
```diff
@@ -0,0 +1,133 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module provides patches used to dynamically modify items from the libraries"""
```
BertKleewein: damn, this is a really complex piece of code, just to make sure that class methods show up in the documentation as a class method instead of an instance method.  I've been reading this for about an hour now trying to find ways to make it simpler.  I've learned a ton about how name binding and closures and exec all work.  I haven't found anything to make it simpler, but I am finding that every question I have leads down a rabbit-hole.  I suspect this will be a "carter is the only one who can touch this code without breaking it" module, and that scares me.

azure-iot-device/azure/iot/device/patch.py
```diff
@@ -0,0 +1,133 @@
...
+    while classname_alias in globals():
+        classname_alias += "_"
+
+    # Import the class we're adding methods to, so that functions defined in this scope can use super()
+    class_module = inspect.getmodule(target_class)
+    exec(
+        "from {module} import {target_class} as {alias}".format(
+            module=class_module.__name__, target_class=target_class.__name__, alias=classname_alias
+        ),
+        globals(),
```
BertKleewein: >       globals(), [](start = 1, length = 17)

why do you have to specify globals here?  The docs I've found are fairly vague.  Isn't this the default?

azure-iot-device/azure/iot/device/patch.py
BertKleewein: One thing that I think would help in maintainability would be to add debug output for every exec call that lets developers grep and read the debug output almost like it's a source file.  e.g.:
```
INFO:azure.iot.device.iothub.patch: exec: from azure.iot.device import IoTHubModuleClient as _IoTHubModuleClient
INFO:azure.iot.device.iothub.patch: exec: def receive_input_message(self, input_name, block=True, timeout=None): return super(_IotHubModuleClient, self).receive_input_message(self=self, input_name=input_name, block=block, timeout=timeout))
INFO:azure.iot.device.iothub.patch: exec: receive_input_message.__doc__ = method_object.__doc__
INFO:azure.iot.device.iothub.patch: exec: setattr(target_class, method_name, receive_input_message)
```

This makes me think you conceptionally have two namespaces here, both living inside globals():
1. The namspace that you get from being a python module without using exec().  Things like method_object and target_class fall into this bucket.
2. The namespace additions that you get by using exec.  _IoTHubModuelClient falls into this bucket.

It seems like exec() code should only use the second, and non-exec() code should use the first.  I'm saying this because I want to look at the strings that go into exec() and understand the patch completely.  I want to be able to copy the text from debug output and make a python file if I want.  Not that I'd ever do that, but it gives me another level to understand the code.  Another way to think about this is that patch.py is emitting code, and I want to be able to look at that code outside the context of patch.py and understand what it's doing.  In the debug output above, the first two lines make complete sense, but I can't understand what the third line is doing without knowing what method_object is set to.

Instead, I'm  wondering if the strings that we pass to exec would be better off looking like this:
```
# first two lines stay the same
INFO:azure.iot.device.iothub.patch: exec: from azure.iot.device import IoTHubModuleClient as _IoTHubModuleClient
INFO:azure.iot.device.iothub.patch: exec: def receive_input_message(self, input_name, block=True, timeout=None): return super(_IotHubModuleClient, self).receive_input_message(self=self, input_name=input_name, block=block, timeout=timeout))
# next two lines change
INFO:azure.iot.device.iothub.patch: exec: receive_input_message.__doc__ = _IoTHubModuleClient.receive_input_messages.__doc__
INFO:azure.iot.device.iothub.patch: exec: setattr(_IoTHubModuleClient, receive_input_messages, receive_input_message)
```

With this, I can look at debug output and see problems I might not otherwise see.

(oops, receive_input_messages isn't on GenericIoTHubClient.  My bad.)



---
In reply to: [315356054](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315356054) [](ancestors = 315356054)

azure-iot-device/azure/iot/device/patch.py
BertKleewein: I think your code is just fine.  I'm just lamenting the absence of a better solution because this is a common problem. 

---
In reply to: [315340955](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315340955) [](ancestors = 315340955,315326449,315324838)

azure-iot-device/azure/iot/device/patch.py
BertKleewein: Another observation that comes out of this exercise is that your code leaves functions in the local namespace that don't need to be there.  in my example above, "receive_input_messages" is a function in the local namespace, and it doesn't need to be there.  No big deal, but the namespace exposed by the patch module is going to be a mish-mash of these dynamic functions from different classes and that just seems sloppy.

---
In reply to: [315386772](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315386772) [](ancestors = 315386772,315356054)

azure-iot-device/azure/iot/device/patch.py
cartertinney: When `exec` is called without any `globals` or `locals` optional args (side note: I need to change the syntax to represent this is an optional arg), exec operates within the current scope, i.e. the local scope of this function `add_shims_for_inherited_methods`. You can override the default scope by providing your own `globals` or `locals`. In this case, we don't bother with `locals` because changes made there aren't reflected - think of it as read-only. Since we're using `exec` to actually define new things, we need to use a globals scope.

For more information, read this doc: [https://docs.python.org/3/library/functions.html#exec](https://docs.python.org/3/library/functions.html#exec)

---
In reply to: [315357730](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315357730) [](ancestors = 315357730)

azure-iot-device/azure/iot/device/patch.py
cartertinney: Re: Comment 1 - I also was concerned about this, and made an effort to document and explain as much as I could with comments as a result. If you feel there's anything lacking, please let me know specifically, and I'll try to capture that inline.

---
In reply to: [315356054](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315356054) [](ancestors = 315356054)

cartertinney: Re: Comment 2 - I think this is a great idea, both the logging, and the explicit division of namespacing. I'll add some additional logging now, but before I get further invested in this code by altering the namespacing, I want to make sure this works for the docs team first. Once we know they can ingest it properly, I'll come back and make these changes in a separate PR. I'll add an item to the taskboard for it now.

---
In reply to: [315453647](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315453647) [](ancestors = 315453647,315356054)

cartertinney: Re: Comment 3 - It seems this problem would be resolved by manually maintaining a separate namespace for `exec`'d definitions, as mentioned in the above comment. They would still exist there, but it wouldn't be tied to the `patch.py` module.

---
In reply to: [315456629](https://github.com/Azure/azure-iot-sdk-python-preview/pull/164#discussion_r315456629) [](ancestors = 315456629,315453647,315356054)

PR #165, opened 2019-08-14
Add setAsSecurityMessage feature

necoh:


azure-iot-device/azure/iot/device/iothub/models/message.py
```diff
@@ -53,6 +54,10 @@ def __init__(
         self.content_encoding = content_encoding
         self.content_type = content_type
         self.output_name = output_name
+        self.iothub_interface_id = None
```
cartertinney: >self.iothub_interface_id = None [](start = 8, length = 31)

This attribute should be prefaced with an underscore, i.e. `self._iothub_interface_id` to indicate it should not be altered by the user (unlike the other attributes, which can be altered)

azure-iot-device/azure/iot/device/iothub/models/message.py
```diff
@@ -5,6 +5,7 @@
 # --------------------------------------------------------------------------
 """This module contains a class representing messages that are sent or received.
 """
+SECURITY_MESSAGE_INTERFACE_ID = "urn:azureiot:Security:SecurityAgent:1"
```
olivakar: >SECURITY_MESSAGE_INTERFACE_ID [](start = 0, length = 29)

is this a constant across everything ? Then should it be set in the `constant.py` file ?

azure-iot-device/azure/iot/device/iothub/models/message.py
necoh: Indeed this attribute could be set on a received message, therefore it is defined as a public attribute.

necoh: Indeed this attribute could be set on a received message, therefore it is defined as a public attribute.

azure-iot-device/azure/iot/device/iothub/models/message.py
cartertinney: This is an interesting idea. I'm good with it in either location for the time being, but I could see the argument for both.

If this value will ever change, it should be in `constant.py` for sure though. Otherwise I'm fine with it here.

cartertinney:
Merged in #176

PR #166, opened 2019-08-14
Fix samples link in README.md

lupino3:


cartertinney:
Included as part of #170

PR #168, opened 2019-08-16
reconnect the transport after the SAS token is renewed

BertKleewein:


azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -49,7 +49,7 @@ def delegate_to_different_op(stage, original_op, new_op):
       original_op in a way that is more specific than the original_op.
     """
 
-    logger.info("{}({}): continuing with {} op".format(stage.name, original_op.name, new_op.name))
+    logger.debug("{}({}): continuing with {} op".format(stage.name, original_op.name, new_op.name))
```
cartertinney: >debug [](start = 11, length = 5)

Good change here. We should start doing this throughout the client tbh

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -80,6 +80,64 @@ def _execute_op(self, op):
...
+            and self.pipeline_root.connected
+        ):
+            logger.info(
+                "{}({}): Connected.  Passing op down and reconnecting after token is updated.".format(
+                    self.name, op.name
+                )
+            )
+
+            # make a callback that can call the user's callback after the reconnect is complete
+            def on_reconnect_complete(reconnect_op):
```
cartertinney: >def on_reconnect_complete(reconnect_op): [](start = 12, length = 40)

This could probably be a private function defined within on_token_update_complete, since it's only used there. Might help readability.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -80,6 +80,64 @@ def _execute_op(self, op):
                 ),
             )
 
+        elif (
+            isinstance(op, pipeline_ops_base.UpdateSasTokenOperation)
+            and self.pipeline_root.connected
```
cartertinney: >self.pipeline_root.connected [](start = 16, length = 28)

My understanding is that if we are not connected, this op gets passed like any other unhandled op, right?

If so, a comment here to that effect could bring clarity, it took me a while to realize this.

cartertinney:
Good PR. Only minor suggestions. Otherwise, I'm on board.

PR #169, opened 2019-08-19
changes to return registration result

olivakar:


azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
```diff
@@ -41,14 +41,17 @@ def register(self):
         """
         logger.info("Registering with Provisioning Service...")
         register_complete = Event()
+        registration_result = [1]
```
cartertinney: >registration_result = [1] [](start = 8, length = 25)

1) Why are you using a list instead of a regular variable? Local variables should be settable via the callback
2) Even if you were to use a list, initializing a list with a value of `1` and then overwriting that value by indexing is not a typical pattern at all.

azure-iot-device/azure/iot/device/provisioning/models/registration_result.py
```diff
@@ -46,6 +46,9 @@ def status(self):
     def registration_state(self):
         return self._registration_state
 
+    def __str__(self):
+        return "\n".join([str(self.registration_state), self.status])
```
cartertinney: >"\n" [](start = 15, length = 4)

I'd prefer this was on one line, but it's not a big deal.

azure-iot-device/samples/advanced-hub-scenarios/register_symmetric_key.py
```diff
@@ -6,8 +6,11 @@
 
 import os
 import asyncio
+import logging
 from azure.iot.device.aio import ProvisioningDeviceClient
 
+logging.basicConfig(level=logging.ERROR)
```
cartertinney: Remove logging from samples

azure-iot-device/samples/legacy-samples/register_symmetric_key.py
```diff
@@ -4,8 +4,12 @@
 # license information.
 # --------------------------------------------------------------------------
 import os
+import logging
 from azure.iot.device import ProvisioningDeviceClient
 
+logging.basicConfig(level=logging.ERROR)
```
cartertinney: Remove logging from samples

azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py
```diff
@@ -20,11 +23,20 @@
...
+fake_sub_status = "FlyingOnHippogriff"
+fake_operation_id = "quidditch_world_cup"
+fake_request_id = "request_1234"
+fake_device_id = "MyNimbus2000"
+fake_assigned_hub = "Dumbledore'sArmy"
+
+fake_registration_state = RegistrationState(fake_device_id, fake_assigned_hub, fake_sub_status)
 
 
 def create_success_result():
```
cartertinney: Should be a fixture

cartertinney:
>         callback()

Since this is used in both sync and async, you should define these either in `tests/provisioning/conftest.py` or a `fixtures.py` file that is imported into the `conftest.py`

---
Refers to: azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py:106 in 82d6f9f. [](commit_id = 82d6f9fc0156cdc73fc253d2ae35b32f7f1e0869, deletion_comment = False)

cartertinney:
> 

Likewise, this should be shared

---
Refers to: azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py:114 in 82d6f9f. [](commit_id = 82d6f9fc0156cdc73fc253d2ae35b32f7f1e0869, deletion_comment = False)

cartertinney:
>         callback(create_success_result(), error=None)

By hardcoding this in, you ensure this can only return success.

What we do for similar methods in iothub (i.e. `get_twin`) is trigger the callback with all arguments to `None` and then override it within the test itself (where the successful result is a fixture).

This allows flexibility and clarity within tests, as there isn't test magic occurring that returns a value with no explanation - everything you need to know to understand the test is in the test itself.

---
Refers to: azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py:100 in 82d6f9f. [](commit_id = 82d6f9fc0156cdc73fc253d2ae35b32f7f1e0869, deletion_comment = False)

azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py
```diff
@@ -111,10 +123,16 @@ def test_client_register_success_calls_polling_machine_register_with_callback(
         mock_polling_machine_init.return_value = mock_polling_machine_success
 
         client = ProvisioningDeviceClient(mqtt_provisioning_pipeline)
-        client.register()
+        result = client.register()
 
         assert mock_polling_machine_success.register.call_count == 1
         assert callable(mock_polling_machine_success.register.call_args[1]["callback"])
+        assert result is not None
```
cartertinney: >assert result is not None [](start = 8, length = 25)

Building on what was said above re: fixtures, once the expected result is brought into the test explicitly as a fixture (say, `success_result`), you can then compare the result directly to the expected result.

You also probably don't need to check all the individual values - if you `assert result is success_result`, you know that the object returned is the same one. This is nice because it also allows us to remove the magic of somehow knowing that the result registration state should be `fake_registration_state`, when, again, from reading just this test, we have no way of knowing what the expected result should be.

Admittedly, though, now that I'm thinking about it, this approach doesn't account for the possibility that the object was modified on the way to being returned....

azure-iot-device/tests/provisioning/models/test_registration_result.py
```diff
@@ -53,27 +44,53 @@ def test_registration_result_instantiated_correctly(self):
...
+        "Has a to string representation composed of device id, assigned hub and sub status"
+    )
+    def test_registration_state_to_string(self):
+        registration_state = create_registraion_state()
+
+        string_repr = "\n".join([fake_device_id, fake_assigned_hub, fake_sub_status])
+        assert str(registration_state) == string_repr
+
+
+def create_registraion_state():
```
cartertinney: This should be a fixture

azure-iot-device/tests/provisioning/models/test_registration_result.py
```diff
@@ -53,27 +44,53 @@ def test_registration_result_instantiated_correctly(self):
...
+        fake_device_id,
+        fake_assigned_hub,
+        fake_sub_status,
+        fake_created_dttm,
+        fake_last_update_dttm,
+        fake_etag,
+    )
+
+
+def create_registration_result(registration_state=None):
```
cartertinney: Also a fixture

azure-iot-device/tests/provisioning/models/test_registration_result.py
```diff
@@ -53,27 +44,53 @@ def test_registration_result_instantiated_correctly(self):
...
 
+    @pytest.mark.it("Has a to string representation composed of registration state and status")
+    def test_registration_result_to_string(self):
+        fake_registration_state = create_registraion_state()
+        registration_result = create_registration_result(fake_registration_state)
+
+        string_repr = "\n".join([str(fake_registration_state), fake_status])
+        assert str(registration_result) == string_repr
+
     @pytest.mark.it("Has attributes that do not have setter")
```
cartertinney: >Has attributes that do not have setter [](start = 21, length = 38)

This should be parametrized so it has a more informative name

azure-iot-device/tests/provisioning/models/test_registration_result.py
```diff
@@ -53,27 +44,53 @@ def test_registration_result_instantiated_correctly(self):
...
+        registration_result = create_registration_result()
         with pytest.raises(AttributeError, match="can't set attribute"):
-            registration_result.request_id = "MyNimbus2000"
+            registration_result.request_id = "RequestId123"
             registration_result.operation_id = "WhompingWillow"
-            registration_result.status = "Flying"
+            registration_result.status = "Apparating"
             registration_result.registration_state = "FakeRegistrationState"
 
     @pytest.mark.it("Has `RegistrationState` with properties that do not have setter")
```
cartertinney: >Has `RegistrationState` with properties that do not have setter" [](start = 21, length = 64)

As above, parametrize this. As is, this does not actually test all properties. It only tests the first one that fails (device id)

azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py
olivakar: as we discussed that currently to pass fixture to the mock polling machine callback I have to also edit the logging method. also individual attribute assert can not be done , so I am keeping the function and not an fixture

---
In reply to: [315792459](https://github.com/Azure/azure-iot-sdk-python-preview/pull/169#discussion_r315792459) [](ancestors = 315792459)

olivakar:
>         callback(create_success_result(), error=None)

I have overridden but not with a fixture, rather with the function

---
In reply to: [523104799](https://github.com/Azure/azure-iot-sdk-python-preview/pull/169#issuecomment-523104799) [](ancestors = 523104799)

---
Refers to: azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py:100 in 82d6f9f. [](commit_id = 82d6f9fc0156cdc73fc253d2ae35b32f7f1e0869, deletion_comment = False)

azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py
olivakar: check replies above

---
In reply to: [315808899](https://github.com/Azure/azure-iot-sdk-python-preview/pull/169#discussion_r315808899) [](ancestors = 315808899)

azure-iot-device/tests/provisioning/models/test_registration_result.py
olivakar: I did not make this a fixture yet. Though this can be dome but I am in favor of instantiating here..

---
In reply to: [315814200](https://github.com/Azure/azure-iot-sdk-python-preview/pull/169#discussion_r315814200) [](ancestors = 315814200)

